PR #6, opened 2020-11-11
JUnit isTrue to AssertJ

sambsnyd:
Creating a PR for Tyler's JUnit assertions -> AssertJ visitor

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
+++ b/src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
@@ -1,0 +1,124 @@
+package org.openrewrite.java.testing.junit4toassertj;
+
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.*;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.openrewrite.Formatting.EMPTY;
+import static org.openrewrite.Formatting.format;
+import static org.openrewrite.Tree.randomId;
+
+/**
+ * This is a refactoring visitor that will convert JUnit-style assertTrue() to assertJ asserts:
+ *
+ * assertTrue("This should be true", EXPRESSION) -> assert(EXPRESSION).with("This should be true").isTrue()
+ * assertTrue(EXPRESSION) -> assert(EXPRESSION).isTrue()
+ *
+ */
+public class ChangeAssertTrue extends JavaIsoRefactorVisitor {
+
+    private static final MethodMatcher assertTrueMatcher = new MethodMatcher("org.junit.Assert assertTrue(..)");
+
+    public ChangeAssertTrue() {
+        setCursoringOn();
+    }
+
+    @Override
+    public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu) {
+
+        maybeRemoveImport("org.junit.Assert");
+        return super.visitCompilationUnit(cu);
+    }
+
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method) {
+        J.MethodInvocation original = super.visitMethodInvocation(method);
+        if (!assertTrueMatcher.matches(method)) {
+            return original;
+        }
+
+        //The expression that evaluates to true/false is either the first argument (if using the one-argument form)
+        //or the second argument (if using the two-argument form)
+
+        List<Expression> originalArgs = original.getArgs().getArgs();
+        Expression assertExpression = originalArgs.size() == 1 ? originalArgs.get(0) : originalArgs.get(1);
+
+        //This create the `assertThat(<EXPRESSION>)` method invocation.
+        J.MethodInvocation assertSelect = new J.MethodInvocation(
+                randomId(),
+                null,
+                null,
+                J.Ident.build(randomId(), "assertThat", JavaType.Primitive.Void, EMPTY),
+                new J.MethodInvocation.Arguments(
+                        randomId(),
+                        Collections.singletonList(assertExpression.withPrefix("")),
+                        EMPTY
+                ),
+                null,
+                EMPTY
+        );
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
```diff
@@ -0,0 +1,124 @@
...
+import static org.openrewrite.Tree.randomId;
+
+/**
+ * This is a refactoring visitor that will convert JUnit-style assertTrue() to assertJ asserts:
+ *
+ * assertTrue("This should be true", EXPRESSION) -> assert(EXPRESSION).with("This should be true").isTrue()
+ * assertTrue(EXPRESSION) -> assert(EXPRESSION).isTrue()
+ *
+ */
+public class ChangeAssertTrue extends JavaIsoRefactorVisitor {
```
sambsnyd, line 24: Since this visitor requires no configuration add `@AutoConfigure` to make it easily addressable from YAML

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method) {
+        J.MethodInvocation original = super.visitMethodInvocation(method);
+        if (!assertTrueMatcher.matches(method)) {
+            return original;
+        }
+
+        //The expression that evaluates to true/false is either the first argument (if using the one-argument form)
+        //or the second argument (if using the two-argument form)
```
sambsnyd, line 47: Between JUnit4 and 5 they switched the position of the optional "message" parameter. In 4 the message always comes first. In 5 the message always comes last. I don't think we'd want to say "Oops you already ran our recipe to go from 4 to 5, so now you can't use the assertJ recipe which only supports 4". So I suspect we'd either want this to support both JUnit 4 and 5, or just 5. What do you think?

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
```diff
@@ -0,0 +1,124 @@
...
+                randomId(),
+                null,
+                null,
+                J.Ident.build(randomId(), "assertThat", JavaType.Primitive.Void, EMPTY),
+                new J.MethodInvocation.Arguments(
+                        randomId(),
+                        Collections.singletonList(assertExpression.withPrefix("")),
+                        EMPTY
+                ),
+                null,
```
sambsnyd, line 63: Leaving the `type` parameter of a `J.MethodInvocation` `null` can have some non-obvious consequences.
If you immediately serialize it back to text everything is OK. But if other visitors stop by they might be confused. The most likely visitor to stop by, be confused, and actually mess things up is `OrderImports` with `removeUnused` set to `true`. That visitor will look at this, see that the type is null, and happily remove the import for `assertThat` because it didn't find any usages. 

Constructing `JavaType.Method` is a bit annoying, but so long as you get the fully qualified name of the class in there it's much less likely that any subsequent other visitor will have troubles here. You may generally find it more convenient to use `treeBuilder` to construct method invocations for this reason.

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
```diff
@@ -0,0 +1,124 @@
...
+     *
+     * @param fullyQualifiedName Fully-qualified name of the class.
+     * @param method The static method to be imported.
+     */
+    public void maybeAddStaticImport(String fullyQualifiedName, String method) {
+        AddImport op = new AddImport();
+        op.setType(fullyQualifiedName);
+        op.setStaticMethod(method);
+        op.setOnlyIfReferenced(false);
+        andThen(op);
```
sambsnyd, line 121: The "maybe" bit of the name is misleading since this will always try to add the import. Perhaps we should add a `maybeAddStaticImport` alongside the existing `maybeAddImport` for scenarios such as this.

sambsnyd (commented):
Great work, Tyler. This is good stuff.

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
tkvangorder: This is so this visitor can be included declaratively in a recipe? I will add it.

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
tkvangorder: Yeah, I definitely wanted to support both migration from JUnit 4 and JUnit 5. My initial thinking here was to put all of the JUnit 4 visitors in the junit4toassertj package and then have similar visitors for junit5toassertj. This would keep each visitor single purpose. Would you prefer that we have one visitor for each type and then have logic for both types of asserts?

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
tkvangorder: Hah, I ran into exactly this issue with my "maybeAddStaticImport()" method. I did have `setOnlyIfReferenced` set to true but it was not adding my import. This makes total sense now that you are describing it. I also wasn't sure how to correctly setup the type but I like the idea of exploring the tree builder, I will give that a spin tonight.

src/main/java/org/openrewrite/java/testing/junit4toassertj/ChangeAssertTrue.java
tkvangorder: yep, I wasn't sure if this is was something that you wanted me to generalize (or if I have the correct function signature), so I kept it local for the time being. Would you like me to create a PR to add this to JavaRefactorVisitor?

tkvangorder:
Thanks for the feedback, really helpful!

tkvangorder:
Since we decided we are going to initial target junit5 to assertJ assertions, I am going to close this PR out and move this work to a feature branch.

PR #21, opened 2020-12-12
Implemented AssertSameToAssertThat visitor

SofiaBrittoSchwartz:
Implemented visitor to convert JUnit's `assertSame()` to assertJ's `assertThat().isSameAs()`.

Refrained from creating a generic `ToAssertThat` utility class considering that it would likely get deprecated in a few weeks with the projected changes to TreeBuilder.

### DIFF
--- a/src/test/kotlin/org/openrewrite/java/testing/junitassertj/AssertSameToAssertThatTest.kt
+++ b/src/test/kotlin/org/openrewrite/java/testing/junitassertj/AssertSameToAssertThatTest.kt
@@ -1,0 +1,228 @@
+/*
+ * Copyright 2020 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/openrewrite/java/testing/junitassertj/AssertSameToAssertThatTest.kt
```diff
@@ -0,0 +1,228 @@
...
+                import org.junit.jupiter.api.Test;
+
+                import static org.junit.jupiter.api.Assertions.assertSame;
+
+                public class A {
+ 
+                    @Test
+                    public void test() {
+                        String str = "string";
+                        assertSame(notification(), str, "Should be null");
```
tkvangorder, line 82: All the "Should be null" messages -> "These Should be the same"

tkvangorder (changes requested):
Looks good, just a minor "search and replace" and then we can merge this.

sambsnyd (approved):
Fix the thing Tyler mentioned then this looks fine to me.

tkvangorder (approved):
LGTM

PR #25, opened 2020-12-17
Reduce visibility of JUnit5 methods

### LINKED ISSUE #23: Reduce visibility of JUnit5 methods
It's idiomatic in junit5 to not have @BeforeEach, @Before, @Test, etc. methods be publicly visible. in junit4 they had to be public. so they should be reducing visibility of these methods when changing the annotations

image

tkvangorder:
This PR introduces logic to reduce the visibility of methods marked with JUnit annotations to "package" to comply with JUnit 5 best practices. Fixes #23

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/UpdateBeforeAfterAnnotations.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/UpdateBeforeAfterAnnotations.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2020 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


src/main/java/org/openrewrite/java/testing/junit5/UpdateBeforeAfterAnnotations.java
```diff
@@ -0,0 +1,90 @@
...
+        andThen(changeType);
+
+        return super.visitCompilationUnit(cu);
+    }
+
+    @Override
+    public J.MethodDecl visitMethod(J.MethodDecl method) {
+        J.MethodDecl m = super.visitMethod(method);
+
+        boolean changed = false;
```
sambsnyd, line 72: This looks unused

jkschneider:
LGTM

src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -15,20 +15,8 @@
 #
 ---
 type: specs.openrewrite.org/v1beta/visitor
-name: org.openrewrite.java.testing.junit5.ReplaceAnnotations
+name: org.openrewrite.java.testing.junit5.ReplaceIngoreAnnotations
```
sambsnyd: Not necessary to list this visitor explicitly since you've given it `@AutoConfigure` and a package that the Recipe's pattern will match

sambsnyd (changes requested):
Based on the CI results you need to update the expected value of the integration test. After that and my very minor comments this looks good.

Issue #24, opened 2020-12-16
Use RemoveImport rather than OrderImports in CleanupJUnitImports


jkschneider:
Otherwise, sorting only is possible, as shown in the following example from Netflix/eureka. Might have to go improve `RemoveImport` to accept removal of _packages_ rather than just types.

![image](https://user-images.githubusercontent.com/1697736/102391104-1d078180-3f8a-11eb-80d4-fdea7d5c4e3f.png)

pway99:
Reopening since it is affecting SpringBoot Upgrades and JUnit4To5 migration.
per discussion with @sambsnyd 
```
No call to OrderImports, no call to RemoveUnused imports
Honestly removing line 81 doAfterVisit(new OrderImports(false)); and updating tests might be all you have to do
Maybe also AutoFormat the surviving imports so they don't have extra blank lines
```

PR #26, opened 2020-12-23
JUnit recipe changed to removing unused imports

tkvangorder:


tkvangorder:
Fixes #24

build.gradle.kts
```diff
@@ -73,6 +73,7 @@ dependencies {
 
     testImplementation("org.junit.jupiter:junit-jupiter-api:latest.release")
     testImplementation("org.junit.jupiter:junit-jupiter-params:latest.release")
+    testImplementation(enforcedPlatform("com.fasterxml.jackson:jackson-bom:latest.release"))
```
sambsnyd, line 76: Hmm this concerns me. If we're running the tests with different dependencies than we ship with it would be easy for all the tests to pass when in production you'd get classpath problems 100% of the time.

Does it work if you say `implementation(platform("com.fasterxml.jackson:jackson-bom:latest.release"))` ? I suspect that if we use `platform` instead of `enforcedPlatform` it wouldn't be a problem for consumers of this library.

sambsnyd (changes requested):

build.gradle.kts
tkvangorder: I am good with changing, it still works. I did this because Jonathan removed the enforcedPlatform from the rewrite-xml (which in turn manifested runtime errors in the tests only.

sambsnyd:
This is out of date for rewrite 7.0. Closing for now but feel free to update & re-open if we still need this

PR #28, opened 2020-12-31
Created JUnitFailToAssertJFail visitor and tests

SofiaBrittoSchwartz:
Created visitor to migrate from JUnit's `fail(..)` to AssertJ's `fail(..)`.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junitassertj/JUnitFailToAssertJFail.java
+++ b/src/main/java/org/openrewrite/java/testing/junitassertj/JUnitFailToAssertJFail.java
@@ -57,7 +57,7 @@ public class JUnitFailToAssertJFail extends JavaIsoRefactorVisitor {
 
     @Override
     public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method) {
-        J.MethodInvocation original = super.visitMethodInvocation(method);
+        J.MethodInvocation original = method;
         if (!JUNIT_FAIL_MATCHER.matches(method)) {
             return original;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/junitassertj/JUnitFailToAssertJFail.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * This matcher finds the junit methods that will be migrated by this visitor.
+     */
+    private static final MethodMatcher JUNIT_FAIL_MATCHER = new MethodMatcher(
+            JUNIT_QUALIFIED_ASSERTIONS_CLASS_NAME + " fail(..)"
+    );
+
+    @Override
+    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method) {
+        J.MethodInvocation original = super.visitMethodInvocation(method);
```
sambsnyd, line 60: This may be a bit pedantic, but calling `super.visit...()` it's going to call all of the visit methods for the sub-AST elements, potentially returning a modified, non-original AST element if any sub-element has been altered.  _This_ visitor doesn't overload any other visit method, so you will always get the original back. But then, why call `super.visit...()` at all? Our AST elements are immutable so if you want to rename the parameter to original say 
```java
J.MethodInvocation original = method;
```
or rename the method argument.
```java
public J.MethodInvocation visitMethodInvocation(J.MethodInvocation original)
```

sambsnyd (approved):
This is good work. I suggest a really minor tweak but this is OK to merge.

PR #38, opened 2021-02-23
JUnit5 Assertions to Assert J assertions pre-packaged declarative recipes

aegershman:
In order to enable direct, easy consumption of the JUnit Assertions to AssertJ recipes from the `junitassertj` package, pre-package the recipes in a declarative yaml.

The recipe is titled `JUnit5AssertionsToAssertJMigration` (fully qualified, `org.openrewrite.java.testing.JUnit5AssertionsToAssertJMigration`)

### DIFF
--- a/src/main/resources/META-INF/rewrite/junit5-assertions-to-assertj.yml
+++ b/src/main/resources/META-INF/rewrite/junit5-assertions-to-assertj.yml
@@ -1,0 +1,30 @@
+#
+# Copyright 2020 the original author or authors.
+# <p>
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


sambsnyd:
Where's the bit making sure that all assertj static methods are imported statically?

aegershman:
@sambsnyd :

- I was going to do that in a separate pull request, but I can include that here as part of this.
- I'm thinking to change the name from `JUnitAssertionsToAssertJMigration` to `JUnit5AssertionsToAssertJMigration` to specifically including the `5` in the name to remove any ambiguity of whether this would work on a Junit-4 codebase without going to junit 5 first to make it clearer, if that's alright?

Issue #31, opened 2021-01-23
Recipe that ensures that the static import for AssertJ Assertions is always used.


jkschneider:
To avoid a bunch of these in the test code:

![image](https://user-images.githubusercontent.com/1697736/105609614-5ab4b280-5d5f-11eb-9124-db2351f253aa.png)

aegershman:
To clarify, which of these is the goal:

- **always** convert a qualified name of `Assertions.{someMethod...}` to the static method, regardless of whether the static import already exists. For example, always convert uses of the qualified `Assertions.someMethod` to static `someMethod`. Always.

OR

- **IF** the static method in question is _already imported_, then make sure that static method is used in all places where the qualified name is being used? For example, if `assertThat` is already statically imported, then convert any other uses of `Assertions.assertThat` in the file to be static. However, if static `assertThat` is NOT already imported, and all existing usages are qualified as `Assertions.assertThat`, then leave it be.

sambsnyd:
All imports of AssertJ methods should be static imports.
All invocations of AssertJ methods should be the receiver-less form you can only use when a static import for the method is present.

aegershman:
Wait, couldn't this just be done as-is with a normal `UseStaticImport`? https://github.com/openrewrite/rewrite-docs/blob/926f06a694cab31b2566d0f7c92c572cfaf5d4a5/reference/java/refactoring-java-source-code/usestaticimport.md

(source for it: https://github.com/openrewrite/rewrite/blob/8e9ccb2b4796deee48c17b244f31733b3962c5e2/rewrite-java/src/main/java/org/openrewrite/java/UseStaticImport.java)

I'll test this, but shouldn't this do everything?

```yml
type: specs.openrewrite.org/v1beta/recipe
name: org.openrewrite.java.testing.UseStaticAssertJAssertions
recipeList:
  - org.openrewrite.java.UseStaticImport:
    method: 'org.assertj.core.api.Assertions *(..)'
```

If this _does_ work, then I think this would be acceptable, right? Since this is stylistic, I don't know if we should package this in as it's own file shipped with the framework, but if we did, I could just put it in `META-INF.rewrite/use-static-assertj-assertions.yaml`

aegershman:
closed by https://github.com/openrewrite/rewrite-testing-frameworks/pull/38

Is referenced as part of...

```
<recipe>org.openrewrite.java.testing.JUnit5AssertionsToAssertJMigration<recipe>
```

PR #39, opened 2021-02-23
Separate assertj assertions using static imports from the junit5 assertions migration

aegershman:
see: https://github.com/openrewrite/rewrite-testing-frameworks/pull/38
see: https://github.com/openrewrite/rewrite-testing-frameworks/issues/31

Separates assertj assertions static imports from the junit5 assertions to assertj migration.

This provides two benefits:
1.) **It's less surprising**, since it gives users a choice on whether to use static imports with assertJ from JUnit 5 -> assertJ migration as a choice to make, and;
2.) **It's more efficient** by making it easier to consume "AssertJ Use Static Imports" as a separate recipe without having to waste time going through the other Junit migration recipes if that isn't desired (and vice versa).

Easy peezy PR, feel free to either merge or close whenever.

### DIFF
--- a/src/main/resources/META-INF/rewrite/assertj-assertions-use-static-import.yml
+++ b/src/main/resources/META-INF/rewrite/assertj-assertions-use-static-import.yml
@@ -1,0 +1,21 @@
+#
+# Copyright 2020 the original author or authors.
+# <p>
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sambsnyd:
It's easy to reintroduce Junit assertions even if you've once migrated to assertj assertions, since it's not like JUnit is ever going to stop being on the classpath. So this migration, more than most, is one you'd want to leave turned on in perpetuity if you want your assertions consistent. So this should remain a single recpie.

PR #40, opened 2021-02-23
Use correct assertThat static import, consolidate assertj declarative recipes

aegershman:
part of https://github.com/openrewrite/rewrite/issues/296

- Moves the `assertThat` recipes to `assertj.yml`, explicitly names them

I put these in their own recipe, `AssertJUseAssertThatFromAssertionsPackage`, because the recipes are conceptually different. I felt it's better to do smaller, goal-oriented recipes which can be opt-in. Changes welcome if that's what people think would otherwise be better.

### DIFF
--- a/src/main/resources/META-INF/rewrite/assertj.yml
+++ b/src/main/resources/META-INF/rewrite/assertj.yml
@@ -1,0 +1,25 @@
+#
+# Copyright 2020 the original author or authors.
+# <p>
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# <p>
+# https://www.apache.org/licenses/LICENSE-2.0
+# <p>
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.testing.AssertJUseAssertThatFromAssertionsPackage
+recipeList:
+  - org.openrewrite.java.ChangeMethodTargetToStatic:
+      methodPattern: "org.assertj.core.api.AssertionsForClassTypes assertThat(..)"
+      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions"
+  - org.openrewrite.java.ChangeMethodTargetToStatic:
+      methodPattern: "org.assertj.core.api.AssertionsForInterfaceTypes assertThat(..)"
+      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions"
--- a/src/main/resources/META-INF/rewrite/junit5-assertions-to-assertj.yml
+++ b/src/main/resources/META-INF/rewrite/junit5-assertions-to-assertj.yml
@@ -28,9 +28,3 @@ recipeList:
   - org.openrewrite.java.testing.junitassertj.JUnitFailToAssertJFail
   - org.openrewrite.java.UseStaticImport:
       methodPattern: "org.assertj.core.api.Assertions *(..)"
-  - org.openrewrite.java.ChangeMethodTargetToStatic
-      methodPattern: "org.assertj.core.api.AssertionsForClassTypes.assertThat *(..)"
-      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions.assertThat"
-  - org.openrewrite.java.ChangeMethodTargetToStatic
-    methodPattern: "org.assertj.core.api.AssertionsForInterfaceTypes.assertThat *(..)"
-    fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions.assertThat"


aegershman:
I didn't figure out how to run this as a unit test, so instead I ran it against a very simple test file using `./gradlew publishToMavenLocal` and consuming it in a very simple test project. The results look like this.

Given this test file...

```java
package io.example.openrewriteconsumerexample;

import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;

import java.util.ArrayList;
import java.util.List;

import org.assertj.core.api.AssertionsForClassTypes;
import org.assertj.core.api.AssertionsForInterfaceTypes;
import org.springframework.boot.test.context.SpringBootTest;

@SpringBootTest
public class AssertJUseAssertThatFromAssertionsPackageTests {

	@Test
	void assertionsForClassTypesQualified() {
		AssertionsForClassTypes.assertThat(true).isTrue();
	}

	@Test
	void assertionsForClassTypesStatic() {
		assertThat(true).isTrue();
	}

	@Test
	void assertionsForInterfaceTypesQualified() {
		List<String> exampleList = new ArrayList<>();
		AssertionsForInterfaceTypes.assertThat(exampleList).hasSize(0);
	}

	@Test
	void assertionsForInterfaceTypesStatic() {
		List<String> exampleList = new ArrayList<>();
		assertThat(exampleList).hasSize(0);
	}

}
```

and running the recipe `org.openrewrite.java.testing.AssertJUseAssertThatFromAssertionsPackage`

We get this:

```java
package io.slugstack.openrewriteconsumerexample;

import org.assertj.core.api.Assertions;
import org.springframework.boot.test.context.SpringBootTest;

import java.util.ArrayList;
import java.util.List;

import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;

@SpringBootTest
public class AssertJUseAssertThatFromAssertionsPackageTests {

	@Test
	void assertionsForClassTypesQualified() {
		Assertions.assertThat(true).isTrue();
	}

	@Test
	void assertionsForClassTypesStatic() {
		Assertions.assertThat(true).isTrue();
	}

	@Test
	void assertionsForInterfaceTypesQualified() {
		List<String> exampleList = new ArrayList<>();
		Assertions.assertThat(exampleList).hasSize(0);
	}

	@Test
	void assertionsForInterfaceTypesStatic() {
		List<String> exampleList = new ArrayList<>();
		Assertions.assertThat(exampleList).hasSize(0);
	}

}
```

src/main/resources/META-INF/rewrite/junit5-assertions-to-assertj.yml
```diff
@@ -28,9 +28,3 @@ recipeList:
   - org.openrewrite.java.testing.junitassertj.JUnitFailToAssertJFail
   - org.openrewrite.java.UseStaticImport:
       methodPattern: "org.assertj.core.api.Assertions *(..)"
-  - org.openrewrite.java.ChangeMethodTargetToStatic
-      methodPattern: "org.assertj.core.api.AssertionsForClassTypes.assertThat *(..)"
-      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions.assertThat"
-  - org.openrewrite.java.ChangeMethodTargetToStatic
-    methodPattern: "org.assertj.core.api.AssertionsForInterfaceTypes.assertThat *(..)"
-    fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions.assertThat"
```
sambsnyd, line 36: Normally I would say that it is sensible to separate a "best practices" recipe from a "migration" recipe. Migrations you run once, best practices you run continually. 
But, as I alluded to in the other pull request, assertj's migration is also simultaneously a best practices recipe since it's _super easy_ to reintroduce JUnit assertions even after running the migration once. 

So I don't really care if these ChangeMethodTargetToStatics are put in their own recipe as well, but the overall recipe should include that recipe if they are

src/main/resources/META-INF/rewrite/assertj.yml
```diff
@@ -0,0 +1,25 @@
...
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.testing.AssertJUseAssertThatFromAssertionsPackage
+recipeList:
+  - org.openrewrite.java.ChangeMethodTargetToStatic:
+      methodPattern: "org.assertj.core.api.AssertionsForClassTypes assertThat(..)"
+      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions"
+  - org.openrewrite.java.ChangeMethodTargetToStatic:
+      methodPattern: "org.assertj.core.api.AssertionsForInterfaceTypes assertThat(..)"
+      fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions"
```
sambsnyd, line 25: Yaml files can include multiple yaml documents (a new yaml document begins wherever you have that non-indented `---` line). I suggest unifying both assertj yamls into assertj.yml.

sambsnyd (changes requested):

sambsnyd:
Oh, and to write tests for declarative recipes, see [loadRecipeFromClasspath](https://github.com/openrewrite/rewrite/blob/de28afbdb6b802e1a050dee73b8028cb5343de50/rewrite-test/src/main/kotlin/org/openrewrite/test.kt#L26).
[Example usage](https://github.com/openrewrite/rewrite-testing-frameworks/blob/1dc63f6ed301c57483ffa0409f763c7142fbee96/src/test/kotlin/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.kt#L34) from rewrite-testing-frameworks.

src/main/resources/META-INF/rewrite/junit5-assertions-to-assertj.yml
aegershman: There isn't currently an 'overall' / "opinionated default" recipe for assertj, should I make one and hang these off that? The closest thing is "migrate junit5 assertions to assertj assertions". I think that'd solve issues, honestly. That's be like the loose conceptual equivalent of AutoFormat (sort of) where it's a one-stop shop of recipes, which are composed of sub-recipes.

I can do that and call it something like `DefaultAssertJ`?

Issue #70, opened 2021-03-15
JUnit 4 to 5 TestWatcher migration recipe


pway99:
Recipe to convert JUnit 4 TestWatcher to the JUnit5 variant

JUnit-4
https://junit.org/junit4/javadoc/4.12/org/junit/rules/TestWatcher.html

JUnit-5
https://junit.org/junit5/docs/5.4.0-RC2/api/org/junit/jupiter/api/extension/TestWatcher.html

aegershman:
Here's a breakdown. First a proposed workflow, then sections on the problems at each step:

## proposed workflow:

- For any `TestWatcher` instantiation in a test class, create a private class in that same public class file which `implements TestWatcher` (using the name `VariableNameTestWatcherExtension`, maybe)
- Add `@ExtendWith(VariableNameTestWatcherExtension.class)` to the class the TestWatcher was originally implemented in
- Convert the corresponding junit 4 method signatures to the Junit 5 signatures where possible.
  - as for the signatures which don't correspond exactly (`starting`, `skipped`, etc.), add another `implements` with the corresponding `Lifecycle` handlers. Not 100% about this, we'll see.
- Cleanup; as in, remove the old JUnit 4 TestWatcher instantiation, etc.

---

## problem 0: moving from AbstractClass to Interface

- In JUnit 4, TestWatcher is an abstract class.
- In Junit 5, TestWatcher is an interface.
- This can be solved; just pointing it out

## problem 1: callback method equivalents

- In JUnit 4, the TestWatcher Rule API provides us with callback methods.
- Some of these callback methods are no longer present in JUnit 5.
- In JUnit 4, for example, we had "protected void failed", "Invoked when a test fails", which has a generally-analogous method in JUnit 5, "testFailed", "Invoked after a test has failed".
- But some callback methods are no longer present in JUnit 5's TestWatcher.
- For example, in JUnit 4 there was "starting", "Invoked when a test is about to start". In JUnit 5, however, TestWatcher only appears to have methods which deal with the results of tests; there is no "testStarting".
- We can still work around this.

## problem 2a: addition of ExtensionContext

- Hopefully not a big deal though since it wasn't already in-use

## problem 2b: moving from `Description` to `ExtensionContext.

- There are similarities in usage, though.

## problem 3: what of static access?

- Static access for variables moving around within/with/against TestWatcher and the test classes

## problem/solution 4: registration of ExtendWith

- Not really that much of a problem. Generally a solution.
[...]

PR #74, opened 2021-03-17
TestWatcher API for JUnit 4 to JUnit 5

aegershman:
see: https://github.com/openrewrite/rewrite-testing-frameworks/issues/70

wip


aegershman:
By virtue of this PR being opened, the content would be available to see in the future even though it isn't merged. So content such as tests, PR body, etc., are still visible.

Closing without merge, though content is still visible.

PR #97, opened 2021-03-25
Adding a configurable list of runners to RemoveObsoleteRunners

tkvangorder:


tkvangorder:
Adding a parameter to RemoveObsoleteRunners so it can be reused in other places where the action is to simply remove the runner when migrating to Junit5

PR #111, opened 2021-03-31
enable rewrite plugin

aegershman:
see https://github.com/openrewrite/rewrite-gradle-plugin/issues/18

~~placeholder draft / some small experimentation to suss out some preferences for feedback~~

Not the most sophisticated initial configuration, and that's intentional. Can be ran (as of 3.1.0) with `./gradlew rewriteWarn` and/or `./gradlew rewriteFix`

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,10 +33,15 @@ plugins {
 
     id("nebula.integtest") version "7.0.9" apply false
     id("me.champeau.gradle.jmh") version "0.5.2"
+    id("org.openrewrite.rewrite") version "3.1.0"
 }
 
 apply(plugin = "nebula.integtest-standalone")
 apply(plugin = "nebula.publish-verification")
+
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -33,11 +33,16 @@ plugins {
 
     id("nebula.integtest") version "7.0.9" apply false
     id("me.champeau.gradle.jmh") version "0.5.2"
+    id("org.openrewrite.rewrite") version "3.1.0"
 }
 
 apply(plugin = "nebula.integtest-standalone")
 apply(plugin = "nebula.publish-verification")
 
+configure<org.openrewrite.gradle.RewriteExtension> {
```
sambsnyd, line 42: You should be able to get away with just `rewrite {` here. 
After you add the entry to the plugins block click this button in the IDE
image

Then new extensions made available by the plugin are accessible.

PR #115, opened 2021-04-05
Add MockitoRule recipe.

### LINKED ISSUE #86: JUnit 4 to 5 MockitoRule recipe
Discovered while applying `Boot2JUnit4to5Migration` on [spring-cloud-shell](https://github.com/spring-projects/spring-shell)

This task is to create a [MockitoRule](https://www.javadoc.io/doc/org.mockito/mockito-core/2.5.0/org/mockito/junit/MockitoRule.html) migration recipe

spring-shell [ShellTest](https://github.com/spring-projects/spring-shell/blob/296db93462eac8a9d6b10009a2e3f7b9fe22f5ad/spring-shell-core/src/test/java/org/springframework/shell/ShellTest.java#L51-L52)

traceyyoshima:
fixes #86 
Added MockitoRule recipe.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -116,6 +116,7 @@ dependencies {
     testRuntimeOnly("org.springframework:spring-test:4.+")
     testRuntimeOnly("ch.qos.logback:logback-classic:1.0.13")
     testRuntimeOnly("org.mockito:mockito-all:$mockito1Version")
+    testRuntimeOnly("org.mockito:mockito-core:latest.release")
     testRuntimeOnly("org.hamcrest:hamcrest:latest.release")
     testRuntimeOnly("pl.pragmatists:JUnitParams:1.+")
     testRuntimeOnly("com.squareup.okhttp3:mockwebserver:3.+")
--- a/src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitToMockitoExtension.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitToMockitoExtension.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2020 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.testing.junit5;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Parser;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.format.AutoFormatVisitor;
+import org.openrewrite.java.search.FindAnnotations;
+import org.openrewrite.java.search.FindFields;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Statement;
+import org.openrewrite.java.tree.TypeUtils;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Replaces JUnit 4 MockitoJUnit rules with JUnit MockitoExtension.
+ *
+ * Supported MockitoJUnit methods:
+ *  #rule()
+ *  #testRule()
+ *
+ * Does not currently support @Incubating MockitoJUnit.collector().
+ *
+ * @implNote collector() is designed to aggregate multiple verifications into a single output.
+ * Refactoring the method may be fairly complex and would likely benefit from being a separate recipe.
+ *
+ * Must be ran in the JUnit5 suite.
+ */
+public class MockitoJUnitToMockitoExtension extends Recipe {
+    private static final ThreadLocal<JavaParser> JAVA_PARSER = ThreadLocal.withInitial(() ->
+            JavaParser.fromJavaVersion()
+                    .dependsOn(Arrays.asList(
+                            Parser.Input.fromString("package org.junit.jupiter.api.extension;\n" +
+                                    "public @interface ExtendWith {\n" +
+                                    "Class[] value();\n" +
+                                    "}"),
+                            Parser.Input.fromString("package org.mockito.junit.jupiter;\n" +
+                                    "public class MockitoExtension {\n" +
+                                    "}")
+                    ))
+                    .build()
+    );
+
+    @Override
+    public String getDisplayName() {
+        return "MockitoJUnit to MockitoExtension";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replaces MockitoJUnit rules with MockitoExtension.";
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new MockitoRuleToMockitoExtensionVisitor();
+    }
+
+    public static class MockitoRuleToMockitoExtensionVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private static final String EXTEND_WITH_FQN = "org.junit.jupiter.api.extension.ExtendWith";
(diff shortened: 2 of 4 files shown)


src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitToMockitoExtension.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    public static class MockitoRuleToMockitoExtensionVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private static final String MOCKITO_RULE_FQN = "org.mockito.junit.MockitoRule";
+        private static final String MOCKITO_TEST_RULE_FQN = "org.mockito.junit.MockitoTestRule";
+
+        private static final String MOCKITO_RULE_INVOCATION_KEY = "mockitoRuleInvocation";
+        private static final String MOCKITO_TEST_RULE_INVOCATION_KEY = "mockitoTestRuleInvocation";
+
+        @SneakyThrows
```
sambsnyd, line 100: As much as I dislike checked exceptions I'm not sure about this. If surrounding code were to try and catch a checked exception that was thrown "sneakily" IDEs and the like would put up spurious warnings about catching it. What checked exception is being thrown in this case?

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitToMockitoExtension.java
```diff
@@ -0,0 +1,219 @@
...
+                                            J.Annotation::getSimpleName,
+                                            new RuleBasedCollator("< ExtendWith")
+                                    ))
+                    );
+
+                    maybeAddImport(extendWithFqn);
+                    maybeAddImport(mockitoExtensionFqn);
+
+                    // @ExtendsWith(MockitoExtension.class) depends on the JUnit5 test annotation to work.
+                    doAfterVisit(new UpdateTestAnnotation());
```
sambsnyd, line 140: Delete these two lines.

It isn't so much that `@ExtendsWith(MockitoExtension.class)` depends on the JUnit5 version of `@Test` as it is that `@ExtendsWith` is itself a JUnit 5 concept. Upgrading just the `@Test` annotation and leaving the rest of the project JUnit4 would obviously not be enough. So to really be sure you'd have to `doAfterVisit()` the whole migration recipe. But that migration recipe is going to _include_ this recipe, so that'd be a mess. Best case makes performance worse. Worst case messes things up some way I don't anticipate.

build.gradle.kts
```diff
@@ -116,6 +116,7 @@ dependencies {
     testRuntimeOnly("org.springframework:spring-test:4.+")
     testRuntimeOnly("ch.qos.logback:logback-classic:1.0.13")
     testRuntimeOnly("org.mockito:mockito-all:$mockito1Version")
+    testRuntimeOnly("org.mockito:mockito-core:latest.release")
```
sambsnyd, line 119: `testRuntimeOnly("org.mockito:mockito-core:latest.release")` is going to pull in Mockito 3. Are the symbols this operates on not present in mockito 1? _Some_ symbols overlap between Mockito 1 and 3, I wonder if this could cause any problems.

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitToMockitoExtension.java
```diff
@@ -0,0 +1,219 @@
...
+ * Supported MockitoJUnit methods:
+ *  #rule()
+ *  #testRule()
+ *
+ * Does not currently support @Incubating MockitoJUnit.collector().
+ *
+ * @implNote collector() is designed to aggregate multiple verifications into a single output.
+ * Refactoring the method may be fairly complex and would likely benefit from being a separate recipe.
+ */
+public class MockitoJUnitToMockitoExtension extends Recipe {
```
sambsnyd, line 48: You need to add a line to src/main/resources/META-INF/rewrite/junit.yml to include this recipe into the JUnit4to5Migration recipe

sambsnyd (changes requested):
Good work. Couple questions and bits of feedback

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitToMockitoExtension.java
traceyyoshima: I snaked this concept from a recipe Tyler wrote, the new RuleBasedCollator("< ExtendWith") has a checked java.text.ParseException.

I'll ask him about it tomorrow, and see if there's anything I can do to refactor the code.

build.gradle.kts
traceyyoshima: I'll check on this tomorrow. The tests fail without it due to missing class info.

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitToMockitoExtension.java
traceyyoshima: Alright, I'll create a test with an existing annotation, and make sure it isn't added if it already exists.

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitToMockitoExtension.java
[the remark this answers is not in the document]
traceyyoshima: Ah potentially, I'll create a test with an existing annotation, and make sure it isn't added if it already exists.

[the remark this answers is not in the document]
traceyyoshima: Updated to check if the @RunWith(MockitoRunner.class) exists and doesn't add the annotation when it exists

src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -96,6 +96,8 @@ recipeList:
   - org.openrewrite.java.testing.junit5.UpdateTestAnnotation
   - org.openrewrite.java.testing.junit5.ParameterizedRunnerToParameterized
   - org.openrewrite.java.testing.junit5.JUnitParamsRunnerToParameterized
+  - org.openrewrite.java.testing.junit5.MockitoJUnitToMockitoExtension
+  - org.openrewrite.java.testing.junit5.ExpectedExceptionToAssertThrows
```
traceyyoshima, line 100: Added the ExpectedExceptionToAssertThrows, since it was missed in a previous PR.

src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitToMockitoExtension.java
```diff
@@ -0,0 +1,206 @@
...
+                maybeRemoveImport("org.mockito.junit.MockitoJUnit");
+                maybeRemoveImport("org.mockito.quality.Strictness");
+
+
+                if (classDecl.getBody().getStatements().size() != cd.getBody().getStatements().size() &&
+                        (FindAnnotations.find(classDecl.withBody(null), MOCKITO_JUNIT_RUNNER_ANNOTATION).isEmpty() &&
+                                FindAnnotations.find(classDecl.withBody(null), EXTEND_WITH_MOCKITO_EXTENSION).isEmpty())) {
+
+                    final String extendWithFqn = "org.junit.jupiter.api.extension.ExtendWith";
+                    final String mockitoExtensionFqn = "org.mockito.junit.jupiter.MockitoExtension";
```
sambsnyd, line 115: These could also be static final fields

src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitToMockitoExtension.java
```diff
@@ -0,0 +1,206 @@
...
+                    .filter(it -> !isTargetMethodInvocation(it))
+                    .collect(Collectors.toList());
+
+            StringBuilder printedStatements = new StringBuilder();
+            for (Statement stmt : statements) {
+                printedStatements.append(stmt.print()).append(';');
+            }
+
+            return printedStatements.toString();
+        }
```
sambsnyd, line 190: Printing the Statements to a String only to re-parse back into Statements is unnecessary overhead. There's no need to serialize and re-parse when you already have the data in the right format.

You should do something like `m = m.withBody(m.getBody().withStatements(filteredListOfStatements));`.

sambsnyd (changes requested):
One final change then this looks good to go to me 👍

src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitToMockitoExtension.java
traceyyoshima: Oh, that's interesting. I'll refactor the code a bit, thanks for letting me know

sambsnyd (approved):

PR #123, opened 2021-04-19
Add recipe for Sonar RSPEC-2699 Tests should have assertions.

### LINKED ISSUE #121: Add Recipe for Sonarqube RSPEC-2699 Tests should include assertions
Tests that only have statements without any assertions prove that an exception is not thrown. Wrapping those statements in an `assertDoesNotThrow()` is more explicit and will satisfy RSPEC-2699

[RSPEC-2699 Tests should include assertions](https://rules.sonarsource.com/java/tag/tests/RSPEC-2699)

pway99:
Add recipe for Sonar RSPEC-2699 Tests should have assertions. Fixes #121

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/sonar/TestsShouldIncludeAssertions.java
+++ b/src/main/java/org/openrewrite/java/testing/sonar/TestsShouldIncludeAssertions.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2020 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.testing.sonar;
+
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+import org.openrewrite.*;
+import org.openrewrite.java.AnnotationMatcher;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.tree.J;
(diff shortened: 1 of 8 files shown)


src/main/resources/META-INF/rewrite/sonar.yml
```diff
@@ -0,0 +1,31 @@
...
+displayName: Sonar Best Practices for Java Testing Frameworks
+description: Sonar Best Practices for Java Testing Frameworks.
+tags:
+  - sonar
+  - testing
+recipeList:
+  - org.openrewrite.java.testing.sonar.TestsShouldIncludeAssertions:
+      assertionPackages:
+        - org.junit.jupiter.api.Assertions
+        - org.hamcrest.MatcherAssert
```
sambsnyd, line 28: We should definitely include AssertJ assertions in this list. testing-frameworks includes recipes for adopting AssertJ. It would be unfortunate to also ship a recipe unaware of them

src/main/java/org/openrewrite/java/testing/sonar/TestsShouldIncludeAssertions.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Option(displayName = "Assertion Packages",
+            description = "List of fully qualified package names of test assertion packages used for finding assertion statements.",
+            example = "org.junit.jupiter.api.Assertions, org.hamcrest.MatcherAssert")
+    List<String> assertionPackages;
+
+    @Option(displayName = "Assert Select Methods",
+            description = "List of fully qualified assertion method names used for finding assertion statements.",
+            example = "org.mockito.Mockito.verify, io.vertx.ext.unit.TestContext.verify")
+    List<String> assertMethods;
```
sambsnyd, line 72: These look really similar. What if it were a single configuration option that handled both fully qualified class and method names?

Also `assertionPackages` name is misleading since you populate it with fully qualified class names, not packages.

src/main/java/org/openrewrite/java/testing/sonar/TestsShouldIncludeAssertions.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    public String getDisplayName() {
+        return "Sonar RSPEC-2699. Tests should include assertions";
```
sambsnyd, line 76: I'd put the "Sonar RSPEC" bit into the description rather than the display name.

sambsnyd (changes requested):

PR #131, opened 2021-05-11
Implement inexpensive single source applicability tests

pway99:


src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertFalseToAssertThat.java
```diff
@@ -62,14 +64,18 @@ public String getDescription() {
         return "Convert JUnit-style assertFalse() to assertJ's assertThat().isFalse().";
     }
 
+    @Override
+    protected @Nullable TreeVisitor<?, ExecutionContext> getApplicableTest() {
```
jkschneider, line 68: Could we remove `@Nullable` from these? It's nullable in the `Recipe` base class for good reason, but in this specific recipe it will never be null.

PR #138, opened 2021-05-20
Disable test method access modifications

pway99:
Method access modifications require information from the super class to avoid assigning weaker access privileges.


src/test/kotlin/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.kt
```diff
@@ -309,6 +308,7 @@ class UpdateTestAnnotationTest : JavaRecipeTest {
         """
     )
 
+    @Disabled("test visibility changes require super class info")
```
sambsnyd, line 311: A link to a github issue would be preferable

PR #139, opened 2021-05-26
Use latest.integration when building normally, latest.release when releasing

sambsnyd:
This removes the "-x test" from the release builds, so they will now run tests. That takes a bit longer, but the previous CI runs will have been built against snapshots so they wont necessarily be valid results for releasing.


aegershman (approved):
lgtm 👍  hit it and let's try it out. The only thing that's an issue is what happens if the build fails, in which case the WORST case scenario is either "fix the issue and roll the git tag version forward" OR "fix the issue, remove the attempted git tag, and then re-release by re-tagging the latest commit".

In either case, let's try it out

PR #146, opened 2021-06-10
Update TemporaryFolderToTempDir.

pway99:
Remove generated newFile method declaration and fix associated remove issue caused by null types.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
@@ -15,19 +15,24 @@
  */
 package org.openrewrite.java.testing.junit5;
 
-import org.openrewrite.*;
-import org.openrewrite.java.JavaIsoVisitor;
-import org.openrewrite.java.JavaParser;
-import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Parser;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.*;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.*;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class TemporaryFolderToTempDir extends Recipe {
+    private static final JavaType.Class FILE_TYPE = JavaType.Class.build("java.io.File");
+    private static final JavaType.Class STRING_TYPE = JavaType.Class.build("java.lang.String");
     private static final ThreadLocal<JavaParser> TEMPDIR_PARSER = ThreadLocal.withInitial(() ->
             JavaParser.fromJavaVersion().dependsOn(Collections.singletonList(
                     Parser.Input.fromString("" +
@@ -38,7 +43,7 @@ public class TemporaryFolderToTempDir extends Recipe {
 
     @Override
     public String getDisplayName() {
-        return "Migrate JUnit 4 `TemporaryFolder` to JUnit Jupiter `TempDir`";
+        return "Use JUnit Jupiter `@TempDir`";
     }
 
     @Override
@@ -56,14 +61,16 @@ public class TemporaryFolderToTempDir extends Recipe {
         return new TemporaryFolderToTempDirVisitor();
     }
 
-    private static class TemporaryFolderToTempDirVisitor extends JavaIsoVisitor<ExecutionContext> {
-        private static final JavaType.Class FILE_TYPE = JavaType.Class.build("java.io.File");
-        private static final JavaType.Class STRING_TYPE = JavaType.Class.build("java.lang.String");
+    private static class TemporaryFolderToTempDirVisitor extends JavaVisitor<ExecutionContext> {
+        private static final AnnotationMatcher JUPITER_TEST_ANNOTATION_MATCHER = new AnnotationMatcher("@junit.jupiter.api.Test");
+        private static final AnnotationMatcher JUNIT_TEST_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.Test");
+        private static final AnnotationMatcher TEMPDIR_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.api.io.TempDir");
 
         @Override
-        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext executionContext) {
+        public J visitCompilationUnit(J.CompilationUnit cu, ExecutionContext executionContext) {
             J.CompilationUnit c = (J.CompilationUnit) super.visitCompilationUnit(cu, executionContext);
-            if(c != cu) {
+            if (c != cu) {
+                doAfterVisit(new ChangeType("org.junit.rules.TemporaryFolder", "java.io.File"));
                 maybeAddImport("java.io.File");
                 maybeAddImport("org.junit.jupiter.api.io.TempDir");
                 maybeRemoveImport("org.junit.ClassRule");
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDirTest.kt
```diff
@@ -235,10 +262,10 @@ class TemporaryFolderToTempDirTest : JavaRecipeTest {
                 File tempDir1;
             
                 File subDir = newFolder(tempDir1, "sub");
-                File subDirs = newFolder(tempDir1, "foo", "bar", "baz");
+                File subDirs = newFolder(tempDir1, "foo/bar/baz");
```
sambsnyd, line 265: What if the arguments aren't String literals but identifiers or method invocations?

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
```diff
@@ -93,187 +100,181 @@ protected TemporaryFolderToTempDirVisitor getVisitor() {
...
-            private final J.VariableDeclarations tempDirFields;
-
-            private ReplaceTemporaryFolderMethods(J.VariableDeclarations tempDirFields) {
-                this.tempDirFields = tempDirFields;
+        @Override
+        public J visitNewClass(J.NewClass newClass, ExecutionContext executionContext) {
+            J.NewClass nc = (J.NewClass) super.visitNewClass(newClass, executionContext);
+            JavaType.FullyQualified fq = TypeUtils.asFullyQualified(nc.getType());
+            if (fq != null && "org.junit.rules.TemporaryFolder".equals(fq.getFullyQualifiedName())) {
+                return null;
```
tkvangorder, line 111: not sure I understand this, are you trying to clear the initializer for a named variable here? 

This would also impact places where `new org.junit.rules.TemporaryFolder()` is used in code (which probably isn't very often)

src/test/kotlin/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDirTest.kt
```diff
@@ -117,16 +143,10 @@ class TemporaryFolderToTempDirTest : JavaRecipeTest {
                 @TempDir
                 File tempDir2;
             
-                File file2 = newFile(tempDir2, "sam");
+                File file2 = new File(tempDir2, "sam");
```
sambsnyd, line 146: 😬 good catch

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
```diff
@@ -93,187 +100,181 @@ protected TemporaryFolderToTempDirVisitor getVisitor() {
...
+                switch (mi.getSimpleName()) {
+                    case "newFile":
+                        return convertToNewFile(mi);
+                    case "newFolder":
+                        doAfterVisit(new AddNewFolderMethod(mi));
+                        break;
+                    case "create":
+                        return null;
+                    default:
+                        return mi.getSelect().withPrefix(Space.format(" "));
```
tkvangorder, line 130: why are we changing the select's prefix? when would we hit the default here ?

src/test/kotlin/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDirTest.kt
```diff
@@ -117,16 +143,10 @@ class TemporaryFolderToTempDirTest : JavaRecipeTest {
...
             
                 void foo() throws IOException {
-                    File file1 = File.createTempFile("junit", null, tempDir1);
-                }
-            
-                private static File newFile(File root, String fileName) throws IOException {
-                    File file = new File(root, fileName);
-                    file.createNewFile();
-                    return file;
+                    File file1 = new File(tempDir1, "junit");
```
sambsnyd, line 149: No no no this is bugs. `new File()` and `File.createTempFile()` do _not_ have the same meaning. 

```java
// a and b point to _different_ files that both exist
File a = File.createTempFile("junit", null, tempDir1);
File b = File.createTempFile("junit", null, tempDir1);

// c and d point to the same file which does not exist
File c = new File(tempDir1, "junit");
File d = new File(tempDir2, "junit")
```

sambsnyd (changes requested):

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
```diff
@@ -93,187 +100,181 @@ protected TemporaryFolderToTempDirVisitor getVisitor() {
...
+                cd = cd.withTemplate(template(
+                        "private static File newFolder(File root, String name) throws IOException {\n" +
+                                "    File result = new File(root, name);\n" +
+                                "    if(!result.mkdirs()) {\n" +
+                                "        throw new IOException(\"Couldn't create folders \" + root);\n" +
+                                "    }\n" +
+                                "    return result;\n" +
+                                "}"
+                ).imports("java.io.File", "java.io.IOException").javaParser(TEMPDIR_PARSER::get).build(), cd.getBody().getCoordinates().lastStatement());
+                maybeAddImport("java.io.File");
```
tkvangorder, line 239: After adding the method via the template, does the method have a MethodType (and is its declaring type correct)?

tkvangorder (changes requested):
Wow, this was a lot of work. See my comments.

src/test/kotlin/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDirTest.kt
pway99: nice catch @sambsnyd.  The subdirectory would reflect the identifier names.

src/test/kotlin/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDirTest.kt
pway99: that's a really good catch and explains why `File.createTempFile` should be used instead of `new `File(...)` 
thanks @sambsnyd.

src/test/kotlin/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDirTest.kt
```diff
@@ -260,32 +288,65 @@ class TemporaryFolderToTempDirTest : JavaRecipeTest {
...
                 public void newNamedFileIsCreatedUnderRootFolder() throws IOException {
                     final String fileName = "SampleFile.txt";
-                    @TempDir
-                    File tempFolder;
+                    final String otherFileName = "otherText.txt";
+                    TemporaryFolder tempFolder = new TemporaryFolder();
+                    TemporaryFolder tempFolder2 = new TemporaryFolder();
                     tempFolder.create();
-                    File f = newFile(tempFolder, fileName);
+                    tempFolder2.create();
```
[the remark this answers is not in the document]
pway99: the `create()` calls are not necessary the `tempFolder` is annotated with `@TempDir`.
A simple test can be used to confirm.
```java
@Test
void bar(@TempDir File tempDir) throws IOException {
	Assertions.assertNotNull(tempDir);
	File fnew = File.createTempFile("junit", null, tempDir);
	Assertions.assertTrue(fnew.exists());
}
```

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
pway99: Yes, the goal is to clear `TemporaryFolder` initializers, then annotate their associated variables with `@TempDir` and finally change `TemporaryFolder` types to `File`.

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
pway99: Thanks, @tkvangorder.  The `select` has an empty prefix which would result in `File root =tempDir;`.
Setting the prefix to a single space is fragile; it should actually be the prefix from the method invocation.
`return mi.getSelect().withPrefix(mi.getPrefix());`

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
pway99: I would assume that it does, can you elaborate on your concern?

tkvangorder (approved):
This look a lot better Pat.

pway99:
> This looks a lot better Pat.

Thanks @tkvangorder

src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
```diff
@@ -74,206 +78,144 @@ protected TemporaryFolderToTempDirVisitor getVisitor() {
...
-        private static class ReplaceTemporaryFolderMethods extends JavaVisitor<ExecutionContext> {
-
-            private final J.VariableDeclarations tempDirFields;
-
-            private ReplaceTemporaryFolderMethods(J.VariableDeclarations tempDirFields) {
-                this.tempDirFields = tempDirFields;
-            }
+        private boolean isRuleAnnotatedTemporaryFolder(J.VariableDeclarations vd) {
+            return vd.getTypeAsFullyQualified() != null
+                    && "org.junit.rules.TemporaryFolder".equals(vd.getTypeAsFullyQualified().getFullyQualifiedName())
```
sambsnyd, line 103: This could be simplified with `TypeUtils.isOfType()` or `TypeUtils.isOfClassType()`

pway99: thanks will make that change

PR #148, opened 2021-06-15
Use ChangeMethodAccessLevelVisitor

murdos:
See https://github.com/openrewrite/rewrite/issues/622


jkschneider:
Thanks @murdos! Test failures are unrelated to what you've done -- we're trying to tighten up the contract for recipes to leave the AST in a state with correct type attribution.

PR #150, opened 2021-06-17
TemporaryFolderToTempDir bug fix where the newly generated method invocations were not fully type attributed.

pway99:
TemporaryFolderToTempDir generates a method declaration for creating new folders and then generates method invocations for the new method declaration.  This fixes a bug where the newly generated method invocations were not fully type attributed.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
@@ -23,7 +23,9 @@ import org.openrewrite.java.*;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.*;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
```diff
@@ -209,10 +216,13 @@ public TranslateNewFolderMethodInvocation(J.MethodInvocation method) {
...
                         List<Object> templateArgs = new ArrayList<>(args);
                         templateArgs.add(0, tempDir);
-                        return mi.withTemplate(template(sb.toString())
+                        mi = mi.withTemplate(template(sb.toString())
                                         .imports("java.io.File").javaParser(TEMPDIR_PARSER::get).build(),
                                 mi.getCoordinates().replace(), templateArgs.toArray());
                     }
+                    mi = mi.withType(newMethodType);
+                    J.ClassDeclaration parentClass = getCursor().dropParentUntil(J.ClassDeclaration.class::isInstance).getValue();
+                    mi = mi.withName(mi.getName().withType(parentClass.getType()));
```
pway99, line 225: is this correct?

jkschneider: i don't understand the context enough to decide. do you think it is?

pway99: Yes, I do think it's correct.
 
This recipe adds a new method declaration, which is then used for generated method invocations.  Previously the generated method invocations would have null types.

This fix ensures that a subsequent visitor will encounter a fully type attributed method invocation.

sambsnyd: Looks correct to me

PR #154, opened 2021-06-24
Fix expected exception expect exception message string

### LINKED ISSUE #153: ExpectedException#expectMessage is not properly converted to JUnit 5
I think this test is expecting the wrong behavior: https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/test/kotlin/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrowsTest.kt#L162-L198

More specifically, this line in the source:
```java
thrown.expectMessage("Index 1 out of bounds for length 1");
```

should _not_ be converted to a string argument that's passed into the `assertThrows()` method like this:
```java
 }, "Index 1 out of bounds for length 1");
```
That string is only used as the failure message for `assertThrows()`, _not_ the expected exception's message.

Instead, it should be converted into an assertion like this:
```java
IndexOutOfBoundsException exception = assertThrows(IndexOutOfBoundsException.class, ...);
assertThat(exception.getMessage(), containsString("Index 1 out of bounds for length 1"));
```
It works correctly when the `expectMessage()` call is _already_ using the `containsString()` matcher, as shown here: https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/test/kotlin/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrowsTest.kt#L200-L238

pway99:
ExpectedExceptionToAssertThrows asserts the expected message string. Fixes #153

@skingsland, thanks again for identifying this issue.  The solution is slightly different from your suggestion because the `AssertToAssertions` does not assume AssertJ is on the classpath.

```java
public void statementsBeforeExpected() {
    Throwable exception = assertThrows(IndexOutOfBoundsException.class, () -> {
        int[] a = new int[]{1};
        int b = a[1];
    });
    assertTrue(exception.getMessage().contains("Index 1 out of bounds for length 1"));
}
```

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
@@ -170,17 +170,19 @@ public class ExpectedExceptionToAssertThrows extends Recipe {
                 }
             }
 
+            String exceptionDeclParam = ((isExpectArgAMatcher || isExpectMessageArgAMatcher || isExpectedCauseArgAMatcher)
+                    || expectMessageMethodInvocation != null) ?
+                    "Throwable exception =" : "";
+
             Object expectedExceptionParam = (expectMethodInvocation == null || isExpectArgAMatcher) ?
                     "Exception.class" : expectMethodInvocation.getArguments().get(0);
-
-            String exceptionDeclParam = ((isExpectArgAMatcher || isExpectMessageArgAMatcher || isExpectedCauseArgAMatcher) || expectMessageMethodInvocation != null) ? "Throwable exception =" : "";
 
             String templateString = expectedExceptionParam instanceof String ? "#{} assertThrows(#{}, () -> #{});" : "#{} assertThrows(#{any()}, () -> #{});";
 
             m = m.withTemplate(
                     template(templateString)
                             .javaParser(ASSERTIONS_PARSER::get)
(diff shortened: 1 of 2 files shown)


skingsland:
> @skingsland, thanks again for identifying this issue. The solution is slightly different from your suggestion by not converting to a `Hamcrest.matcherAssert`.
> 
> ```java
> public void statementsBeforeExpected() {
>     Throwable exception = assertThrows(IndexOutOfBoundsException.class, () -> {
>         int[] a = new int[]{1};
>         int b = a[1];
>     });
>     assertTrue(exception.getMessage().contains("Index 1 out of bounds for length 1"));
> }
> ```

Awesome, thanks so much for the quick turnaround on this bug! My group is currently in the process of upgrading _dozens_ of repos from JUnit 4 to 5 (as part of a Dropwizard 1.x -> 2.0 upgrade), and this will be a big help. What's your release cadence? (any idea when the next release will happen?)

pway99:
> Out of curiosity, what's the reason for using assertTrue() instead org.hamcrest.MatcherAssert.assertThat? Is it because you can't be sure that Hamcrest is on the classpath?

Yes, Using `jupiter.Assertions` does not assume that Hamcrest is on the classpath or is otherwise used.

jkschneider:
@skingsland We release ~1-2 weeks. We were planning a release of https://github.com/openrewrite/rewrite 7.8.0 today/tomorrow and then the rest of the recipe repos like this directly after.

PR #157, opened 2021-06-26
TestMethodsNotPublic recipe

murdos:
Remove `public` from JUnit5 `@Test` methods


jkschneider:
Polished slightly and merged with b8fc919050ed39c81561edc31ed8858426d433fd. We do need to make an enhancement to  the base `ChangeMethodAccessLevelVisitor` visitor to not reduce the visibility of methods that override higher visibility methods. But for a lot of projects this recipe will do the right thing.

PR #168, opened 2021-08-20
Fix for "Recipe validation error ..."

dkorotych:


jkschneider:
Good catch, thanks @dkorotych !

PR #178, opened 2021-10-15
TestsShouldNotBePublic: ignore overridden public test method

murdos:


murdos:
@aegershman : done :+1:

PR #185, opened 2022-01-26
Mockito 4

sullis:


jkschneider:
These particular dependencies need to be pinned to 3 because we are testing Mockito 2 to 3 changes, but you make me wonder now what breaking changes there are between Mockito 3 and 4!

jkschneider:
Maybe it helps to add more context. In `rewrite-spring` we have a [similar scenario](https://github.com/openrewrite/rewrite-spring/blob/main/build.gradle.kts#L212-L228) and actually create whole dependency configurations to isolate changes from one minor release to another. Then we can group the unit tests for say Mockito 2->3 in one source set that depends on Mockito 2 (the beginning state of a test to Mockito 3), and another source set that depends on Mockito 3 (the beginning state of a test to Mockito 4)

pway99:
added a issue https://github.com/openrewrite/rewrite-testing-frameworks/issues/186 for tracking the mockito 3 to 4 migration

tkvangorder:
Closing this as we are tracking the mockito 4 with #186

PR #240, opened 2022-07-23
Remove test prefix

### LINKED ISSUE #237: Strip `test` prefix from test method names
Inspired by [this thread](https://twitter.com/maciejwalkowiak/status/1549332873219657730)
> Don't use "test" prefix for test methods. This had to be done in JUnit 3 (!) but now it only adds noise. Instead use method names that express the intent.

Seems to me OpenRewrite would be well suited to strip that prefix, and it should nudge developers to rethink their test method naming strategy going forward.

Some things to consider:
1. remove `test` prefix only where (meta) annotated with `@Test`, `@ParameterizedTest` or `@RepeatedTest` (or other `@TestTemplate` annotations
2. probably lowercase the first character after `test` prefix
3. Handle edge cases such as `@Test void test1() {...}` and `@Test void test() {...}` correctly

timtebeek:
Fixes #237.

Wasn't sure if I should reference this recipe somewhere in `/rewrite-testing-frameworks/src/main/resources/META-INF/rewrite`, or if this is even something you would consider adding. So here's a proposal that should work as is, to facilitate further discussion around this subject.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ build/
 .gradle/
 out/
 .idea/
+.classpath
+.project
+.settings/
+bin/
--- a/src/main/java/org/openrewrite/java/testing/cleanup/RemoveTestPrefix.java
+++ b/src/main/java/org/openrewrite/java/testing/cleanup/RemoveTestPrefix.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


pway99:
Thanks @timtebeek, I went ahead and added a reference `RemoveTestPrefix` reference in the `org.openrewrite.java.testing.cleanup.BestPractices` along with a single source applicability test.  We can run BestPractices on the SaaS, if this one creates too much noise then we can just remove the reference.  

345ebfe62951455016758ffea6122a8c01c2413c
970a677aed134af0220866ae8fc37ba7f4cbd80b
14877edd3c39c21096897be9b11c83c03b26de8a

timtebeek:
Perfect! Thanks for considering, improving & merging this one.
Should be an interesting nudge to change naming conventions. :)

timtebeek:
Tried it just now an it works well; here's how I applied it to Maven projects:
> ./mvnw -U org.openrewrite.maven:rewrite-maven-plugin:4.30.0:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:1.25.0 -DactiveRecipes=org.openrewrite.java.testing.cleanup.RemoveTestPrefix

pway99:
Thats fantastic news;  Thanks Tim!

PR #246, opened 2022-08-09
Do not strip prefix when remainder results in methods such as toString()

timtebeek:
Came across this when migrating a large repository; potential source of errors to avoid.


tkvangorder (approved):
Thank you! Nice catch

PR #247, opened 2022-08-15
Strip test_ prefix from snake cased methods

timtebeek:
Small addition after exposure to new code base.


timtebeek:
Failure seems unrelated:
> JUnitAssertTrueToAssertThatTest > singleStaticMethodNoMessage() FAILED

pway99:
Thanks @timtebeek, 
I added another check to prevent renaming a method if there is another method having the same name as the new name. 
feb241ae86230232b64a3dd00131f63779cdeb67

timtebeek:
Ah perfect, thanks! That's one of the issues I ran into as well; good to have that covered.

pway99:
Thanks for your effort to keep improving this one :)

PR #250, opened 2022-08-19
Improved JUnit AssertToAssertions behavior with missing type info

nmck257:
Depends on https://github.com/openrewrite/rewrite/pull/2160

Apps which use Lombok are especially vulnerable to issues with the JUnit4 to 5 Migration recipe, as they commonly use Lombok-generated getters within `assert` statements, which breaks type information for the whole line and causes those `assert` invocations to "hide" from the recipe.

This PR improves the migration's resiliency in those cases by allowing matching on unknown types.


nmck257:
(build failure is expected until the pre-requisite PR is merged into rewrite)

Issue #264, opened 2022-10-03
Further cucumber-jvm recipes and improvements


timtebeek:
A tracking issue for a few more cucumber-jvm recipes, to get these out of my head and into a GitHub Issue.
Each individually should be easy enough to contribute and check off here; whenever I or anyone finds the time.

- [x] Replace needless-regex cucumber expressions with non-regex cucumber expressions
  - `@When("^no need for a regex here$")` -> `@When("no need for a regex here")`
  - possibly even convert regexes with capturing groups to instead use `{}`
- [x] Convert [once upon a time](https://github.com/cucumber/cucumber-jvm/blob/main/release-notes/v5-CHANGELOG.md#changed-6) `cucumber.api` package to `io.cucumber`
- [x] Replace [old `info.cukes` groupIds](https://stackoverflow.com/a/50212219) with `io.cucumber`
- [x] Replace [`@Cucumber` with `@Suite @SelectClasspathResource("some/path")`](https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/Cucumber.java#L26)
- [x] Drop `SummaryPrinter` marker interface; replace with `Plugin` interface if not already present
- [ ] Support method references in the cucumber-java8 to cucumber-java migration recipes
- [ ] Migrate `cucumber-java8` datatables to annotation based step definitions

Tagging @mlvandijk & @mpkorstanje for review & further suggestions.

timtebeek:
Captured both outstanding items as new feature requests on:
https://github.com/cucumber/cucumber-jvm-migration/issues/2
https://github.com/cucumber/cucumber-jvm-migration/issues/3

PR #266, opened 2022-10-10
Drop SummaryPrinter #264

timtebeek:
@mpkorstanje do you agree with the samples & replacements here? Should we cover any other cases?
The deprecation notes mentioned EventListeners, but might require further changes.


mpkorstanje (approved):
Replacements look good to me.

pway99 (approved):
Thanks @timtebeek

PR #268, opened 2022-10-13
Cucumber annotation to suite

timtebeek:
Not sure why I can't replace a single annotation with two annotations from a single template like this. 🤔

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -125,6 +125,8 @@ dependencies {
     implementation("org.openrewrite:rewrite-maven:$rewriteVersion")
     runtimeOnly("com.fasterxml.jackson.core:jackson-core:2.12.+")
     runtimeOnly("org.openrewrite:rewrite-java-17:$rewriteVersion")
+    runtimeOnly("io.cucumber:cucumber-junit-platform-engine:7.+")
+    runtimeOnly("org.junit.platform:junit-platform-suite-api:latest.release")
 
     compileOnly("org.projectlombok:lombok:latest.release")
     annotationProcessor("org.projectlombok:lombok:latest.release")
--- a/src/main/java/org/openrewrite/java/testing/cucumber/CucumberAnnotationToSuite.java
+++ b/src/main/java/org/openrewrite/java/testing/cucumber/CucumberAnnotationToSuite.java
@@ -1,0 +1,74 @@
+package org.openrewrite.java.testing.cucumber;
+
+import java.time.Duration;
+import java.util.List;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.*;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.J.ClassDeclaration;
+
+public class CucumberAnnotationToSuite extends Recipe {
+
+    private static final String IO_CUCUMBER_JUNIT_PLATFORM_ENGINE_CUCUMBER = "io.cucumber.junit.platform.engine.Cucumber";
+    private static final AnnotationMatcher ANNOTATION_MATCHER = new AnnotationMatcher(
+            IO_CUCUMBER_JUNIT_PLATFORM_ENGINE_CUCUMBER);
+
+    private static final String SUITE = "org.junit.platform.suite.api.Suite";
+    private static final String SELECT_CLASSPATH_RESOURCE = "org.junit.platform.suite.api.SelectClasspathResource";
+
+    @Override
+    public String getDisplayName() {
+        return "Replace @Cucumber with @Suite";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replace @Cucumber with @Suite and @SelectClasspathResource(\"cucumber/annotated/class/package\").";
+    }
+
+    @Override
+    public @Nullable Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(2);
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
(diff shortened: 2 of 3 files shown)


build.gradle.kts
```diff
@@ -125,6 +125,8 @@ dependencies {
     implementation("org.openrewrite:rewrite-maven:$rewriteVersion")
     runtimeOnly("com.fasterxml.jackson.core:jackson-core:2.12.+")
     runtimeOnly("org.openrewrite:rewrite-java-17:$rewriteVersion")
+    runtimeOnly("io.cucumber:cucumber-junit-platform-engine:7.+")
+    runtimeOnly("org.junit.platform:junit-platform-suite-api:latest.release")
```
pway99, line 129: In general dependencies necessary for recipe testing should belong to the `testRuntimeOnly` group.

src/main/java/org/openrewrite/java/testing/cucumber/CucumberAnnotationToSuite.java
```diff
@@ -0,0 +1,74 @@
...
+                maybeRemoveImport(IO_CUCUMBER_JUNIT_PLATFORM_ENGINE_CUCUMBER);
+                maybeAddImport(SUITE);
+                maybeAddImport(SELECT_CLASSPATH_RESOURCE);
+
+                return classDecl.withLeadingAnnotations(ListUtils.flatMap(classDecl.getLeadingAnnotations(), ann -> {
+                    String code = "@Suite\n@SelectClasspathResource(\"#{}\")";
+                    String path = classDecl.getType().getPackageName().replace('.', '/');
+                    JavaTemplate template = JavaTemplate.builder(this::getCursor, code)
+                            .javaParser(
+                                    () -> JavaParser.fromJavaVersion().classpath("junit-platform-suite-api").build())
```
pway99, line 64: can it be assumed that any project using the `io.cucumber.platform` will also have a `junit-platform-suite-api` dependency?  If not then adding the dependency and using a stub for the template would be a better option

timtebeek: In [the latest junit5 example](https://github.com/cucumber/cucumber-jvm/blob/main/examples/calculator-java-junit5/pom.xml#L52) I think we recommend people add both dependencies; is that still correct @mpkorstanje?

timtebeek: Didn't want to push any conflicts in case you're working through those double annotation template issues.
Am I correct in thinking the stubs would be something along these lines?
```java
.javaParser(() -> JavaParser.fromJavaVersion().dependsOn(
        "package org.junit.platform.suite.api;\npublic @interface Suite {}",
        "package org.junit.platform.suite.api;\npublic @interface SelectClasspathResource {\nString value();\n}")
    .build())
```
And that would then mean we can remove the build.gradle dependency, while adding `junit-platform-suite-api` as an additional step?

mpkorstanje: The two depencies that are needed `cucumber-junit-platform-engine` and `junit-platform-suite`. Note this isn't the same as the `junit-platform-suite-api`.

The explanation as to why these two depencies is a bit complicated.

pway99: @timtebeek your JavaParser stubs look correct

pway99: thanks @mpkorstanje and @timtebeek, I will fix up the JavaTemplate JavaParser and  dependencies, then push the changes :)

pway99:
closed as completed by 50ba4492368200480b9e13e665e7a307e60ceb55

pway99:
@timtebeek, when you have a chance it would be a good idea to add a `maven project` to the test, I ran out of steam today and did not fully track down the inital maven project configuration
```java
@Test
void should_replace_cucumber_annotation_with_suite_with_selected_classpath_resource() {
rewriteRun(
        mavenProject("project",
                srcTestJava(
                        java("""
                            package com.example.app;
            
                            import io.cucumber.junit.platform.engine.Cucumber;
            
                            @Cucumber
                            public class CucumberJava8Definitions {
                            }
                            """,
                            """
                            package com.example.app;
            
                            import org.junit.platform.suite.api.SelectClasspathResource;
                            import org.junit.platform.suite.api.Suite;
            
                            @Suite
                            @SelectClasspathResource("com/example/app")
                            public class CucumberJava8Definitions {
                            }
                            """)
                ),
                pomXml(
                        """
                        <project>
                            <groupId>com.mycompany.app</groupId>
                            <artifactId>my-app</artifactId>
                            <version>1</version>
                        </project>
                        <dependencies>
                            ...
                        </dependencies>
                        """,
                        """
                        <project>
                            <groupId>com.mycompany.app</groupId>
                            <artifactId>my-app</artifactId>
                            <version>1</version>
                        </project>
                        <dependencies>
                            <dependency>
                                <groupId>org.junit.platform</groupId>
                                <artifactId>junit-platform-suite</artifactId>
                                <version>1.9.1</version>
                            <dependency>
                        </dependencies>
                        """)
        )
);
}
```

pway99:
Hello @timtebeek and @mpkorstanje,
I had made an error in suggesting that the cucumber dependencies should be moved to the `testRuntimeOnly` class path and the `JavaParser` class-path references replaced with stubs.  If you're interested here are the changes 5cd92db447ba875de0b21ea4d9b0dec045d9e086 changes.  Thanks again for your help on this one :)

pway99:
On second thought, I decided to remove the maven changes from the test to avoid future test failures after the next `junit-platform-suite` release. 59c9c57a9285bb780beaae682742bbc8476552a1

timtebeek:
Awesome; thanks a lot for the help, especially throughout the various iterations around testing.
I could see the maven project tests indeed require frequent upgrades if we had left those in.
Perhaps in the future we can look at some form of version matching to not have to update those.

That makes https://github.com/openrewrite/rewrite-testing-frameworks/issues/264 ready for a write up; I'll try to squeeze in both a blog and some release notes for Cucumber-JVM.

PR #274, opened 2022-10-20
Should not use equals when comparing object identity

### LINKED ISSUE #273: AssertTrueComparisonToAssertEquals is not appropriate when comparing objects.
The following example: 

```
// both page must have the same bean instance
assertTrue(page.getBean() == notRequiredpage.getBean());
```

Is converted to the following and these are not equivalent:

```
assertEquals(page.getBean(), notRequiredpage.getBean());
``` 

The original is comparing object reference equality whereas the conversion is calling `Object.equals()`. 

This recipe should probably not make this conversion if the J.Binary arguments are not primitives.

vandeven:
fixes #273


tkvangorder (approved):
Thank you for your contribution!

PR #290, opened 2022-12-08
JUnit 5 migration now removes surefire-junit plugins

nmck257:
depends on just-pushed new recipe in rewrite; snapshot builds might not be caught up yet


traceyyoshima:
Hi @nmck257! Thanks for the PR! Would you please rebase this, thanks in advance!

nmck257:
@traceyyoshima rebased

PR #303, opened 2023-01-24
Fixed NPE in TestsShouldNotBePublic

nmck257:


nmck257:
would be cool if we could turn nullability warnings into build failures -- IDEA warned about this line, but it still got merged

PR #307, opened 2023-01-31
Add BDDMockito methods to the MOCKITO_METHOD_NAMES

rrvk:
When i tried this recipe on my project there where soms imports that where removed while stil in use. That is because i used given() and this is in mockito core (BDDMockito) and those where not in this recipe that is why i made this pr

See here for the BDDMockito java doc
https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/BDDMockito.html

I appriciate the feedback, thanks for this project


sambsnyd:
Thanks for your contribution @rrvk! Appreciate any and all bug fixes

Issue #293, opened 2022-12-20
Recipe - Replace PowerMockito with raw Mockito


nmck257:
Many apps in the wild use PowerMockito to provide static mocking capabilities to Mockito in JUnit 4. 
https://github.com/powermock/powermock/wiki/Mockito

But, Mockito 3.4.0+ can mock statics natively:
https://www.baeldung.com/mockito-mock-static-methods
And PowerMock does not have a version compatible with JUnit 5 (and perhaps never will):
https://github.com/powermock/powermock/issues/929

Developers would likely benefit from a recipe which takes an app using Powermock+Mockito and ~~burns all of the static mocking to the ground~~ replaces usages of PowerMock with equivalent Mockito native syntax (this new recipe could also be called from the JUnit5 recipe)

tkvangorder:
It would be helpful to have specific examples of the PowerMock API and then an equivalent in Mockito 3.4+.

klauerm:
I've recently migrated a huge code base from Mockito 1 to Mockito 4. Unfortunately some migrated modules are using PowerMockito to mock static methods. Those tests are failing after the migration. We urgently need such a recipe. So I've forked this repository and try to write it on my own. I will raise a pull request as soon as I'm finished.

klauerm:
Hi @tkvangorder,
I've drafted the first example what the recipe shall do in a unit test. I've also tried to do some changes by using existing recipes (see mockito.yml) but these are not working.

klauerm:
Continue development in PR  #312 . The integration test src/test/java/org/openrewrite/java/testing/mockito/ReplacePowerMockitoIntegrationTest.java holds all the requirements for the recipe. Can someone please look over and see if something is missing.

PR #312, opened 2023-02-09
Improvement of Replace PowerMockito with Mockito static mocking

klauerm:
Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/293


klauerm:
Hi @sambsnyd @timtebeek,
Can you have a look. The integration test src/test/java/org/openrewrite/java/testing/mockito/ReplacePowerMockitoIntegrationTest.java holds all the requirement we need for our code base. But it is still failing. I don't know how to fix the issue I mentioned in [slack](https://rewriteoss.slack.com/archives/C01A843MWG5/p1675949738405979)

klauerm:
@timtebeek  I've fixed all tests except for 1 in src/test/java/org/openrewrite/java/testing/mockito/ReplacePowerMockitoIntegrationTest.java
Here the feature for using testng annotations, if the mockito tests are written with testng is still missng. Can you please activate this integration test for the pipeline again?
Thank you, Matthias

timtebeek:
As discussed on Slack
> ReplacePowerMockitoIntegrationTest > testThatPowerMockitoIsReplacedInTestNGTests() FAILED
> it seems to use `@AfterEach` instead of `@AfterMethod`
> I think [the `maybeAddTearDownMethodBody` function](https://github.com/klauerm/rewrite-testing-frameworks/blob/feature/replace-power-mockito-recipe/src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoMockStaticToMockito.java#L168) either needs to be aware of TestNG, and then add the corresponding AfterEach/AfterMethod annotation, or needs a twin to handle TestNG tear down.. now it seems specific to JUnit, and I'm not seeing an equivalent for TestNG

sambsnyd:
I haven't reviewed this in great detail, so there may be suggestions or improvements to come later, but for now we can merge and iterate on it

PR #320, opened 2023-03-02
Replace Mockito 1.x any(Class) and anyString() with nullable(Class)

### LINKED ISSUE #319: Consider logic change of mockito argument matchers
From Mockito 1.x to 3.x the logic of the argument matchers has changed. Mockito.any(<Class>) doesn't match a null value that is passed to the mocked method as parameter. (see https://site.mockito.org/javadoc/current/org/mockito/ArgumentMatchers.html and https://javadoc.io/static/org.mockito/mockito-core/5.1.1/org/mockito/ArgumentMatchers.html#nullable(java.lang.Class))

Therefore, it is necessary to replace all instances of Mockito.anyString() and any(<Class>) with nullable(String) or nullable(<Class>). Nullable behaves exactly like anyString() or any(<Class>) in the previous Mockito version 1.x.

timtebeek:
Fixes #319


src/main/java/org/openrewrite/java/testing/mockito/AnyStringToNullable.java
```diff
@@ -15,24 +15,20 @@
...
 
 /**
- * S
+ * Replace Mockito 1.x `anyString()` with `nullable(String.class)`
  */
 public class AnyStringToNullable extends Recipe {
-
-    public static final String ARGUMENT_MATCHERS_ANY_STRING = "org.mockito.Mockito anyString()";
+    private static final MethodMatcher ANY_STRING = new MethodMatcher("org.mockito.ArgumentMatchers anyString()");
+    private static final String MOCKITO_CLASS_PATH = "mockito-core-3.12.4";
```
timtebeek, line 31: This recipe is intended to be used only when migrating from Mockito 1.x to 2.10 and above, as that's when the behavior surrounding the nullability of `anyString()/any(Class)` changed. We do not want to migrate usages of `anyString()/any(Class)` from Mockito 3+ to `nullable(Class)`, as that would change the semantics again.

Given that context, would you then agree with the following change?
```suggestion
    private static final MethodMatcher ANY_STRING = new MethodMatcher("org.mockito.Mockito anyString()");
    private static final String MOCKITO_CLASS_PATH = "mockito-all-1.10.19";
```

I think since we match on `Mockito` rather than `ArgumentMatchers`, newer code should be reasonable safe. Would you agree to that as well?

src/main/resources/META-INF/rewrite/mockito.yml
```diff
@@ -121,6 +122,30 @@ recipeList:
...
+  - testing
+  - mockito
+applicability:
+  anySource:
+    - org.openrewrite.maven.search.FindDependency:
+        groupId: org.mockito
+        artifactId: mockito-all
+#    - org.openrewrite.gradle.search.FindDependency:
+#        groupId: org.mockito
+#        artifactId: mockito-all
```
timtebeek, line 138: Curiously, when I enable this, AnyToNullableTest starts to fail, even though it's [within an anySource block](https://docs.openrewrite.org/reference/yaml-format-reference) 🤔 
> If the anySource list is specified, OpenRewrite will run every recipe in the anySource list on every file in the repository the recipe is being run on. If any of those recipes result in a change to at least one file, then the declarative recipe is considered "applicable" for that repository and the recipes specified in the recipeList will be run on every file in the repository.

```suggestion
    - org.openrewrite.gradle.search.FindDependency:
        groupId: org.mockito
        artifactId: mockito-all
```
Any idea why @kunli2 ?

timtebeek: We can of course limit it to what it is now, and only perform these recipes for Maven.
What I hope to prevent with these applicability tests is that we do not migrate calls from 2.x or 3.x to `nullable`, as we should not; only when we migrate from 1.x to 2.10+ should we migrate these calls, as per #319

timtebeek:
Kun Li also reported a possible workaround; captured here in case we stick to `AND` and update the docs to reflect that:
```yaml
---
type: specs.openrewrite.org/v1beta/recipe
name: org.openrewrite.java.testing.mockito.AnyToNullableApplicabilityCheck
displayName: todo
description: todo
tags:
  - testing
  - mockito
recipeList:
  - org.openrewrite.maven.search.FindDependency:
      groupId: org.mockito
      artifactId: mockito-all
  - org.openrewrite.gradle.search.FindDependency:
      groupId: org.mockito
      artifactId: mockito-all
---
type: specs.openrewrite.org/v1beta/recipe
name: org.openrewrite.java.testing.mockito.AnyToNullable
displayName: Replace Mockito 1.x `any(Class)` and `anyString()` with `nullable(Class)`
description: Since Mockito 2.10 `any(Class)` and `anyString()` no longer match null values. Use `nullable(Class)` instead.
tags:
  - testing
  - mockito
applicability:
  anySource:
    - org.openrewrite.java.testing.mockito.AnyToNullableApplicabilityCheck
recipeList:
  - org.openrewrite.java.ChangeMethodName:
      methodPattern: org.mockito.Mockito any(java.lang.Class)
      newMethodName: nullable
  - org.openrewrite.java.ChangeMethodTargetToStatic:
      methodPattern: org.mockito.Mockito nullable(java.lang.Class)
      fullyQualifiedTargetTypeName: org.mockito.ArgumentMatchers
  - org.openrewrite.java.testing.mockito.AnyStringToNullable
---
```

shanman190:
I think this may still yield some missing cases as you're just checking to see if the dependency is there. Wouldn't it be better to check that the type or method was in use instead? That'd be a lot more exact and no longer be dependent upon GAV coordinates (since you could pull in the smaller mockito dependencies and desire this recipe to run for that case).

timtebeek:
Thanks for chiming in! Appreciate your concern to be exact. The issue here is there's a semantic change to the same methods `any(Class)` and `anyString()` when going from 1.x to 2.10+. Their signatures stay the same, but their handling of `null` changes (no longer allowed). I don't know if there's another way to detect that, but I'm open to suggestions.

On `main` currently we're sure to break those assumptions about `null` handling, by not changing anything at all.
With this change I'm hoping to selectively change to `nullable` methods where we can detect that we should (exclusively when going from 1.x to 2.x), while avoiding we change any methods where we should not (anything after 2.10).

Still missing some cases with this PR as-is might then still be preferred over doing nothing, and strongly preferred over incorrectly changing anything when users are already on 2.10+.

How would you feel we should proceed here given the above?

shanman190:
So this one is going to be difficult for a few reasons.

* We're won't be able to rely upon the LST itself because of the case of if the upgrade mockito dependency came first
* As you said, you only want to apply this recipe IF AND ONLY IF, the project was on a Mockito <2.10.

The only way that I can think of to achieve this would be to use the `visit(List<SourceFile>, ExecutionContext)` mechanic. We would then use an internal visitor to visit the Maven and Gradle dependency descriptors and interrogate the markers to see what mockito version was present at the time the LST was generated, then if we find out `true` case, we would schedule the other `Recipe` instances via `doNext(Recipe)` or `doNext(Visitor)`.

kunli2 (approved):

src/test/java/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.java
```diff
@@ -141,14 +141,38 @@ static class Foo {
...
+                  static class Foo {
+                      boolean bool(String str, int i, Object obj) { return false; }
+                      int in(boolean b, List<String> strs) { return 0; }
+                      int bar(byte[] bytes, String[] s, int i) { return 0; }
+                      boolean baz(String ... strings) { return true; }
+                  }
+
+                  public void usesMatchers() {
+                      Foo mockFoo = mock(Foo.class);
+                      when(mockFoo.bool(nullable(String.class), anyInt(), nullable(Object.class))).thenReturn(true);
```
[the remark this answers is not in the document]
timtebeek: This change is now not in the final change set right? It's not desired there, as we don't want to migrate when it's not explicitly from 1.x to anything above 2.10+.

sambsnyd:
Thanks @timtebeek. Good ideas @shanman190 , would be happy to see those improvements made

PR #324, opened 2023-03-22
verifyNoInteractions should not be removed

tmontovaneli:
By running the recipe, statements importing this method are removed.

This method still is in [Mockito](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#verifyNoInteractions(java.lang.Object...)) and is not deprecated.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/CleanupMockitoImports.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/CleanupMockitoImports.java
@@ -62,7 +62,7 @@ public class CleanupMockitoImports extends Recipe {
 
     public static class CleanupMockitoImportsVisitor extends JavaIsoVisitor<ExecutionContext> {
         private static final List<String> MOCKITO_METHOD_NAMES = Arrays.asList("mock", "mockingDetails", "spy",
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks a lot for taking the time not only to report the issue, but immediately provide a fix as well. Much appreciated!
We typically release every two to three weeks; until then you can already try this out through our [snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions).

harsh-pamnani:
Should we also include `clearInvocations` method along with it? We faced issues with that method as well and it is _not_ [deprecated in Mockito](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#clearInvocations(T...)).

timtebeek:
@harsh-pamnani Thanks for chiming in! Added that in https://github.com/openrewrite/rewrite-testing-frameworks/commit/a6829b2b5792599eee188e8d12d363690f77e473

harsh-pamnani:
Awesome. Thanks @timtebeek

timtebeek:
No problem; thanks for reporting these two fixes! Hope you enjoy the updated Mockito tests. :)

And any more reports & fixes always welcome!

PR #330, opened 2023-04-15
Support member references in AssertJ assertThrows

Meijuh:
This PR adds support for rewriting JUnit assertions to AssertJ assertions for throwing member references.
Lambdas in the form of `() -> { throw new Exception(); }` were already supported. Now `var::throwingMemberReference` is  also supported.

I'm not really sure what rewrite's default behavior is for unsupported semantics; the added code simply does not rewrite unsupported semantics. Alternatively, the PR can be adjusted to throw an `IllegalArgumentException` in the last if-else branch.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
@@ -67,23 +67,34 @@ public class JUnitAssertThrowsToAssertExceptionType extends Recipe {
         }
         private static final MethodMatcher ASSERT_THROWS_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions assertThrows(..)");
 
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Clear improvement to cover a case we had not identified before; thanks a lot!

timtebeek:
> I'm not really sure what rewrite's default behavior is for unsupported semantics; the added code simply does not rewrite unsupported semantics. Alternatively, the PR can be adjusted to throw an `IllegalArgumentException` in the last if-else branch.

We strive to "[do no harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm)", as you might have discovered in the docs already; that often means not making a change is preferred over making a partial change. Sometimes we add in a comment when we can not convert a particular case where conversion is necessary; in this case I think it's fine to leave in the original statement, as is implied by the fall through `return mi;`.

Throwing an `IllegalArgumentException` is typically undesirable; we want to allow users to confidently run recipes against large code bases; uncovered cases are ideally then best not converted rather than breaking a recipe run that might be chained into a larger composite recipe.

Thanks for bringing this case to our attention, and providing an immediate fix as well!

PR #331, opened 2023-04-15
Add test that generated invalid code

Meijuh:
This PR adds a unit test `assertThrowsAssignment` that generates invalid code. In JUnit `assertThrows` returns the exception that is thrown, while in AssertJ there is no way (currently) to obtain the exception that is thrown.

The purpose of this PR is check whether it is desired behavior open rewrite generates invalid Java code. A solution would be to skip rewriting cases where `assertThrows` is used as RHS in an assignment, but that would leave JUnit assertions in code bases.

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionTypeTest.java
+++ b/src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionTypeTest.java
@@ -118,4 +118,30 @@ class JUnitAssertThrowsToAssertExceptionTypeTest implements RewriteTest {
           )
         );
     }
+
+    /**
+     * A degenerate case showing we need to make sure the <code>assertThrows</code> appears
+     * immediately inside a J.Block.
+     */
+    @Test
+    void assertThrowsTernaryAssignment() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import static org.junit.jupiter.api.Assertions.assertThrows;
+
+              public class SimpleExpectedExceptionTest {
+                  public void throwsExceptionWithSpecificType() {
+                      NullPointerException npe = hashCode() == 42
+                        ? new NullPointerException()
+                        : assertThrows(NullPointerException.class, () -> {
+                          throw new NullPointerException();
+                      });
+                  }
+              }
+              """
+          )
+        );
+    }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionTypeTest.java
```diff
@@ -63,4 +63,36 @@ public void throwsExceptionWithSpecificType() {
...
+              import static org.junit.jupiter.api.Assertions.assertThrows;
+              
+              public class SimpleExpectedExceptionTest {
+                  public void throwsExceptionWithSpecificType() {
+                      NullPointerException npe = assertThrows(NullPointerException.class, () -> {
+                          throw new NullPointerException();
+                      });
+                  }
+              }
+
```
timtebeek, line 93: It looks like you want to assert no change is made; in such a case you only have to pass the first argument.
```suggestion
```

timtebeek:
Thanks for adding a test to reproduce an issue! Indeed best to not convert such cases if there's no direct replacement in AssertJ. If you add [`@ExpectedToFail`](https://junit-pioneer.org/docs/expected-to-fail-tests/) we can already merge your PR as is; or if you'd like to attempt to fix the issue too, you're more than welcome to have a go. Whatever you prefer.

Meijuh:
I'll push a fix that leaves such a case as is.  Shouldn't be too hard to check if the assertThrows appears on the RHS of an assignment.

Meijuh:
Instead of blacklisting the case where `assertThrows` appears on the RHS of an assignment, I decided to whitelist the case where `assertThrows` appears immediately inside a `J.Block`. This is to ensure the degenerate test case I added (and more like it) would not fail.

Also, I am unsure what exactly the code style is, where can I find it? Is the style I used okay?

timtebeek:
> Instead of blacklisting the case where `assertThrows` appears on the RHS of an assignment, I decided to whitelist the case where `assertThrows` appears immediately inside a `J.Block`. This is to ensure the degenerate test case I added (and more like it) would not fail.

Looks like a fine addition to me; thanks for taking the time to think through other limitations as well, and providing a fix.

> Also, I am unsure what exactly the code style is, where can I find it? Is the style I used okay?

We typically use the IntelliJ default, sometimes supplemented with `.editorconfig`.

I like the changes you've made! My only change is adding the Issue annotation with links here, such that we can refer to this discussion.

timtebeek (approved):

timtebeek:
Thanks a lot! While we wait for the for the next release in a week or two you can already [use our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions).

timtebeek:
- Support here has been extended through https://github.com/openrewrite/rewrite-testing-frameworks/pull/702, after AssertJ added an `actual()` method.

PR #337, opened 2023-05-13
Add JUnit ThrowingRunnableToExecutable recipe

peter-janssen:
Recently migrated using openrewrite and was missing this recipe. Possibly not as simple as a ChangeType recipe.


knutwannheden (approved):
LGTM. Thanks for contributing!

PR #338, opened 2023-05-14
Fix missing types in JUnit5BestPracticesTest

timtebeek:
To prevent propagation of undesirable patterns with the addition of new tests, as seen in #337

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/JUnit5BestPracticesTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/JUnit5BestPracticesTest.java
@@ -22,7 +22,6 @@ import org.openrewrite.config.Environment;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
-import org.openrewrite.test.TypeValidation;
 
 import static org.openrewrite.java.Assertions.java;
 
@@ -32,7 +31,7 @@ class JUnit5BestPracticesTest implements RewriteTest {
     public void defaults(RecipeSpec spec) {
         spec
           .parser(JavaParser.fromJavaVersion()
-            .classpathFromResources(new InMemoryExecutionContext(), "junit-4.13.+"))
+            .classpathFromResources(new InMemoryExecutionContext(), "junit-4.13.+", "junit-jupiter-api-5.9.2"))
           .recipe(Environment.builder()
             .scanRuntimeClasspath("org.openrewrite.java.testing")
             .build()
@@ -152,13 +151,12 @@ class JUnit5BestPracticesTest implements RewriteTest {
     void changeIgnoreToDisabled() {
         //language=java
         rewriteRun(
-          spec -> spec.typeValidationOptions(TypeValidation.none()),
           java(
             """
               import org.junit.Ignore;
               import org.junit.Test;
 
-              public class Example {
+              public class ExampleTest {
                   @Ignore
                   @Test
                   public void something() {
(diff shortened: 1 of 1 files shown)


src/test/java/org/openrewrite/java/testing/junit5/JUnit5BestPracticesTest.java
```diff
@@ -194,27 +192,28 @@ void somethingElse() {
     void changeThrowingRunnableToExecutable() {
         //language=java
         rewriteRun(
-          spec -> spec.typeValidationOptions(TypeValidation.none()),
```
timtebeek, line 197: This is what prompted the change; the rest then came along after seeing failure with 5.9.2.

src/main/java/org/openrewrite/java/testing/cleanup/AssertEqualsNullToAssertNull.java
```diff
@@ -54,7 +54,7 @@ protected JavaVisitor<ExecutionContext> getVisitor() {
             private Supplier<JavaParser> javaParser(ExecutionContext ctx) {
                 if(javaParser == null) {
                     javaParser = () -> JavaParser.fromJavaVersion()
-                            .classpathFromResources(ctx, "junit-jupiter-api-5.9.2")
+                            .classpathFromResources(ctx, "junit-jupiter-api-5.9.+")
```
knutwannheden, line 57: I am unsure about this. Looking at the implementation it looks like the string here will be interpreted as the prefix in a regex, where both `.` and `+` have a special meaning. I still don't understand why the builds fail in CI but not locally.

src/test/java/org/openrewrite/java/testing/junit5/JUnit5BestPracticesTest.java
knutwannheden: The `5\\.9` was just a test from my side trying to figure this out.

src/main/java/org/openrewrite/java/testing/cleanup/AssertEqualsNullToAssertNull.java
timtebeek: Hmm; yes we seem to be inconsistent on the main branch when you compare the search results:
https://github.com/search?q=repo%3Aopenrewrite%2Frewrite-testing-frameworks%205.9.2&type=code
https://github.com/search?q=repo%3Aopenrewrite%2Frewrite-testing-frameworks+5.9.%2B&type=code

I picked `5.9.+` as it was already present often, and to pick up the new 5.9.3 that was released three weeks ago.

We can standardize on something else in this PR if that's preferred, although `.+` without escaped dots is pretty common throughout our repositories already. I'd assumed that to be an [internal version selector](https://docs.openrewrite.org/reference/dependency-version-selectors), but I guess it's indeed a regex as we don't support `.+`, only `.x` and others according to the docs.

src/test/java/org/openrewrite/java/testing/mockito/ReplacePowerMockitoIntegrationTest.java
```diff
@@ -34,11 +34,11 @@ public void defaults(RecipeSpec spec) {
...
               "mockito-core-3.12.4",
-              "junit-jupiter-api-5.9.2",
+              "junit-jupiter-api-5.9.3",
               "junit-4.13.2",
-              "powermock-core-1.6.5",
-              "powermock-api-mockito-1.6.5",
-              "powermock-api-support-1.6.5",
+              "powermock-core-1.7.4",
+              "powermock-api-mockito-1.7.4",
+              "powermock-api-support-1.7.4",
```
timtebeek, line 41:
```suggestion
              "powermock-core-1.6.5",
              "powermock-api-mockito-1.6.5",
              "powermock-api-support-1.6.5",
```

PR #342, opened 2023-05-23
Harmonize `JavaParser` classpath setup

knutwannheden:


knutwannheden:
@timtebeek Looks like my commit https://github.com/openrewrite/rewrite/commit/3aa3cfa76789de8b781a112f276d3daf1112b535 did the trick! I would however still suggest that we harmonize the classpath entries as in this PR. What do you think?

knutwannheden:
> Indeed better to be consistent in using major.minor only.. might even go for `-5\\.9` to really drive home that it's a regular expression, as that was not always clear before. Nice find in fixing that upstream!

I raised the question on Slack regarding the interpretation as regex. I suggest we leave it as is until we have reached a conclusion there.

Issue #212, opened 2022-04-02
Hamcrest to JUnit Jupiter migration recipes


yeikel:
While the Hamcrest  repo is still [active](https://github.com/hamcrest/JavaHamcrest/commits/master), the latest release is from  [Oct 16, 2019 ](https://github.com/hamcrest/JavaHamcrest/tags)

I think that this is a potential good use case for a migration recipe

We could have two different types of migration recipes (and issues) 

1. https://github.com/openrewrite/rewrite-testing-frameworks/issues/212
2. https://github.com/openrewrite/rewrite-testing-frameworks/issues/357

Relevant discussion : https://github.com/hamcrest/JavaHamcrest/issues/353

tkvangorder:
For context, we had outlined some specifics around this migration. We ended up closing the issue because we did not have the resources to work on it at the time.  

https://github.com/openrewrite/rewrite-testing-frameworks/issues/16

rovarga:
One thing that might make sense is the straightforward migration of
`assertThat(foo, instanceOf(Foo.class)`
to JUnit5 `assertInstanceOf(Foo.class, foo)`.
That would provide a framework-independent migration, potentially lowering dependencies on Hamcrest for JUnit5 users.

rovarga:
> @rovarga That would indeed make sense as a starting point; is that something you'd be willing to contribute with some guidance? I suspect there's quite a few more in #16 that are easy to tackle once an initial sample recipe is available.

Yes I would certainly be interested, unfortunately I do not have the cycles right now. Perhaps in a couple of weeks (I will keep this on my radar).

timtebeek:
Glad to hear you're interested in helping out! Let us know how we can best support you. I'll pitch a quick outline right now.

Looking through the list in #16 I'm seeing quite a few matchers that have one to one replacements in AssertJ, which might be the easiest to tackle. Assuming here that people picked Hamcrest for the fluent matchers, and then AssertJ is the most direct replacement to that.

Alternatively we can (only) migrate to JUnit 5, and then optionally after that chain the existing migration from JUnit 5 to AssertJ; that way we can have both, provided there's equivalents for all matchers in both JUnit 5 and AssertJ, and recipes to go between those.

Either way I suppose you can start by creating a recipe that takes in two options:
1. a String [method pattern](https://docs.openrewrite.org/reference/method-patterns) of a Matchers method to replace, for instance `org.hamcrest.Matchers  equalTo`.
[...]

PR #343, opened 2023-05-23
Migrate Hamcrest to JUnit 5

matusmatokpt:
This is a draft pull request for 
- #212


src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,55 @@
+package org.openrewrite.java.testing.junit5;
```
knutwannheden, line 1: The license header is missing. To add it you can either copy it from another source file or run `./gradlew lF`, which will add the header to all files which don't have it yet.

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+            throw new IllegalArgumentException("Trying to strip an expression which is not a matcher invocation:\n" + e);
+        }
+
+        private String getTranslatedAssert(J.MethodInvocation methodInvocation) {
+            //to be replaced with a static map
+            switch (methodInvocation.getSimpleName()) {
+                case "equalTo":
+                    return "assertEquals";
```
timtebeek, line 105: Rather than statically defining the various options here, we could also look at turning the `hamcrestMatcher` and `targetAssertion` into `@Option` arguments to the recipe, such that we can configure this repeatedly from a declarative Yaml description file. We already do [something similar for Mockito](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/resources/META-INF/rewrite/mockito.yml), except in thise case you'd be configuring this new recipe rather than `ChangeMethodName`. Such an approach pull those replacements out of code and into configuration, which is typically easier to expand on, even locally, for our users.

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,119 @@
...
+        private J.MethodInvocation withTemplate(J.MethodInvocation method, JavaTemplate template, Expression firstArg, List<Expression> matcherArgs) {
+            switch (matcherArgs.size()) {
+                case 0:
+                    return method.withTemplate(template, method.getCoordinates().replace(), firstArg);
+                case 1:
+                    return method.withTemplate(template, method.getCoordinates().replace(), firstArg, matcherArgs.get(0));
+                case 2 :
+                    return method.withTemplate(template, method.getCoordinates().replace(), firstArg, matcherArgs.get(0), matcherArgs.get(1));
+                case 3 :
+                    return method.withTemplate(template, method.getCoordinates().replace(), firstArg, matcherArgs.get(0), matcherArgs.get(1), matcherArgs.get(2));
```
timtebeek, line 85: The last argument in withTemplate is actually a varargs. So you could reduce this when opting for a `Object[] args` as seen here. https://github.com/openrewrite/rewrite-testing-frameworks/blob/f25ff0b0ed2b9e227b07972986b35f7a1cc958af/src/main/java/org/openrewrite/java/testing/cleanup/AssertTrueEqualsToAssertEquals.java#L75-L80

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,119 @@
...
+                    return "assertEquals";
+            }
+            throw new IllegalArgumentException("Translation of matcher " + methodInvocation.getSimpleName() + " not yet supported.");
+        }
+
+        private String getTemplateForTranslatedAssertion(String translatedAssertion) {
+            //to be replaced with a static map
+            switch (translatedAssertion) {
+                case "assertEquals":
+                    return "assertEquals(#{any(java.lang.Object)}, #{any(java.lang.Object)})";
```
timtebeek, line 114: I think the recipe will fail currently with any 3/4 method arguments, as the template hardcodes the expected number of arguments to two; Might be best to add tests for various amounts of arguments and work from there to repeatedly append `, #{any(java.lang.Object)}` as needed, as also seen here.
https://github.com/openrewrite/rewrite-testing-frameworks/blob/f25ff0b0ed2b9e227b07972986b35f7a1cc958af/src/main/java/org/openrewrite/java/testing/cleanup/AssertTrueEqualsToAssertEquals.java#L74-L82

src/test/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrestTest.java
```diff
@@ -0,0 +1,78 @@
...
+    void equalToObject() {
+        //language=java
+        rewriteRun(
+          java("""
+            class Biscuit {
+                String name;
+                Biscuit(String name) {
+                    this.name = name;
+                }
+            }
```
timtebeek, line 45: We might also want to verify our use of `equalTo` with Strings and other primitives, just to be sure we don't inadvertently produce errors there.

src/test/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrestTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import static org.openrewrite.java.Assertions.java;
+
+class MigrateFromHamcrestTest implements RewriteTest {
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion()
+            .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9", "hamcrest-2.2"))
+          .recipe(new MigrateFromHamcrest());
```
timtebeek, line 32: If we do end up going for declarative yaml recipes for the various matcher-to-method replacements, then we will likely want to import yaml recipes and run those instead of this direct invocation. Here's an example. https://github.com/openrewrite/rewrite-testing-frameworks/blob/f25ff0b0ed2b9e227b07972986b35f7a1cc958af/src/test/java/org/openrewrite/java/testing/mockito/AnyToNullableTest.java#L36-L39

timtebeek:
Appreciate you getting back to this! I've asked @knutwannheden to have a look as I'm caught up in other work this week; from the failures it appears your recipe might not apply or make changes. Did you already try to step through with the debugger?

matusmatokpt:
Yes I did. Seemingly, the recipe instantiates alright, but during the run phase it fails and I don't understand why. Wanted to double check it so I inserted some console prints into the recipe and they are printed as expected when the test is run. I will add some more simple matcher-to-method cases while I wait for the review.

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,107 @@
...
+        return new MigrationFromHamcrestVisitor();
+    }
+
+    private static class MigrationFromHamcrestVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
+            J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
+            MethodMatcher matcherAssertTrue = new MethodMatcher("org.hamcrest.MatchAssert assertThat(String, boolean)");
+            MethodMatcher matcherAssertMatcher = new MethodMatcher("org.hamcrest.MatcherAssert assertThat(*, org.hamcrest.Matcher)");
```
timtebeek, line 50: Not sure why this matcher is failing, but you can work around it by for now adopting:
```suggestion
            MethodMatcher matcherAssertMatcher = new MethodMatcher("org.hamcrest.MatcherAssert assertThat(..)");
```
You might need to match the three different MethodMatcher assertThat methods in a specific order with this `..` variant last, but at least that will get you going.

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,107 @@
...
+                      .build();
+
+                    Expression strippedMatcher = mi.getArguments().get(1);
+                    if (strippedMatcher instanceof J.MethodInvocation) {
+                        strippedMatcher = ((J.MethodInvocation) strippedMatcher).getArguments().get(0);
+                    } else {
+                        throw new IllegalArgumentException("Second parameter expected to be a matcher constructor call.");
+                    }
+                    template.apply(getCursor(), method.getCoordinates().replace(), mi.getArguments().get(0), strippedMatcher);
+                    maybeAddImport("org.junit.jupiter.api.Assertions", targetAssertion);
```
timtebeek, line 78: Don't forget to return the changed template! With this return a first of your tests now passes 🎉 
```suggestion
                    maybeAddImport("org.junit.jupiter.api.Assertions", targetAssertion);
                    return template.apply(getCursor(), method.getCoordinates().replace(), mi.getArguments().get(0), strippedMatcher);
```

timtebeek:
> Yes I did. Seemingly, the recipe instantiates alright, but during the run phase it fails and I don't understand why. Wanted to double check it so I inserted some console prints into the recipe and they are printed as expected when the test is run. I will add some more simple matcher-to-method cases while I wait for the review.

Sorry about the delays in getting you feedback; It's been pretty busy with the new 8.0 release coming up.

I've posted two suggestions to get your first tests passing; That should help to start iterating by adding options, configuring those from Yaml, and see how far we can get with covering different Hamcrest matchers. Hope that helps!

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
matusmatokpt: this seemingly has something to do with matcherMethod being called on primitives, i.e., `equalTo(5)`.

timtebeek:
> You keep mentioning something with YAML, but I think I will need more explanation in that department. I will probably get to work on this in the next week, so the review is in no rush. Thanks

I did not want to thoroughly change your pull request, so instead I opted to start a parallel effort to migrate to AssertJ, for people that are interested in that: https://github.com/openrewrite/rewrite-testing-frameworks/pull/355
There you can see I have a recipe that takes in two `@Option`s, and uses those for replacements. I then configure that recipe in a [Yaml description file](https://docs.openrewrite.org/authoring-recipes/writing-a-java-refactoring-recipe#declarative-yaml-usage), such that we can easily extend that without needing to update the Java code. Adding tests for more cases is then optional.

timtebeek:
Once https://github.com/openrewrite/rewrite-testing-frameworks/pull/355 is through we can even reuse some of the components used in there here, such as removing the intermediate `is(Matcher)` invocations before a further migration. Also note how I've created three separate recipes there to keep complexity manageable; the easiest of which is the recipe to replace `assertThat(String, boolean)` with `assertThat(boolean).as(String).isTrue()`. I imagine you could do much of the same here.

matusmatokpt:
I see how the `@Option` works now, thanks. The problem with Hamcrest to JUnit5 is that the matcher-method mapping is not always straightforward. Therefore I had to get creative in some cases by creating wild templates. Therefore if I we to use `@option` approach, I would have to probably have the entire template string as an option. Moreover, if the template gets complicated, the order of arguments may get shuffled, or some of the original arguments may get used in the translation more than once. That I do not know how to handle with `@Option`. Moreover I still dont know if those implemented work as the tests keep failing, will look into that again.

My approach now is to add all the mappings that I am capable of figuring out to this single recipe and then we can have a debate on how to split it in the most sensible way, how about that?

Regarding your previous comment, I will make a separate recipe for `assertThat (String, boolean)` after I finish with the mappings.

timtebeek:
Yeah sure; good to continue this way until we get more coverage; I guess by then it will be easier to spot overlap between different number and order of arguments, and the corresponding templates. We might by then choose to split that up either in separate visitors, or recipes, or keep it as is depending on what seems best by then. Thanks again for your work towards this!

matusmatokpt:
Hi! Managed to finish the translations for all the simple cases (i.e. matchers that do not consume other matchers as parameters (except `is` and `not` - more on that later)). I would appreciate if you gave your opinion on how to reasoanbly split it. 
I also managed to transalte `is` and `not` matchers, so we should not need to bother with those no more.

I came with an idea how to tackle the more complicated cases by using "chaining" operations.
For example:
```
List<Integer> examinedList = List.of(1,2,3,4);
assertThat(examinedList, not(hasSize(greaterThan(1))));
```
could get translated into something like:
```
List<Integer> examinedList = List.of(1,2,3,4);
assertFalse(examinedList.size() > (1));
|.................|
  flipped by not
                    |...............................|
                            first matcher
                                                       |....|
                                                       second matcher .. and potentially so on
```

timtebeek:
@joanvr are you able to weigh in here for a review?
I'm unavailable tomorrow and over the weekend.

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,375 @@
...
+            MethodMatcher matcherAssertTrue = new MethodMatcher("org.hamcrest.MatchAssert assertThat(String, boolean)");
+            MethodMatcher matcherAssertMatcher = new MethodMatcher("org.hamcrest.MatcherAssert assertThat(..)");
+            MethodMatcher matcherAssertMatcherWithReason = new MethodMatcher("org.hamcrest.MatcherAssert assertThat(String,*,org.hamcrest.Matcher)");
+
+            if (matcherAssertTrue.matches(mi)) {
+                //TODO simple
+            } else if (matcherAssertMatcher.matches(mi)) {
+                Expression hamcrestMatcher = mi.getArguments().get(1);
+                if (hamcrestMatcher instanceof J.MethodInvocation) {
+                    System.out.println("matched");
```
timtebeek, line 62:
```suggestion
```

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,375 @@
...
+                      .staticImports("org.junit.jupiter.api.Assertions." + targetAssertion)
+                      .build();
+
+                    maybeAddImport("org.junit.jupiter.api.Assertions", targetAssertion);
+                    return template.apply(getCursor(), method.getCoordinates().replace(),
+                        getArgumentsForTemplate(matcherInvocation, null, mi.getArguments().get(0), false));
+                }
+                else throw new IllegalArgumentException("Parameter mismatch for " + mi + ".");
+            }
+            System.out.println("FINISH RUN");
```
timtebeek, line 82:
```suggestion
```

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,375 @@
...
+                    return negated ? "assertFalse" : "assertTrue";
+                case "instanceOf":
+                case "isA":
+                    return negated ? "assertFalse" : "assertInstanceOf";
+                case "is":
+                    if (methodInvocation.getArguments().get(0).toString().startsWith("org.hamcrest")) {
+                        if (methodInvocation.getArguments().get(0) instanceof J.MethodInvocation) {
+                            return getTranslatedAssert((J.MethodInvocation)methodInvocation.getArguments().get(0), negated);
+                        } else {
+                            throw new IllegalArgumentException();
```
timtebeek, line 122: Instead of throwing an exception we prefer not to make a change; would that be possible here?
Otherwise we trip up for instance a complete Spring Boot 3.1 migration, if that somewhere down to line includes the Hamcrest migration as part of the JUnit 4 to 5 migration.

src/main/java/org/openrewrite/java/testing/junit5/MigrateFromHamcrest.java
```diff
@@ -0,0 +1,375 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new MigrationFromHamcrestVisitor();
+    }
+
+    private static class MigrationFromHamcrestVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
+            System.out.println("RECIPE RUN");
```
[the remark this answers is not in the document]
matusmatokpt: all of these console prints are just to help to debug the non-working tests. They will of course disappear once I make the tests work.

matusmatokpt:
Could anyone please help me resolve this?
```
                    String statikImport = "org.junit.jupiter.api.Assertions." + targetAssertion;

                    JavaTemplate template = JavaTemplate.builder(getTemplateForMatcher(matcherInvocation,null, false))
                      .javaParser(JavaParser.fromJavaVersion().classpathFromResources(executionContext, "junit-jupiter-api-5.9"))
                      .staticImports(statikImport)
                      .build();

                    maybeAddImport("org.junit.jupiter.api.Assertions", targetAssertion);
                    return template.apply(getCursor(), method.getCoordinates().replace(),
                        getArgumentsForTemplate(matcherInvocation, null, mi.getArguments().get(0), false));
                    ```
In the cases where `targetAssertion == "assertEquals"`, the static import gets successfully added however, if `targetAssertion == "assertTrue"`, it does not get added. Everything gets refactored as I wish, only the import is missing:

```
expected: 
  "import org.junit.jupiter.api.Test;
  
  import static org.junit.jupiter.api.Assertions.assertTrue;
  
  class BiscuitTest {
      @Test
      void testEquals() {
          int intt = 7;
          assertTrue(10 > intt);
      }
  }"
 but was: 
  "import org.junit.jupiter.api.Test;
  
  class BiscuitTest {
      @Test
      void testEquals() {
          int intt = 7;
          assertTrue(10 > intt);
      }
  }"
  ```

knutwannheden:
@matusmatokpt The problem here is the condition in your template for `assertTrue()`. You have `"(Math.abs(#{any(java.lang.Object)} - #{any(java.lang.Object)}) < #{any(java.lang.Object)}"`. But `JavaTemplate` won't be able to compile that, since using `java.lang.Object` doesn't make any sense for the binary `-` and `<` operators. Rather try with `"(Math.abs(#{any(double)} - #{any(double)}) < #{any(double)}"`.

This being said, I think it would make much more sense to translate these Hamcrest assertions to AssertJ. There is things which correspond to `closeTo()` there and will give much better error messages than `assertTrue()`.

knutwannheden:
I suggest you take a look at #355 and see if there is anything missing there you would like to improve. If you on the other hand cannot use AssertJ, it would be interesting to hear more about this use case.

matusmatokpt:
Thanks for your review. I was specifically asked to address Hamcrest -> JUnit5 translation, as we need it for another project where we used to use hamcrest and use JUnit5 at the moment. That is why I am pushing towards this direction. I am not sure I will finish a complete solution, however there may be this partial one that either I or someone else can finish in the future. 
Regarding the faulty templates, thanks for your hint, I did not assume that could be a problem as the template was succesfully applied to the method invocation body.

matusmatokpt:
Hello, added a bunch of tests, all of them pass, but the last. I tried all the hints you ever gave me, but still cannot make it work. So if anyone had time and could take a look I would appreaciate it. It is the very last test method in MigrateFromHamcrestTest (hasEntry()). Thank you

matusmatokpt:
Feels like I finished this pull request. Added all the necessary tests and they all pass  (localy). Also, split the implementation into two recipes and integrated all the miscellaneous suggestions given by @timtebeek. I am looking forward to the code review. 
I am considering how to approach the more complicated cases, where the static matcher method consumes different matchers (such as anyOf does).

timtebeek:
Thanks a lot for your continued effort on this! I've moved your classes such that they are more in line with the AssertJ recipes that we have a in parallel. What I like about your approach is that you're better able to handle nested Matchers, which is a bit of a struggle with the alternative implementation for AssertJ.

What I'm wondering with your implementation is if it would make sense to introduce an enum with fields to hold the before, after and template method strings. For instance with some rough pseudo code here
```java
enum Replacements {
    EQUALTO("equalTo", "assertEquals", "#{any(java.lang.Object)}, #{any(java.lang.Object)}",
            matcher -> matcher.getArguments()),
   // TODO More replacements
    ;

    String before, after, template;
    Function<J.MethodInvocation, List<Expression>> arguments;

    Replacements(String before, String after, String template, Function<J.MethodInvocation, List<Expression>> arguments) {
        this.before = before;
        this.after = after;
        this.template = template;
        this.arguments = arguments;
    }
}
```

Not sure if that would work out for all cases, but it could help group together the logic that's now maintained in three distinct switch statements into a structure that's perhaps easier to read and expand. Would you feel there's any merit to exploring that option?

It would maybe need some special handling up front for `is` and `not` and other special cases, perhaps even using what we already have in `org.openrewrite.java.testing.hamcrest.RemoveIsMatcher`.

matusmatokpt:
The enum would probably have to hold  the entire template as well, but sometimes the template is diffrent if the context is negated by `not`, which would need some headscratching. I am busy in the upcomming days, but I'll let this live at the back of my head and hopefully I ll come up with something.
I was thinking of rehosting the `not` and `is` cases to separate class that would remove them and would modify the `ExecutionContext` variable so that the inheritors of that class would not need to worry about `is` and `not` which would get removeved by calling `super.visitMethodInvocation(.., ExecutionContext)`. The only problem is the removal of the information about negation from `ExecutionContext`. It would be nice, if the super method would remove the information from there so that the callee doesnt need to bother with that. That is some more thinking to be done.
Do these ideas sound reasonable? All suggestions are appreciated.

matusmatokpt:
Obviously, this PR is not yet merged, but it has functionality sufficient for some cases on our project. Therefore I have an offtopic question. Can I setup the recipe for the project from my fork or does it have to be merged into upstream?

timtebeek:
> I was thinking of rehosting the not and is cases to separate class that would remove them and would modify the ExecutionContext variable so that the inheritors of that class would not need to worry about is and not

I think that's a perfectly valid approach; I've split those out in the AssertJ migration as well, as they're quite special. The `is` case you can work around in a similar manner as seen here
https://github.com/openrewrite/rewrite-testing-frameworks/blob/266c018a1680400df1acc7cd263fd0a2cc1da897/src/main/resources/META-INF/rewrite/hamcrest.yml#L41-L50
There we _first_ remove any wrapping `is` before applying further recipes. You can likely use that exact same recipe to simplify the conversion.

> Obviously, this PR is not yet merged, but it has functionality sufficient for some cases on our project. Therefore I have an offtopic question. Can I setup the recipe for the project from my fork or does it have to be merged into upstream?

Absolutely; you can install the library to your local Maven repository through `./gradlew publishToMavenLocal`, and then pick it up in a similar manner to [our snapshots versions](https://docs.openrewrite.org/reference/snapshot-instructions), but using your local Maven repository. That's a great way to verify your recipes work in practice as well, and find more edge cases to convert.

matusmatokpt:
Hello, attempted to apply the recipe to our project, but cannot make it work. First I used the command u provided `./gradlew publishToMavenLocal` which amongst everything else yielded: 
```
> Configure project :
Inferred project: rewrite-testing-frameworks, version: 0.1.0-SNAPSHOT

```
So I went to try it in our local project, and added the plugin as follows:
```
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>4.41.1</version>
    <configuration>
        <activeRecipes>
            <recipe>org.openrewrite.java.testing.junit5.MigrateFromHamcrest</recipe>
            <recipe>org.openrewrite.java.testing.junit5.AssertThatBooleanToJUnit5</recipe>
        </activeRecipes>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-testing-frameworks</artifactId>
            <version>0.1.0-SNAPSHOT</version>
        </dependency>
    </dependencies>
</plugin>
```
Then I simply ran `mvn clean install` but none of the test files got refactored, even though they should (they contain hamcrest matchers translatable by my recipe). Is there anything I am doing wrong?

knutwannheden:
> Hello, attempted to apply the recipe to our project, but cannot make it work. First I used the command u provided `./gradlew publishToMavenLocal` which amongst everything else yielded:
> 
> ```
> > Configure project :
> Inferred project: rewrite-testing-frameworks, version: 0.1.0-SNAPSHOT
> ```

This is very strange. It should instead print `Inferred project: rewrite-testing-frameworks, version: 2.1.0-SNAPSHOT`. I am not sure where that 0.1.0 version is coming from when you run the command. Possibly this could explain why it then doesn't work. Can you check if your local Maven repo really does contain an artifact with the `0.1.0-SNAPSHOT`?

matusmatokpt:
Negative. `mvn dependency:get -Dartifact=org.openrewrite.recipe:rewrite-testing-frameworks:0.1.0-SNAPSHOT -o -DrepoUrl=file://~/.m2/repository` results in a nasty error message, therefore I assume it does not contain it, unless I butchered the command.

knutwannheden:
You can also just inspect the contents of your `~/.m2/repository/org/openrewrite/recipe/rewrite-testing-framework/` directory, to check what you find there.

matusmatokpt:
Okay. Found it

knutwannheden:
I still don't understand why it would publish it to your local Maven repo using the version 0.1.0-SNAPSHOT. 2.1.0-SNAPSHOT would be the expected version.

matusmatokpt:
well, neither do I. I have not touched any config files while working on this PR. Only the recipes, so I do not know. Either way, assume that 0.1.0-SNAPSHOT is the correct version. What could be other factors causing my recipe not to work in our project? (mvn config as above, prints no errors on build)

knutwannheden:
Hmm. If you run `mvn clean install` that would typically not cause the rewrite plugin to do anything. Have you tried explicitly running the `run` or the `dryRun` goals? See https://docs.openrewrite.org/running-recipes/running-rewrite-on-a-maven-project-without-modifying-the-build and https://docs.openrewrite.org/reference/rewrite-maven-plugin

timtebeek:
Hi @matusmatokpt ! Was wondering what the current status of the PR is to you; Were you able to apply the changes locally? Are you still aiming to see this through to a merge?

We've made some process since with recipes that prepare a migration such as removing `is` and flatting `allOf`; perhaps those could apply here as well?

matusmatokpt:
Hi. I have been swamped with different tasks. I want to eventually finish this in my free time, but I am not sure when I ll have chance to get back to this. Maybe next week. 
I have not noticed any new recipes, will take a look. A link to the related PR(s) please?

timtebeek:
> Hi. I have been swamped with different tasks. I want to eventually finish this in my free time, but I am not sure when I ll have chance to get back to this.

All good! No rush; I'll mark the PR as a draft and move it to in progress then, just so that it's clear when you'd want to start a new round of review.

> I have not noticed any new recipes, will take a look. A link to the related PR(s) please?

Sure! Specifically these two I thought could be interesting to reuse to simplify your migration recipes.
https://github.com/openrewrite/rewrite-testing-frameworks/blob/244aea7c3807fc4b1724c9bec11fd64323a58fca/src/main/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcher.java#L25-L28

https://github.com/openrewrite/rewrite-testing-frameworks/blob/244aea7c3807fc4b1724c9bec11fd64323a58fca/src/main/java/org/openrewrite/java/testing/hamcrest/FlattenAllOf.java#L34-L37

We have [similar plans still for `anyOf`](https://github.com/openrewrite/rewrite-testing-frameworks/issues/357#issuecomment-1643643813), the concepts of which can again be reused here.

matusmatokpt:
Looks neat. However, the the effort to migrate `anyOf` as described above will not be of much use to the hamcrest -> JUnit5 case.

When it comes to this recipe, I will try to wire it up using enums as you suggested, or maybe I will create a nested class called for example RecipeCase, which will contain the necessary fields. Will see what works the best. Either way, I can probably remove the `is` matcher case from the recipe.

timtebeek:
Thanks for your continued plans to work on this! In general I would say try to break the recipe up into smaller reusable components that we can then wire together in a declarative yaml file; that way it's easier to implement, reason about and maintain.

matusmatokpt:
Hello, reworked this. Locally, everything is building fine and all tests pass. Could anyone take a look?

knutwannheden:
I don't have the permission to push to the fork, but it looks like this patch would fix the compilation error. Maybe that is something you can apply, @timtebeek ?

```patch
Index: src/test/java/org/openrewrite/java/testing/hamcrest/RemoveNotMatcherTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveNotMatcherTest.java b/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveNotMatcherTest.java
--- a/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveNotMatcherTest.java	(revision a18e7aceb1905fdd14a790899626ab2b875f19b3)
+++ b/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveNotMatcherTest.java	(revision 0a4ef0be250926b82829d816d981d9d9e422dd0b)
@@ -34,8 +34,6 @@
     }
 
     @DocumentExample
-          java(
-                """
     void nestedNotMatcher() {
         rewriteRun(
           //language=java
@@ -44,7 +42,7 @@
             import static org.hamcrest.MatcherAssert.assertThat;
             import static org.hamcrest.Matchers.equalTo;
             import static org.hamcrest.Matchers.not;
-                            
+
             class ATest {
                 @Test
                 void testEquals() {
@@ -58,7 +56,7 @@
             import static org.hamcrest.MatcherAssert.assertThat;
             import static org.hamcrest.Matchers.equalTo;
-                            
+
             class ATest {
                 @Test
                 void testEquals() {
@@ -79,7 +77,7 @@
             import static org.hamcrest.MatcherAssert.assertThat;
             import static org.hamcrest.Matchers.not;
-                            
+
             class ATest {
                 @Test
                 void testEquals() {
@@ -93,7 +91,7 @@
             import static org.hamcrest.MatcherAssert.assertThat;
             import static org.hamcrest.Matchers.equalTo;
-                            
+
             class ATest {
                 @Test
                 void testEquals() {
```

PR #344, opened 2023-05-25
Added recipe to use explicit isEmpty in AssertJ assertions

### LINKED ISSUE #1: Create ReplaceAnnotations visitor which transforms Junit 4 annotations into Junit 5 annotations
Junit 5 replaced several Junit 4 annotations with very similar annotations.

Create a `ReplaceAnnotations` visitor which finds these method-scoped JUnit4 annotations and replaces them with their Junit5 successors:

* [@org.junit.Before](https://junit.org/junit4/javadoc/4.12/org/junit/Before.html) annotation is replaced by [@org.junit.jupiter.api.BeforeEach](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/BeforeEach.html)
* [@org.junit.After](https://junit.org/junit4/javadoc/4.12/org/junit/After.html) annotation is replaced by  [@org.junit.jupiter.api.AfterEach](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/AfterEach.html)
* [@org.junit.BeforeClass](https://junit.org/junit4/javadoc/4.12/org/junit/BeforeClass.html) annotation is replaced by  [@org.junit.jupiter.api.BeforeAll](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/BeforeAll.html)
* [@org.junit.AfterClass](https://junit.org/junit4/javadoc/4.12/org/junit/AfterClass.html) annotation is replaced by  [@org.junit.jupiter.api.AfterAll](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/AfterAll.html)
* [@org.junit.Ignore](https://junit.org/junit4/javadoc/4.12/org/junit/Ignore.html) annotation is replaced by  [@org.junit.jupiter.api.Disabled](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/api/Disabled.html)

## Implementation

It is likely that this can be accomplished only with yaml/declarative visitors calling [ChangeType](https://github.com/openrewrite/rewrite/blob/1162273ad0bafbe2d01059649735500ef19b5266/rewrite-java/src/main/java/org/openrewrite/java/ChangeType.java#L38) since most or all of these are drop in replacements.
For a similar example of declaratively using ChangeType to switch annotation types, see this similar example from the [mockito recipes](https://github.com/openrewrite/rewrite-spring/blob/master/src/main/resources/META-INF/rewrite/mockito.yml#L20).

Note that `@Ignore` can optionally accept an argument. If such an argument exists, it should be preserved as an argument to the replacing `@Disabled`. If `@Ignore` and `@Disabled` accept the same arguments then fully yaml/declarative ChangeType is sufficient. If the arguments have to be transformed then you'll probably have to write a visitor in Java to handle that.

If it does turn out that you need to implement a new visitor in Java, you will probably only have to override `JavaRefactorVisitor.visitMethod()`. For a similar annotation-editing java visitor, see the similar (but more complex) [ChangeTestAnnotation](https://github.com/openrewrite/rewrite-testing/blob/master/src/main/java/org/openrewrite/java/testing/junit5/ChangeTestAnnotation.java) visitor which replaces the Junit4 version of @Test with the Junit5 version of @Test. 

## Testing

Regardless of whether or not your visitor is written in YAML or Java, create a test case that exercises your visitor, asserting that it refactors a java source file that used the Junit4 annotations into one which uses the Junit 5 annotations.
Here is an example of a similar [test class](https://github.com/openrewrite/rewrite-spring/blob/e738adad2c6f63ceb1365e385c83bffe4a01e6b0/src/test/kotlin/org/openrewrite/java/mockito/MockitoUpgrade1To3Tests.kt#L26) that exercises visitors declared in a yaml file.

## Documenting

There should be a new page for this visitor added to our docs

pibizza:
## What's changed?
I have added a new recipe to simplify assertions in tests that use AssertJ isEmpty().
Specifically, this recipe simplifies assertThat(collection.isEmpty()).isTrue() into assertThat(collection).isEmpty() and assertThat(collection.isEmpty()).isFalse() into assertThat(collection).isNotEmpty()

## What's your motivation?

I work a lot with AssertJ and I have lots of legacy code to cleanup assertions.

This is similar to other recipes I have contributed, so there should be no big issue.

### Checklist
- [X] I've added the license header to any new files through `./gradlew licenseFormat`
- [ ] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2023 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
@@ -30,16 +30,16 @@ import org.openrewrite.java.tree.J.MethodInvocation;
 import java.time.Duration;
 import java.util.function.Supplier;
 
-public class UseExplicitContains extends Recipe {
+public class UseExplicitIsEmpty extends Recipe {
     @Override
     public String getDisplayName() {
-        return "Use AssertJ `contains()` on collections";
+        return "Use AssertJ `isEmpty()` on collections";
     }
 
     @Override
     public String getDescription() {
-        return "Convert AssertJ `assertThat(collection.contains(element)).isTrue()` with assertThat(collection).contains(element) "
-                + "and `assertThat(collection.contains(element)).isFalse()` with assertThat(collection).doesNotContain(element).";
+        return "Convert AssertJ `assertThat(collection.isEmpty()).isTrue()` with assertThat(collection).isEmpty() "
+                + "and `assertThat(collection.isEmpty()).isFalse()` with assertThat(collection).isNotEmpty().";
     }
 
     @Override
@@ -54,15 +54,15 @@ public class UseExplicitContains extends Recipe {
 
     @Override
     protected TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new UseExplicitContainsVisitor();
+        return new UseExplicitContainsIsEmpty();
     }
 
-    public static class UseExplicitContainsVisitor extends JavaIsoVisitor<ExecutionContext> {
+    public static class UseExplicitContainsIsEmpty extends JavaIsoVisitor<ExecutionContext> {
         private Supplier<JavaParser> assertionsParser;
         private Supplier<JavaParser> assertionsParser(ExecutionContext ctx) {
             if(assertionsParser == null) {
                 assertionsParser = () -> JavaParser.fromJavaVersion()
-                        .classpathFromResources(ctx, "assertj-core-3.24")
+                        .classpathFromResources(ctx, "assertj-core-3.24.2")
                         .build();
             }
             return assertionsParser;
@@ -71,7 +71,7 @@ public class UseExplicitContains extends Recipe {
         private static final MethodMatcher ASSERT_THAT = new MethodMatcher("org.assertj.core.api.Assertions assertThat(..)");
         private static final MethodMatcher IS_TRUE = new MethodMatcher("org.assertj.core.api.AbstractBooleanAssert isTrue()");
         private static final MethodMatcher IS_FALSE = new MethodMatcher("org.assertj.core.api.AbstractBooleanAssert isFalse()");
-        private static final MethodMatcher CONTAINS = new MethodMatcher("java.util.Collection contains(..)", true);
+        private static final MethodMatcher IS_EMPTY = new MethodMatcher("java.util.Collection isEmpty()");
 
         @Override
         public J.MethodInvocation visitMethodInvocation(J.MethodInvocation m, ExecutionContext ctx) {
@@ -95,24 +95,22 @@ public class UseExplicitContains extends Recipe {
                 return method;
             }
 
-            J.MethodInvocation contains = (J.MethodInvocation) assertThat.getArguments().get(0);
-            if (!CONTAINS.matches(contains)) {
+            J.MethodInvocation isEmpty = (J.MethodInvocation) assertThat.getArguments().get(0);
+            if (!IS_EMPTY.matches(isEmpty)) {
                 return method;
             }
 
-            Expression list =  contains.getSelect();
-            Expression element = contains.getArguments().get(0);
+            Expression list =  isEmpty.getSelect();
 
-            String template = isTrue ? "assertThat(#{any()}).contains(#{any()});" :
-                "assertThat(#{any()}).doesNotContain(#{any()});";
+            String template = isTrue ? "assertThat(#{any()}).isEmpty();" :
+                "assertThat(#{any()}).isNotEmpty();";
             JavaTemplate builtTemplate = JavaTemplate.builder(this::getCursor, template)
                     .javaParser(assertionsParser(ctx))
                     .build();
             return method.withTemplate(
                     builtTemplate,
                     method.getCoordinates().replace(),
-                    list,
-                    element);
+                    list);
         }
     }
 }
--- a/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
@@ -39,7 +39,7 @@ public class UseExplicitIsEmpty extends Recipe {
     @Override
     public String getDescription() {
         return "Convert AssertJ `assertThat(collection.isEmpty()).isTrue()` with assertThat(collection).isEmpty() "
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/java/testing/assertj/UseExplicitisEmptyTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+import static org.openrewrite.java.Assertions.java;
+
+@SuppressWarnings("ConstantConditions")
+class UseExplicitisEmptyTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion().classpath("junit", "assertj-core"))
```
timtebeek, line 31: I think we might be implicitly matching `junit-jupiter-api` here from just `junit`; might be better to make it explicit.
```suggestion
          .parser(JavaParser.fromJavaVersion().classpath("junit-jupiter-api", "assertj-core"))
```

src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class UseExplicitIsEmpty extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Use AssertJ `isEmpty()` on collections";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Convert AssertJ `assertThat(collection.isEmpty()).isTrue()` with assertThat(collection).isEmpty() "
```
knutwannheden, line 41:
```suggestion
        return "Convert AssertJ `assertThat(collection.isEmpty()).isTrue()` to `assertThat(collection).isEmpty()` "
```

src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
```diff
@@ -0,0 +1,116 @@
...
+                + "and `assertThat(collection.isEmpty()).isFalse()` with assertThat(collection).isNotEmpty().";
```
knutwannheden, line 42:
```suggestion
                + "and `assertThat(collection.isEmpty()).isFalse()` to `assertThat(collection).isNotEmpty()`.";
```

src/test/java/org/openrewrite/java/testing/assertj/UseExplicitisEmptyTest.java
knutwannheden: Agreed.

knutwannheden (commented):
Apart from a few minor details this looks good to me.

We could consider creating a follow-up commit where we also apply this to `CharSequence#isEmpty()` and `Optional#isEmpty()` for which AssertJ AFAIK also has dedicated methods.

pibizza:
All comments applied. I agree with @knutwannheden that we should add recipes for other cases - for example Optional needs recipes for  isPresent() and isEmpty(). There is lots to do for assertj cleanup.

src/test/java/org/openrewrite/java/testing/assertj/UseExplicitisEmptyTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion().classpath("junit-jupiter-api", "assertj-core"))
+          .recipe(new UseExplicitIsEmpty());
+    }
+
+    @Test
+    void IsEmptyAndIsTrueBecomeIsEmpty() {
```
timtebeek, line 36:
```suggestion
    @Test
    @DocumentExample
    void IsEmptyAndIsTrueBecomeIsEmpty() {
```

pibizza: Then we should add the same annotation to the other test also?

timtebeek: Would also need the corresponding import; this ensures we see a code snippet example in the documentation, which helps people grasp the intended change.

timtebeek: > Then we should add the same annotation to the other test also?

I think a single test is enough for this recipe; not sure if we support more than one. 😅

pibizza: Ok, this should be also done.

src/test/java/org/openrewrite/java/testing/assertj/UseExplicitisEmptyTest.java
```diff
@@ -0,0 +1,125 @@
...
+                        assertThat(collection).isEmpty();
+                    }
+                }
+              """
+          )
+        );
+    }
+
+    @Test
+    @DocumentExample
```
timtebeek, line 83: A single example will do; I'm not sure we'd handle two examples well, and it should be clear how the first translates to the second from the description.
```suggestion
```

src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
```diff
@@ -0,0 +1,116 @@
...
+            if (!(assertThat.getArguments().get(0) instanceof J.MethodInvocation)) {
+                return method;
+            }
+
+            J.MethodInvocation isEmpty = (J.MethodInvocation) assertThat.getArguments().get(0);
+            if (!IS_EMPTY.matches(isEmpty)) {
+                return method;
+            }
+
+            Expression list =  isEmpty.getSelect();
```
timtebeek, line 103:
```suggestion
            Expression collection =  isEmpty.getSelect();
```

src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
```diff
@@ -0,0 +1,116 @@
...
+
+            String template = isTrue ? "assertThat(#{any()}).isEmpty();" :
+                "assertThat(#{any()}).isNotEmpty();";
+            JavaTemplate builtTemplate = JavaTemplate.builder(this::getCursor, template)
+                    .javaParser(assertionsParser(ctx))
+                    .build();
+            return method.withTemplate(
+                    builtTemplate,
+                    method.getCoordinates().replace(),
+                    list);
```
timtebeek, line 113:
```suggestion
                    collection);
```

src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java
timtebeek: Figured best be explicit that we're matching a collection, not a list.

timtebeek (approved):

murdos:
> Apart from a few minor details this looks good to me.
> 
> We could consider creating a follow-up commit where we also apply this to `CharSequence#isEmpty()` and `Optional#isEmpty()` for which AssertJ AFAIK also has dedicated methods.

In fact we can even go further and migrate 'assertThat(string.size()).isEqualTo(0)' to 'assertThat(string).isEmpty()'

But I wonder if it belongs to the same recipe

pibizza:
@murdos I think we can put this in the recipe to cleanup the size. I can take care of that.

Issue #208, opened 2022-04-01
Add missing @ParameterizedTest annotation when @ValueSource is used or replace @Test with @ParameterizedTest


yeikel:
The following test is ignored by JUnit

```java
    @ValueSource(ints = {1, 3, 5, -3, 15, Integer.MAX_VALUE})
    void testIsOdd(int number) {
        assertTrue(number % 2 != 0);
    }
```

This is not ignored but it is not the correct usage of `@ValueSource`

```java
   @Test
    @ValueSource(ints = {1, 3, 5, -3, 15, Integer.MAX_VALUE})
    void testIsOdd(int number) {
        assertTrue(number % 2 != 0);
    }
```

The right syntax is : 

```java
    @ParameterizedTest
    @ValueSource(ints = {1, 3, 5, -3, 15, Integer.MAX_VALUE})
    void testIsOdd(int number) {
        assertTrue(number % 2 != 0);
    }
```

rpau:
Done

PR #347, opened 2023-06-02
adding recipe to replace Test annotation with ParameterizedTest annot…

AlekSimpson:
…ation when value source annotation is provided

## What's changed?
This is a draft pull request for a recipe that adds missing @ParameterizedTest annotation when @ValueSource is used or replace @Test with @ParameterizedTest.

## What's your motivation?
[issue #208](https://github.com/openrewrite/rewrite-testing-frameworks/issues/208)

@timtebeek 

## Any additional context
I currently have two tests created but I think there are probably still a couple more that should be added. 

### Checklist
- [ ] I've added the license header to any new files through `./gradlew licenseFormat`
- [ ] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
@@ -1,0 +1,50 @@
+package org.openrewrite.java.testing.junit5;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+public class AddParameterizedTestAnnotationTest implements RewriteTest {
+    @Test
+    void replaceTestWithParameterizedTest() {
+        rewriteRun(
+          java(
+            """
+              @Test
+              @ValueSource(ints = {1, 3, 5, -3, 15, Integer.MAX_VALUE})
+              void testIsOdd(int number) {
+                  assertTrue(number % 2 != 0);
+              }
+              """,
+            """
+              @ParameterizedTest
+              @ValueSource(ints = {1, 3, 5, -3, 15, Integer.MAX_VALUE})
+              void testIsOdd(int number) {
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
```diff
@@ -0,0 +1,50 @@
+package org.openrewrite.java.testing.junit5;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+public class AddParameterizedTestAnnotationTest implements RewriteTest {
+    @Test
```
timtebeek, line 9: Feel free to link these back to specific issues, such as
```suggestion
    @Test
    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/314")
```

src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
```diff
@@ -0,0 +1,50 @@
...
+              @Test
+              void testIsOdd(int number) {
+                assertTrue(number % 2 != 0);
+              }
+              """,
+            """
+              @Test
+              void testIsOdd(int number) {
+                assertTrue(number % 2 != 0);
+              }
```
timtebeek, line 45: Looks like you want to assert that this recipe does not make any changes; in that case the proper way to do that is to only pass in a before argument.
```suggestion
```

timtebeek (commented):
In addition to `@ValueSource` there's also for instance `@CsvSource` and `@MethodSource`; I expect we'd want them all to be handled the same with regards to adding any missing `@ParameterizedTest` annotations, and removing `@Test` annotations.

You might also want to look at any use of meta annotations, when people extend `@ParameterizedTest`, such that we don't add the annotation twice.

Would be good to cover the above in separate unit tests, such that we know how far along we are in the implementation.

Issue #314, opened 2023-02-10
Remove duplicates uses of `@TestTemplate` implementations for a single method


yeikel:
See: https://github.com/junit-team/junit5/issues/1224

```java
    @Test
    @RepeatedTest(3)
    @DisplayName("When an entry does not exist, it should be created and initialized to 0")
```

```diff



-- @Test
    @RepeatedTest(3)
    @DisplayName("When an entry does not exist, it should be created and initialized to 0")
```

timtebeek:
Related to, but different from https://github.com/openrewrite/rewrite-testing-frameworks/issues/208 ; could be worth to explore one after the other.

timtebeek:
Fixed in https://github.com/openrewrite/rewrite-testing-frameworks/pull/356 !

knutwannheden:
@timtebeek Hmm. Doesn't #356 actually fix #208 instead of this issue?

timtebeek:
No #356 is a new one that was added yesterday, and does actually fix this one as seen in the unit tests
https://github.com/openrewrite/rewrite-testing-frameworks/blob/6219e8198d0680f80b3b7afa32886d7778d95f94/src/test/java/org/openrewrite/java/testing/junit5/RemoveDuplicateTestTemplatesTest.java#L50-L67

knutwannheden:
I see. Thanks. I remember seeing that #353 was linked to this one. But I now see that that one is still open.

PR #349, opened 2023-06-07
Test to ParameterizedTest annotation

AlekSimpson:
## What's changed?
This draft PR adds a new recipe which adds missing ParameterizedTest annotations when ValueSource is used or replaces Test with ParameterizedTest.

## What's your motivation?
https://github.com/openrewrite/rewrite-testing-frameworks/issues/314

For some reason whenever the visitor gets to the ValueSource annotation the ValueSource AnnotationMatcher check does not return true. I was having a similar issue with the Test annotation but that was because I wasn't including junit and jupiter resources in the recipe spec. ValueSource is from jupiter so I don't think its the same issue now that I have included those resources in the spec.

@timtebeek

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotation.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotation.java
@@ -1,0 +1,67 @@
+package org.openrewrite.java.testing.junit5;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.java.AnnotationMatcher;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.tree.J;
+
+public class AddParameterizedTestAnnotation extends Recipe {
+    private static final AnnotationMatcher TEST_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.api.Test");
+    private static final AnnotationMatcher VALUE_SOURCE_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.params.provider.ValueSource");
+    private static final AnnotationMatcher CSV_SOURCE_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.params.provider.CsvSource");
+    private static final AnnotationMatcher METHOD_SOURCE_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.params.provider.MethodSource");
+    boolean foundTestAnnotation = false;
+
+    @Override
+    public String getDisplayName() {
+        return "Add missing @ParameterizedTest annotation when @ValueSource is used or " +
+               "replace @Test with @ParameterizedTest";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Add missing @ParameterizedTest annotation when @ValueSource is used or " +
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+public class AddParameterizedTestAnnotationTest implements RewriteTest {
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion()
+            .classpathFromResources(new InMemoryExecutionContext(), "junit-4.13", "junit-jupiter-api-5.9"))
```
timtebeek, line 17: We're only using JUnit Jupiter here, so there's no need for the JUnit 4 dependency.
```suggestion
            .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9"))
```

src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    void onlyReplacesWithValueSourceAnnotation() {
+        /*
+        This test ensures that the recipe will only run on code that includes a
+        @ValueSource(...) annotation.
+         */
+        rewriteRun(
+          java(
+            """
+              @Test
```
timtebeek, line 64: This test is lacking imports, and thus likely to have missing types, and through that failing matchers.

Same goes for the tests below.

timtebeek: It might help to check that the code snippets that you use in a text block are actually valid stand alone classes, by briefly copy-pasting these into a new file. Your IDE will then immediately tell you if you're missing imports for instance.

src/main/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotation.java
```diff
@@ -0,0 +1,74 @@
...
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.java.AnnotationMatcher;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.tree.J;
+
+public class AddParameterizedTestAnnotation extends Recipe {
+    private static final AnnotationMatcher TEST_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.api.Test");
+    private static final AnnotationMatcher VALUE_SOURCE_ANNOTATION_MATCHER = new AnnotationMatcher("@org.junit.jupiter.params.provider.ValueSource");
```
timtebeek, line 12: Had you noticed that this matcher takes in a pattern that can include parameters as well? That might be why your matcher fails. Perhaps something to trace with the debugger.

AlekSimpson: How should I update the pattern? I kinda know what needs to be added but I'm not sure how to do it in a way thats generic enough? Would something like ValueSource("#{}") work?

timtebeek:
Superseeded by #353

PR #355, opened 2023-06-15
Migrate Hamcrest Matchers to AssertJ Assertions

timtebeek:
## What's changed?
Add a generic recipe to replace Hamcrest Matchers with AssertJ Assertions.

## What's your motivation?
Hamcrest has not been released in a _long_ time, whereas AssertJ is more actively maintained.
https://github.com/openrewrite/rewrite-testing-frameworks/issues/212

If there's any additional matchers we can replace, or argument number corner cases missed.

@knutwannheden @kunli2 

I'd assume it's easier to do it this way rather than go for Java templates for match & replacement, since there's less code to maintain this way and this way there's no templates that depend on multiple libraries. We can still use templates for the more complicated cases.

## Any additional context
Similar to #343 ; this is mostly to demonstrate the Yaml recipes I had intended there.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToAssertJAssertionTest.java
+++ b/src/test/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToAssertJAssertionTest.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.testing.hamcrest;
+
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToAssertJAssertionTest.java
```diff
@@ -0,0 +1,161 @@
...
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class HamcrestMatcherToAssertJAssertionTest implements RewriteTest {
```
timtebeek, line 27: We can remove this class and only rely on MigrateHamcrestToAssertJTest below.

knutwannheden:
@timtebeek I notice this is still in draft. I take it that means it isn't ready for review yet?

timtebeek:
This already migrates some ~50 Hamcrest Matchers. I suggest we pick up any further work separately in https://github.com/openrewrite/rewrite-testing-frameworks/issues/357

kunli2:
Nice work, it looks good to me!

kunli2 (approved):

joanvr (approved):
LGTM from a code style perspective :)

timtebeek:
Thanks both! Then I'll merge and try it out on the platform when I get back Monday.

knutwannheden:
@timtebeek Did you try to cover all Hamcrest matchers or only a specific subset? It looks like the `closeTo()` matcher (see https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#closeTo(double,%20double)) isn't covered yet, which could be migrated to one of the `isCloseTo()` methods in AssertJ: http://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractDoubleAssert.html#isCloseTo-double-org.assertj.core.data.Offset-

I didn't compare the lists in detail.

timtebeek:
I'd mostly covered those for strings, objects and collections. There's indeed likely more matches that we could quite easily covered, and more complex recipes that require more work. We can track these in #357

PR #356, opened 2023-06-15
Alek/remove duplicate test templates

AlekSimpson:
## What's changed?
Remove duplicates uses of @TestTemplate implementations for a single method.

## What's your motivation?
[#314](https://github.com/openrewrite/rewrite-testing-frameworks/issues/314)

I think maybe there could be more test cases but I can't think of anymore scenarios to test right now.

@timtebeek 

### Checklist

### DIFF
--- a/gradle/licenseHeader.txt
+++ b/gradle/licenseHeader.txt
@@ -1,4 +1,4 @@
-Copyright 2021 the original author or authors.
+Copyright 2023 the original author or authors.
 <p>
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


timtebeek (approved):

timtebeek:
Nice work! I've added a small polish commit in https://github.com/openrewrite/rewrite-testing-frameworks/pull/356/commits/a0cf7ff0e56329c53f8a2debe33d2dc7b9c91562

- Use the full type of `RepeatedTest` in `UsesType` to limit when the recipe matches and is applied, otherwise we match too broadly, and evaluate a lot of test methods that most likely do not yet use `RepeatedTest`.
- Reuse the `RemoveAnnotationVisitor` rather than filtering the leading annotations; we have a lot of such small building blocks, and it's easier to write and maintain recipes when we use them as much as sensible
- Flipped the `if` condition to use `anyMatch` instead of a double `noneMatch`; reads clearer and should match marginally quicker
- Added a test `removeDuplicateOnly` which contains a mix of methods that should and should not have changes applied, such that we guarantee we only remove annotations from the target method, not any method if the visitor matches, mostly as a safe guard if there's further changes.
-  renamed the methods in the test templates from `void TestMethod` to `void testMethod` to better match what we expect in Java.
- Updated the copyright year and template for 2023; should only have to be changed once, but looked odd to see 2021 still.

PR #362, opened 2023-06-19
updated junit5 yml files to include the two new recipes in the best p…

AlekSimpson:
…ractices

## What's changed?
I updated the list of best practices recipes in the yml for junit5 with the two new recipes I recently wrote. 

@timtebeek

### DIFF
--- a/src/main/resources/META-INF/rewrite/junit5.yml
+++ b/src/main/resources/META-INF/rewrite/junit5.yml
@@ -81,6 +81,8 @@ recipeList:
   - org.openrewrite.java.testing.junit5.AddMissingNested
   - org.openrewrite.java.testing.hamcrest.AddHamcrestIfUsed
   - org.openrewrite.java.testing.junit5.UpgradeOkHttpMockWebServer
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -81,6 +81,8 @@ recipeList:
   - org.openrewrite.java.testing.junit5.AddMissingNested
   - org.openrewrite.java.testing.hamcrest.AddHamcrestIfUsed
   - org.openrewrite.java.testing.junit5.UpgradeOkHttpMockWebServer
+  - org.openrewrite.java.testing.junit5.AddParameterizedTestAnnotation
+  - org.openrewrite.java.testing.junit5.RemoveDuplicateTestTemplates
```
timtebeek, line 85: Hmm; I'd have placed these under `JUnit5BestPractices`, as they are not specific to the JUnit4 to 5 migration; it's just as valid to replace these for people that add both on `@Test @RepeatedTest void test(){...}`. Care to update that to remove them here and move them above line 29?

PR #363, opened 2023-06-23
refactor: Common static analysis issues

knutwannheden:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.staticanalysis.CommonStaticAnalysis


knutwannheden:
Either the recipe doesn't check the language level or it is attributed incorrectly during the ingestion using the Gradle plugin.

```
/home/runner/work/rewrite-testing-frameworks/rewrite-testing-frameworks/src/main/java/org/openrewrite/java/testing/cleanup/TestsShouldIncludeAssertions.java:153: error: cannot infer type arguments for JavaIsoVisitor<P>
            JavaIsoVisitor<AtomicBoolean> findAssertionVisitor = new JavaIsoVisitor<>() {
                                                                                   ^
  reason: '<>' with anonymous inner classes is not supported in -source 8
    (use -source 9 or higher to enable '<>' with anonymous inner classes)
  where P is a type-variable:
    P extends Object declared in class JavaIsoVisitor
```

PR #366, opened 2023-06-26
Fix `HamcrestMatcherToAssertJ` when applied to arrays

knutwannheden:
Fixes: #364


timtebeek (approved):

timtebeek:
Thanks a lot! I had indeed missed that `anyArray`; silly me. :)

PR #368, opened 2023-06-27
Changed type to functional interface 

joanvr:
## What's changed?
The recipe UpdateTestAnnotation was causing a `ClassCastException` because the generated code in a JavaTemplate was invalid. It was trying to assign a lambda expression to an `Object` variable. It seems that this is possible in general, since it's done in the tests and those are passing. However, when trying to apply the recipe to the spring-petclinic workshop, it was causing this exception in a test, and causing the recipe to fail to apply changes in this specific file.
Changing the type of the variable to a functional interface (`org.junit.jupiter.api.function.Executable` has been chosen since it makes sense in the context of the recipe, actually, we cast the lambda expression to this type afterwards), causes no harm to the already passing tests without any change, and also works well on the spring-petclinic workshop.

## What's your motivation?
To have a clean run execution of the recipe on the spring-petclinic repository, and probably others too.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
@@ -121,7 +121,7 @@ public class UpdateTestAnnotation extends Recipe {
                             .visitNonNull(m, ctx, getCursor().getParentOrThrow());
                 }
                 if (cta.expectedException != null) {
-                    m = JavaTemplate.builder("Object o = () -> #{};")
+                    m = JavaTemplate.builder("org.junit.jupiter.api.function.Executable o = () -> #{};")
                             .contextSensitive()
                             .build()
                             .apply(


src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
```diff
@@ -121,7 +121,7 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
                             .visitNonNull(m, ctx, getCursor().getParentOrThrow());
                 }
                 if (cta.expectedException != null) {
-                    m = JavaTemplate.builder("Object o = () -> #{};")
+                    m = JavaTemplate.builder("org.junit.jupiter.api.function.Executable o = () -> #{};")
                             .contextSensitive()
```
knutwannheden, line 125: I haven't tested or debugged this, but I would have expected the template to require a parser with JUnit on the classpath, so that the type attribution is correct.

joanvr: Do you think it's really needed here since we are throwing away the variable and just keeping the lambda expression and "re-assigning" the type couple of lines after?

knutwannheden: I think the error you saw was already solved by a commit that I had made: I added a missing semicolon at the end. But the fact that the variable type isn't really compatible with a lambda in the initializer makes me wonder what we get as type attribution. I will take a closer look tomorrow.

joanvr: You are completely right, I did not realize about the extra semicolon between 2.0.2 and the last commit on main... and that's the reason why it's working fine, not the functional interface... I'm closing the PR

Issue #216, opened 2022-04-04
RSPEC-3658:Unit tests should throw exceptions


yeikel:
See https://rules.sonarsource.com/java/tag/junit/RSPEC-3658

tkvangorder:
For those contributors that are new to OpenRewrite, please note that there are two additional things to be added to the cleanup recipes: 

Override the time estimate for the issue (per occurrence)

```
    @Override
    public @Nullable Duration getEstimatedEffortPerOccurrence() {
        return Duration.ofMinutes(n);
    }
```

Make sure to label the recipe with the RSPEC number: 
    
```
    @Override
    public Set<String> getTags() {
        return Collections.singleton("RSPEC-3658");
    }

```

PR #370, opened 2023-06-29
Remove try-catch-fail blocks from unit tests

AlekSimpson:
## What's changed?
This PR adds a new recipe to move exception to the method signature when there is a try catch block inside of a unit test method. 

## What's your motivation?
[#216](https://github.com/openrewrite/rewrite-testing-frameworks/issues/216)

I left a comment marking this but I see that the `withThrows()` method takes in a list of `NameTree` types. I can't seem to find how to get that `NameTree` type from the `J.Catch` block. I thought maybe the `.getExceptions()` method is what I should use but that returns an `Expression`. Is there maybe some way to convert that to a `NameTree`? 

@timtebeek 

## Any additional context
RSPEC-3658

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
@@ -54,63 +54,24 @@ public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
     }
 
     private static class RemoveTryCatchBlocksFromUnitsTestsVisitor extends JavaVisitor<ExecutionContext> {
-        private final String KEY = "CHANGES_NEEDED";
-        @Override
-        public J visitMethodDeclaration(J.MethodDeclaration md, ExecutionContext ctx) {
-            J.MethodDeclaration m = (J.MethodDeclaration) super.visitMethodDeclaration(md, ctx);
-            if (md.getBody() == null) {
-                return md;
-            }
-
-            List<Statement> statements = md.getBody().getStatements();
-            // check if method is Unit test
-            List<J.Annotation> annotations = md.getAllAnnotations();
-            if (annotations.stream().noneMatch(ann -> MATCHERS.stream().anyMatch(matcher -> matcher.matches(ann)))) {
-                return md;
-            }
-
-            // find try catch block
-            List<J.Try> tryStatements = statements.stream()
-                    .filter(s -> s instanceof J.Try)
-                    .map(s -> (J.Try) s)
-                    // not sure how to handle try-catch blocks with multiple catches
-                    .filter(t -> t.getCatches().size() == 1)
-                    .collect(Collectors.toList());
-            if (tryStatements.size() == 0) {
-                return md;
-            }
-
-            getCursor().putMessage(KEY, md);
-
-            return (J.MethodDeclaration) super.visitMethodDeclaration(md, ctx);
-        }
-
         @Override
         public J visitTry(J.Try try_, ExecutionContext ctx) {
             J.Try t = (J.Try) super.visitTry(try_, ctx);
 
-            Cursor c = getCursor().dropParentWhile(is -> is instanceof J.Block ||
-                                                         !(is instanceof Tree) ||
-                                                         is instanceof J.Try);
-
-            if (c.getMessage(KEY) != null) {
-                if (try_.getCatches().get(0).getBody().getStatements().stream().noneMatch(
-                        s -> s instanceof J.MethodInvocation && ASSERT_FAIL_MATCHER.matches((J.MethodInvocation) s)
-                )) {
-                    return try_;
-                }
-
-                // replace method body
-                maybeRemoveImport("org.junit.Assert");
-                maybeAddImport("org.junit.jupiter.api.Assertions");
-                return JavaTemplate.builder("Assertions.assertDoesNotThrow(() -> { #{any()} })")
-                        .imports("org.junit.jupiter.api.Assertions")
-                        .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "junit-jupiter-api-5.9"))
-                        .build()
-                        .apply(getCursor(), t.getCoordinates().replace(), t.getBody().getStatements().get(0));
+            if (t.getCatches().get(0).getBody().getStatements().stream().noneMatch(
+                    s -> s instanceof J.MethodInvocation && ASSERT_FAIL_MATCHER.matches((J.MethodInvocation) s)
+            )) {
+                return t;
             }
 
-            return t;
+            // replace method body
+            maybeRemoveImport("org.junit.Assert");
+            maybeAddImport("org.junit.jupiter.api.Assertions");
+            return JavaTemplate.builder("Assertions.assertDoesNotThrow(() -> {#{any()}})")
+                    .imports("org.junit.jupiter.api.Assertions")
+                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "junit-jupiter-api-5.9"))
+                    .build()
+                    .apply(getCursor(), t.getCoordinates().replace(), t.getBody());
         }
     }
 }
--- a/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
@@ -17,10 +17,11 @@ package org.openrewrite.java.testing.junit5;
 
 import org.openrewrite.*;
 import org.openrewrite.java.*;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.*;
 
 import java.util.*;
-import java.util.stream.Collectors;
 
 public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
     private static final MethodMatcher ASSERT_FAIL_MATCHER = new MethodMatcher("org.junit.Assert fail(..)");
@@ -43,12 +44,16 @@ public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new RemoveTryCatchBlocksFromUnitsTestsVisitor();
+        return Preconditions.check(new UsesMethod<>("org.junit.Assert fail(..)", false), new RemoveTryCatchBlocksFromUnitsTestsVisitor());
     }
 
     private static class RemoveTryCatchBlocksFromUnitsTestsVisitor extends JavaVisitor<ExecutionContext> {
         @Override
         public J visitTry(J.Try try_, ExecutionContext ctx) {
+            if (try_.getResources() != null) {
+                return try_;
+            }
+
             if (try_.getCatches().get(0).getBody().getStatements().stream().noneMatch(
                     s -> s instanceof J.MethodInvocation && ASSERT_FAIL_MATCHER.matches((J.MethodInvocation) s)
             )) {
--- a/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
@@ -57,15 +57,12 @@ public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
         private final String KEY = "CHANGES_NEEDED";
         @Override
         public J visitMethodDeclaration(J.MethodDeclaration md, ExecutionContext ctx) {
+            J.MethodDeclaration m = (J.MethodDeclaration) super.visitMethodDeclaration(md, ctx);
             if (md.getBody() == null) {
                 return md;
             }
 
             List<Statement> statements = md.getBody().getStatements();
-            if (statements.stream().noneMatch(s -> s instanceof J.Try)) {
-                return md;
-            }
-
             // check if method is Unit test
             List<J.Annotation> annotations = md.getAllAnnotations();
             if (annotations.stream().noneMatch(ann -> MATCHERS.stream().anyMatch(matcher -> matcher.matches(ann)))) {
@@ -83,13 +80,7 @@ public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
                 return md;
             }
 
-            for (J.Try try_ : tryStatements) {
-                if (try_.getCatches().get(0).getBody().getStatements().stream().anyMatch(
-                        s -> s instanceof J.MethodInvocation && ASSERT_FAIL_MATCHER.matches((J.MethodInvocation)s)
-                )) {
-                    getCursor().putMessage(KEY, md);
-                }
-            }
+            getCursor().putMessage(KEY, md);
 
             return (J.MethodDeclaration) super.visitMethodDeclaration(md, ctx);
         }
@@ -103,8 +94,12 @@ public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
                                                          is instanceof J.Try);
 
             if (c.getMessage(KEY) != null) {
+                if (try_.getCatches().get(0).getBody().getStatements().stream().noneMatch(
+                        s -> s instanceof J.MethodInvocation && ASSERT_FAIL_MATCHER.matches((J.MethodInvocation) s)
+                )) {
+                    return try_;
+                }
 
-                System.out.println(t.getBody().getStatements().get(0));
                 // replace method body
                 maybeRemoveImport("org.junit.Assert");
                 maybeAddImport("org.junit.jupiter.api.Assertions");
--- a/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
@@ -24,13 +24,6 @@ import java.util.stream.Collectors;
 
 public class RemoveTryCatchBlocksFromUnitTests extends Recipe {
     private static final MethodMatcher ASSERT_FAIL_MATCHER = new MethodMatcher("org.junit.Assert fail(..)");
-    private static final List<String> TEST_PATTERNS = Arrays.asList(
-            "@org.junit.jupiter.api.Test",
-            "@org.junit.jupiter.api.RepeatedTest",
-            "@org.junit.jupiter.params.ParameterizedTest"
-    );
-
-    private static final List<AnnotationMatcher> MATCHERS = TEST_PATTERNS.stream().map(AnnotationMatcher::new).collect(Collectors.toList());
 
     @Override
     public String getDisplayName() {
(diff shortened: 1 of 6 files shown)


sambsnyd (commented):
JUnit reports assertion failures differently from uncaught exceptions. 
Uncaught exceptions are considered a worse failure mode as they can indicate that the test itself failed to evaluate its criteria. So this could unintentionally increase the apparent severity of test failures.

In JUnit 5 the idiomatic way of dealing with a `catch` containing a `fail` would be to translate into `Assertions.assertDoesNotThrow()`.

AlekSimpson:
> Assertions.assertDoesNotThrow()

So if I am understanding this correctly instead of getting rid of the `try-catch` block and changing the method signature, it would be better to instead change the `Assert.fail()` to `Assertions.assertDoesNotThrow()`?

AlekSimpson:
Ok the latest changes almost complete the recipe. There are only two issues remaining; The first being that the `ASSERT_FAIL_MATCHER` does not match `Assert.fail()` and just always returns false. I suspect maybe I am missing something in the classpath resources because I am not sure why else this matcher would not be working. The second issue is that for some reason the recipe does not add the import statement for `Assertions.doesNotThrow()`. Maybe this also has something to do with the classpath resources, I am not sure.

src/test/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTestsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Test
+    @DocumentExample
+    void removeTryCatchBlock() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import static org.junit.Assert;
```
knutwannheden, line 46:
```suggestion
              import org.junit.Assert;
```

src/test/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTestsTest.java
```diff
@@ -0,0 +1,100 @@
...
+              import org.junit.jupiter.api.Test;
+              
+              class Test {
```
knutwannheden, line 49: This class name conflicts with the imported `Test` type two lines up.

```suggestion
              class MyTest {
```

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
```diff
@@ -0,0 +1,120 @@
...
+        public J.MethodInvocation build(J.MethodInvocation mi) {
+            Expression argument = mi.getArguments().get(0);
+            if (mi.getArguments().get(0) instanceof J.MethodInvocation){
+                argument = ((J.MethodInvocation) mi.getArguments().get(0)).getSelect();
+            }
+
+            maybeRemoveImport("org.junit.Assert");
+            maybeAddImport("org.junit.jupiter.api.Assertions");
+            return JavaTemplate.builder("Assertions.assertDoesNotThrow(#{any()})")
+                    .staticImports("org.junit.jupiter.api.Assertions")
```
knutwannheden, line 101:
```suggestion
                    .imports("org.junit.jupiter.api.Assertions")
```

knutwannheden: What you are also missing here is a call to `javaParser()` to make it aware of the JUnit Jupiter types:

```java
javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "junit-jupiter-api-5.9"))
```

Please also be aware that `assertDoesNotThrow()` works differently: It is used to wrap a code block. Typically you would let it wrap a lambda containing the statements which you here have in the try-block. See https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable) for details.

knutwannheden (commented):
@AlekSimpson I added a few hints which should allow you to make progress.

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
```diff
@@ -68,53 +67,34 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration md, Execut
...
-        @Override
-        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
-            J.MethodInvocation m = super.visitMethodInvocation(mi, ctx);
-            Cursor c = getCursor().dropParentWhile(is -> is instanceof J.Block
-                                                         || is instanceof J.Try.Catch
-                                                         || is instanceof J.Try
-                                                         || !(is instanceof Tree));
-            if (c.getMessage("KEY") != null) {
-                return build(m);
+            for (Statement statement : tryStatements) {
```
knutwannheden, line 75: This loop over only the `try` statements of the method's body looks a bit strange to me. Can you add a test with a method containing some more statements before and / or after the `try` statement to make sure that it also then behaves correctly?

AlekSimpson: While adjusting for this I realized, what if there was a case where there were multiple try-catch blocks that needed to be replaced? Is that a case I should be thinking about? If so, can these kinds of changes be made through a `visitTry` visitor? I keep getting errors when I try making changes with this visitor. Seems like these issues could be stemming from the fact that whats being returned is a method and not a `J.Try`.

timtebeek: Sounds like you need to switch from an isomorphic visitor (JavaIsoVisitor) to the non isomorphic JavaVisitor , such that you can return a different type of result.

knutwannheden: Yes, `visitTry()` with a visitor extending `JavaVisitor` rather than `JavaIsoVisitor` is the way forward. That way `visitTry()` can return any other statement (like `J.MethodInvocation` in your case) and it will automatically replace the `J.Try` statement.

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
```diff
@@ -53,48 +53,60 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
         return new RemoveTryCatchBlocksFromUnitsTestsVisitor();
     }
 
-    private static class RemoveTryCatchBlocksFromUnitsTestsVisitor extends JavaIsoVisitor<ExecutionContext> {
+    private static class RemoveTryCatchBlocksFromUnitsTestsVisitor extends JavaVisitor<ExecutionContext> {
         @Override
         public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration md, ExecutionContext ctx) {
+            if (md.getBody() == null) {
```
knutwannheden, line 59: Before returning anything, you should add a `super.visitMethodDeclaration()` call, because the `J.Try` could be nested inside a block or loop or so.

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
```diff
@@ -53,48 +53,60 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                return md;
+            }
+
+            List<Statement> statements = md.getBody().getStatements();
+            if (statements.stream().noneMatch(s -> s instanceof J.Try)) {
```
knutwannheden, line 64: We should remove this `if` block, since the try could be nested.

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
```diff
@@ -53,48 +53,60 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                        // replace method body
-                        maybeRemoveImport("org.junit.Assert");
-                        maybeAddImport("org.junit.jupiter.api.Assertions");
-                        return JavaTemplate.builder("Assertions.assertDoesNotThrow(() -> #{any()} )")
-                                .imports("org.junit.jupiter.api.Assertions")
-                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "junit-jupiter-api-5.9"))
-                                .build()
-                                .apply(getCursor(), md.getCoordinates().replaceBody(), tryBody);
-                    }
+            for (J.Try try_ : tryStatements) {
```
knutwannheden, line 84: Most of this code can then be moved down into the `visitTry()` method. Then you don't need any looping either.

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
AlekSimpson: Ah ok, so that will not interfere with the cursor messaging? I can have a super call both at the start and at the last return?

knutwannheden: Maybe I didn't check carefully enough. I just think we need to make sure it gets called at least once per code path in this recipe.

AlekSimpson:
@knutwannheden, Ok I made those changes you suggested. One thing though that I am still not sure about here is that two of the tests are not passing because its throwing an error when it reaches the `JavaTemplate.builder()` call in the `J.Try` visitor. It says: "Expected a template that would generate exactly one statement to replace one statement, but generated 2." I haven't seen something like this when using a template builder.

knutwannheden:
Looking at the code I would probably get rid of the `visitMethodDeclaration()` method entirely and also the cursor messaging. I understand that it is there to make sure that we only modify code in tests. But I think if anybody is using JUnit in non test code then they can simply not run this recipe over that code.

I am also saying this because I know that test code is frequently made more modular by factoring out some assertions into helper methods. And then these would not get refactored. So I suggest we remove the matchers and leave the `visitTry()` only. But let's also see if @timtebeek has any good arguments against this simplification.

Meanwhile I will look at your template problem. It could be a bug in the template engine.

src/test/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTestsTest.java
```diff
@@ -0,0 +1,145 @@
...
+                          Assert.fail(e.getMessage());
+                      }
+                      System.out.println("statements after");
+                      int y = 50;
+                      int z = x + y;
+                  }
+              }
+              """,
+            """
+              import org.junit.Assert;
```
knutwannheden, line 125: By removing the `visitMethodDeclaration()` override as I suggested, I only had to fix this one import to get the test to pass.
```suggestion
              import org.junit.jupiter.api.Assertions;
```

timtebeek:
> I am also saying this because I know that test code is frequently made more modular by factoring out some assertions into helper methods. And then these would not get refactored. So I suggest we remove the matchers and leave the `visitTry()` only. But let's also see if @timtebeek has any good arguments against this simplification.

Skimmed through a bit, but I'm fine replacing any `try { ..1.. } catch ( ..2.. ) { fail(..3..); }` anywhere, inside or outside tests, with `Assertions.assertDoesNotThrow(() -> { ..1.. })`.

I do wonder a bit if we should do anything about the potential information loss in `..2..` and `..3..`; in the current setup we lose `..2..` and `..3..`.
```diff
-                      try {
-                          int divide = 50 / 0;
-                      }catch (ArithmeticException e) {
-                          Assert.fail(e.getMessage());
-                      }
+                      Assertions.assertDoesNotThrow(() -> {
+                          int divide = 50 / 0;
+                      });
```
Perhaps we should adopt either of these to retain the failure message? Which then can't use the exception anymore(!).
[assertDoesNotThrow](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable,java.lang.String))​([Executable](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/function/Executable.html) executable, [String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html) message)
[assertDoesNotThrow](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable,java.util.function.Supplier))​([Executable](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/function/Executable.html) executable, [Supplier](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/Supplier.html)<[String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html)> messageSupplier)

AlekSimpson:
> > I am also saying this because I know that test code is frequently made more modular by factoring out some assertions into helper methods. And then these would not get refactored. So I suggest we remove the matchers and leave the `visitTry()` only. But let's also see if @timtebeek has any good arguments against this simplification.
> 
> Skimmed through a bit, but I'm fine replacing any `try { ..1.. } catch ( ..2.. ) { fail(..3..); }` anywhere, inside or outside tests, with `Assertions.assertDoesNotThrow(() -> { ..1.. })`.
> 
> I do wonder a bit if we should do anything about the potential information loss in `..2..` and `..3..`; in the current setup we lose `..2..` and `..3..`.
> 
> ```diff
> -                      try {
> -                          int divide = 50 / 0;
> -                      }catch (ArithmeticException e) {
> -                          Assert.fail(e.getMessage());
> -                      }
> +                      Assertions.assertDoesNotThrow(() -> {
> +                          int divide = 50 / 0;
> +                      });
> ```
> 
> Perhaps we should adopt either of these to retain the failure message? Which then can't use the exception anymore(!). [assertDoesNotThrow](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable,java.lang.String))​([Executable](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/function/Executable.html) executable, [String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html) message) [assertDoesNotThrow](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable,java.util.function.Supplier))​([Executable](https://junit.org/junit5/docs/5.8.2/api/org.junit.jupiter.api/org/junit/jupiter/api/function/Executable.html) executable, [Supplier](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/Supplier.html)<[String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html)> messageSupplier)

Ya I've been wondering about that too. Also try-catch blocks could have multiple catch blocks I believe. Something like `try { ..1.. } catch ( ..2.. ) { fail(..3..); } catch ( ..4.. ) { fail(..5..); }`

AlekSimpson:
@knutwannheden when you get a chance to look into the possible template engine issue, could you let me know how that goes?

knutwannheden:
I just had another look. With the current code I cannot detect any exceptions thrown by the test. So it is hard to determine if there is indeed a bug or just a usage error.

Regarding the discussion with the catch blocks: I think we should maybe instead migrate to AssertJ here, as it offers more options which don't lose these details. For example:
```java
assertThatCode(() -> {
  throw new RuntimeException("e");
}).isNotInstanceOf(IllegalArgumentException.class);
```
or when we don't want any exception:
```java
assertThatCode(() -> {
  throw new RuntimeException("e");
}).doesNotThrowAnyException();
```
or for when the code is expected to throw an exception and we want to make an assertion about it:
```java
assertThatThrownBy(() -> {
  throw new RuntimeException("e");
}).isInstanceOf(RuntimeException.class);
```

AlekSimpson:
Ok the template error issue is fixed, I just needed to add `.contextSensitive()` to the template. The tests now all pass now.

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new RemoveTryCatchBlocksFromUnitsTestsVisitor();
+    }
+
+    private static class RemoveTryCatchBlocksFromUnitsTestsVisitor extends JavaVisitor<ExecutionContext> {
+        @Override
+        public J visitTry(J.Try try_, ExecutionContext ctx) {
+            if (try_.getCatches().get(0).getBody().getStatements().stream().noneMatch(
```
timtebeek, line 52: - [ ] We seem to imply here that we only ever look at the first `catch` block;
  I'd appreciate a test that uses two catch blocks, and appropriate handling (skip?)
- [ ] Have you given any thought to try-with-resources blocks? We would not want to remove those.
  Might be good to add in a test that uses `try (Foo bar) { ... } catch (Exception e) { fail(); }`

timtebeek (changes requested):

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
AlekSimpson: Yes I am not sure how we should handle try statements with multiple catch blocks. Did you see Knut's comment about using AssertJ instead? What are your thoughts on that?

timtebeek: There might be value in having a separate recipe for AssertJ, but not everyone is ready or willing to use that just yet. To me it then makes the most sense to check here that there is:
1. only one catch block, such that we know it's safe to apply this recipe
2. only one statement in the catch block, which is a `fail()`, with no or a simple String argument. We would not want to convert for instance `fail(cleanUpAndReturnMessage())` might still have side effects that we don't want to remove.

timtebeek: Hmm; considering number two above: `Assert.fail(e.getMessage())` should be fine.. It's not easy thinking of each possible scenario 😅

src/test/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTestsTest.java
```diff
@@ -74,6 +74,28 @@ public void testMethod() {
...
+            """
+              import org.junit.jupiter.api.Test;
+              import java.io.PrintWriter;
+              
+              class MyTest {
+                  @Test
+                  void aTest()  {
+                      try (PrintWriter writer = new PrintWriter("tests.txt")) {
+                          writer.println("hello world");
+                      }
```
timtebeek, line 91: This test might not run the recipe for the wrong reason if we're ever to subtly change the implementation; There's no catch block, and no `fail` in the catch block. Ideally I think this test should prove that even if all other conditions match (catch + fail), then still we don't apply the recipe _only_ because there's a try with resources.
```suggestion
                      } catch (Exception e) {
                          Assert.fail("Some message");
                      }
```

src/test/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTestsTest.java
```diff
@@ -74,6 +74,28 @@ public void testMethod() {
...
     }
 
+    @Test
+    void doesNotRunOnTryWithResources() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.junit.jupiter.api.Test;
+              import java.io.PrintWriter;
```
timtebeek, line 84:
```suggestion
              import java.io.PrintWriter;
              import org.junit.Assert;
```

src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchBlocksFromUnitTests.java
AlekSimpson: > There might be value in having a separate recipe for AssertJ, but not everyone is ready or willing to use that just yet. To me it then makes the most sense to check here that there is:
> 
>     1. only one catch block, such that we know it's safe to apply this recipe
> 
>     2. only one statement in the catch block, which is a `fail()`, with no or a simple String argument. We would not want to convert for instance `fail(cleanUpAndReturnMessage())` might still have side effects that we don't want to remove.

Latest commit addresses this

src/test/java/org/openrewrite/java/testing/junit5/RemoveTryCatchFailBlocksFromUnitTestsTest.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    @Test
+    @DocumentExample
+    void removeTryCatchBlock() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.junit.Assert;
```
timtebeek, line 46: Here and everywhere below the import & use should change to use Assertions from JUnit 5.

timtebeek (changes requested):
There's been a bit of an unfortunate oversight earlier in the use of JUnit 4 Assert, whereas we're more likely to execute this recipe after a JUnit 5 migration, as part of the best practice recipes. Could you update the PR once more to reflect this change?

AlekSimpson:
Ok @timtebeek let me know how these new changes look

timtebeek (approved):

PR #374, opened 2023-07-07
Limit `RemoveIsMatcher` to invocations within `assertThat`

AlekSimpson:
## What's changed?
This PR fixes an issue where the recipe that removed `is()` invocations would remove them even when they were not found directly inside of an `assertThat` invocation.

## What's your motivation?
[#365](https://github.com/openrewrite/rewrite-testing-frameworks/issues/365)

I added two tests based off of the last comment on the ticket. Both tests now pass in this PR, so I think the bug should be fixed. I think how I implemented the fix shouldn't interfere with any other parts of the recipe.

@timtebeek 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
+++ b/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
@@ -98,8 +98,7 @@ class RemoveIsMatcherTest implements RewriteTest {
             """
               import org.junit.jupiter.api.Test;
               import static org.hamcrest.MatcherAssert.assertThat;
-              import static org.hamcrest.Matchers.is;
-              import static org.hamcrest.Matchers.equalTo;
+              import static org.hamcrest.Matchers.*;
               
               class ATest {
                   @Test
--- a/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
+++ b/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
@@ -68,4 +68,28 @@ class RemoveIsMatcherTest implements RewriteTest {
             """));
     }
 
+    @Test
+    void isNotCalledInAssertThat() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.junit.jupiter.api.Test;
+              import static org.hamcrest.Matchers.is;
+              import static org.hamcrest.Matchers.equalTo;
+              
+              class ATest {
+                  @Test
+                  void testMethod() {
+                      String str1 = "Hello world!";
+                      Matcher<String> x = is(equalTo(str1));
+                  }
+              }
+              """,
+            """
+              """
+          )
+        );
+    }
+
 }
--- a/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
+++ b/src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
@@ -113,4 +113,30 @@ class RemoveIsMatcherTest implements RewriteTest {
         );
     }
 
+    @Test
+    void noReplacementForOtherMethodInvocations() {
+        //language=java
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
```diff
@@ -68,4 +68,50 @@ void testEquals() {
...
+              import static org.hamcrest.MatcherAssert.assertThat;
+              import static org.hamcrest.Matchers.is;
+              import static org.hamcrest.Matchers.equalTo;
+              
+              class ATest {
+                  @Test
+                  void testMethod() {
+                      String str1 = "Hello world!";
+                      String str2 = "Hello world!";
+                      assertThat(str1, contains(is(equalTo(str2))));
```
timtebeek, line 109: Perhaps `not` is better here as an example than contains, as the latter does not compile.
```suggestion
                      assertThat(str1, not(is(equalTo(str2))));
```

src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
```diff
@@ -68,4 +68,50 @@ void testEquals() {
...
+                  }
+              }
+              """
```
timtebeek, line 112: When using `not` we might actually want to execute the recipe, because there's still a wrapping `assertThat`.
```suggestion
              """,
            """
              import org.junit.jupiter.api.Test;
              import static org.hamcrest.MatcherAssert.assertThat;
              import static org.hamcrest.Matchers.not;
              import static org.hamcrest.Matchers.equalTo;
              
              class ATest {
                  @Test
                  void testMethod() {
                      String str1 = "Hello world!";
                      String str2 = "Hello world!";
                      assertThat(str1, not(equalTo(str2)));
                  }
              }
              """
```

src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
```diff
@@ -68,4 +68,50 @@ void testEquals() {
...
+          )
+        );
+    }
+
```
timtebeek, line 116: This might be closer to the usage where we do not want replacements to take place.
```suggestion

    @Test
    void noReplacementForOtherMethodInvocations() {
        //language=java
        rewriteRun(
          java(
            """
              import org.junit.jupiter.api.Test;
              import static org.hamcrest.MatcherAssert.assertThat;
              import static org.hamcrest.Matcher;
              import static org.hamcrest.Matchers.*;
              
              class ATest {
                  @Test
                  void testMethod() {
                      String str1 = "Hello world!";
                      String str2 = "Hello world!";
                      foo(is(equalTo(str2)));
                  }
                  
                  void foo(Matcher<String> matcher) {
                  }
              }
              """
          )
        );
    }
```

timtebeek (changes requested):
Quite some suggestions; think when we apply these we ought to be done with this change. Thanks for getting this started!

src/test/java/org/openrewrite/java/testing/hamcrest/RemoveIsMatcherTest.java
AlekSimpson: I see, so `is()` should be replaced when it is nested inside certain function calls that are inside `assertThat()`? Which methods like this do we want the recipe to execute on besides not?

timtebeek: It's not specific to `not` or any other matcher; we want to remove `is(..)` whenever that's used within an `assertThat` method call (however deep), but not when used outside an `assertThat` call. 

We add this limitation to this recipe because hamcrest matchers are also sometimes used in different frameworks such as restassured; our migration recipes do not cover those cases, and we can't be certain it's safe or desirable to remove `is` method invocations there.

timtebeek (approved):

Issue #357, opened 2023-06-18
Further Hamcrest to AssertJ migration recipes


timtebeek:
## What problem are you trying to solve?
After https://github.com/openrewrite/rewrite-testing-frameworks/pull/355 we already support some ~50 Hamcrest Matcher to AssertJ migrations. Not all Hamcrest Matchers are covered however, for a potentially incomplete (but non-breaking) migration. Below is a quick overview of cases we might want to cover still, depending on how common they are, and the effort needed to convert those automatically.

### Sub matchers `allOf(Matcher...)`, `anyOf(Matcher...)`, ...
Matchers that take subsequent matchers as argument are not yet covered, and likely need a dedicated recipe for each case. Here's a likely quick overview of such Matchers:

```java
allOf(Iterable<Matcher<? super T>> matchers)
anyOf(Iterable<Matcher<? super T>> matchers)
both(Matcher<? super LHS> matcher)
either(Matcher<? super LHS> matcher)
everyItem(Matcher<U> itemMatcher)
hasItem(Matcher<? super T> itemMatcher)
hasItems(Matcher<? super T>... itemMatchers)
aMapWithSize(Matcher<? super Integer> sizeMatcher)
hasSize(Matcher<? super Integer> sizeMatcher)
contains(Matcher<? super E> itemMatcher)
contains(Matcher<? super E>... itemMatchers)
containsInAnyOrder(Matcher<? super T>... itemMatchers)
containsInRelativeOrder(Matcher<? super E>... itemMatchers)
iterableWithSize(Matcher<? super Integer> sizeMatcher)
hasEntry(Matcher<? super K> keyMatcher, Matcher<? super V> valueMatcher)
hasKey(Matcher<? super K> keyMatcher)
hasValue(Matcher<? super V> valueMatcher)
hasToString(Matcher<? super String> toStringMatcher)
```

### More inverted matches with `not`
There's likely more direct replacement options in AssertJ for some of the inverted `not(Matcher)` that we don't already cover. Not all will have direct replacements, and for those cases we might need more complicated constructs.

```java
not(Matcher<T> matcher)
```

### Matchers for arrays

```java
array(Matcher<? super T>... elementMatchers)
arrayContaining(List itemMatchers)
arrayContaining(Matcher<? super E>... itemMatchers)
arrayContainingInAnyOrder(Collection itemMatchers)
arrayContainingInAnyOrder(Matcher<? super E>... itemMatchers)
[...]

PR #375, opened 2023-07-09
Flatten Hamcrest assertThat allOf

timtebeek:
## What's changed?
Convert Hamcrest `allOf(Matcher...)` to individual `assertThat` statements for easier migration.

## What's your motivation?
For #357

Correct use of cursor & coordinates.

Initially tried with `visitMethodInvocation`, but that also complained when inserting more than one statement, even when part of the same JavaTemplate string.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/hamcrest/FlattenAllOf.java
+++ b/src/main/java/org/openrewrite/java/testing/hamcrest/FlattenAllOf.java
@@ -1,0 +1,81 @@
+package org.openrewrite.java.testing.hamcrest;
+
+import org.openrewrite.*;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaCoordinates;
+import org.openrewrite.java.tree.Statement;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class FlattenAllOf extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Convert Hamcrest `allOf(Matcher...)` to individual `assertThat` statements";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Convert Hamcrest `allOf(Matcher...)` to individual `assertThat` statements for easier migration.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(
+                new UsesMethod<>("org.hamcrest.Matchers allOf(..)"),
+                new FlattenAllOfVisitor());
+    }
+}
+
+class FlattenAllOfVisitor extends JavaVisitor<ExecutionContext> {
+    private final MethodMatcher ASSERT_THAT_MATCHER = new MethodMatcher("org.hamcrest.MatcherAssert assertThat(..)");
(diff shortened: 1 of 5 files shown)


timtebeek:
@AlekSimpson I'd started on this one but it's late and I don't quite know right now how to resolve the test failures; perhaps you've solved this elsewhere already by now? Otherwise we can ask someone else to have a look.

src/main/java/org/openrewrite/java/testing/hamcrest/FlattenAllOf.java
```diff
@@ -0,0 +1,81 @@
...
+                        }
+                    }
+
+                }
+            }
+            // Keep the original statement
+            newStatements.add(statement);
+        }
+
+        return md.withBody(md.getBody().withStatements(newStatements));
```
timtebeek, line 79: Might want to add a conditional here to only return a different element if there were changes to newStatements.

knutwannheden: The code should instead be using `ListUtils#map()`. That method automatically figures out if there were any changes to any of the elements and if not returns the original input list. That is nice because it avoids unnecessarily allocating a new list.

Note: In this case the `J.Body#withStatements()` will not be able to simply return `this`, just because the list is still the same instance. It will still end up checking all the elements for referential equality. But if all elements are indeed the same, then the `withStatements()` method will return `this`. So a conditional boolean variable is IMHO unnecessary, if using `ListUtils#map()` here.

AlekSimpson:
> @AlekSimpson I'd started on this one but it's late and I don't quite know right now how to resolve the test failures; perhaps you've solved this elsewhere already by now? Otherwise we can ask someone else to have a look.

I don't believe I've seen this error before. I looked at the code and it seems like you are passing null into `template.apply(...)`. The tests seem to failing for that reason. I tried just passing in `getCursor()` but that didn't work. Is there some way maybe to get the cursor of the `assertThat` invocation? Maybe one of those cursors would work.

AlekSimpson:
@timtebeek Ok so I thought of a different approach that maybe would be simpler. Basically it uses the `visitMethodInvocation` visitor to find `assertThat` instances with `allOf` and then it loops through each matcher argument in `allOf` and adds a new `assertThat(...)` statement string to a `StringBuilder` while also updating an array with the two corresponding arguments (`actual` and the matcher). Then after that it just applies the template with the array. I thought this might be easier since now you don't have to worry about what coordinates or cursor each time you apply a new template. Only problem is that I am again running into another error where it says it was expecting one statement to replace one statement but got 2. I tried using `.contextSensitive()` which is what resolved this issue last time I bumped into it but that doesn't seem to be doing the trick. I'm honestly not sure why else this error would be given. I thought `JavaVisitor` should allow for it.

timtebeek:
Thanks for picking this up! Generating multiple statements at once is an approach I had also tried indeed, but got the same issues that you're having. Would need to see if we recall this pattern used elsewhere, but didn't manage to fit that in at the time. 🤔

timtebeek:
Thanks for the trigger to rethink this! I now remembered that I had done something similar a while ago in cucumber-jvm:
https://github.com/openrewrite/rewrite-cucumber-jvm/blob/53ccf2c566c2398edde82a9dcb98f0462326686e/src/main/java/org/openrewrite/cucumber/jvm/CucumberJava8ClassVisitor.java#L72C1-L73

The JavaTemplate can only return a `J`, but not a collection of statements. By wrapping those statements in a block and then afterwards cleaning up the block we achieve what we want, with a slight detour. Makes the code much simpler, so thanks again!

src/main/java/org/openrewrite/java/testing/hamcrest/FlattenAllOf.java
timtebeek: thanks again; solved differently by reverting to visitMethodInvocation, and returning a block of statements there which we unwrap after visit.

PR #371, opened 2023-07-05
fix: assertj contains is weaker than hamcrests contains

davidburkhart:
## What's changed?

Hamcrest's `contains` matches AssertJ's `containsExactly`. AssertJs `contains` does not care about order and completeness. Hamcrests `contains` does, see the [docs](https://hamcrest.org/JavaHamcrest/javadoc/2.1/org/hamcrest/Matchers.html#contains-E...-)!

## What's your motivation?

Fix this bug.

## Anything in particular you'd like reviewers to focus on?

Should I create an issue or is just the pr fine?

Another topic: I am not so sure about the mapping of [containsInRelativeOrder to containsExactly](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/resources/META-INF/rewrite/hamcrest.yml#L148C30-L148C30) as I don't know how `containsInRelativeOrder` works. From the docs I don't understand if it matches for completeness in both directions. Kind of related to this pr, but also a different topic.

## Anyone you would like to review specifically?

no

## Have you considered any alternatives or workarounds?

It is a bug to me.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ auto-formatter on affected files
- [x] I've updated the documentation (if applicable)

PR #376, opened 2023-07-10
removed containsInRelativeOrder from yaml list

AlekSimpson:
## What's changed?
The containsInRelativeOrder matcher recipe has been removed from the `hamcrest.yml` list. 

## What's your motivation?
[this issue](https://github.com/openrewrite/rewrite-testing-frameworks/pull/371#issuecomment-1622442507)

@timtebeek 

It seems that there is currently no direct replacement for the hamcrest matcher `containsInRelativeOrder`in AssertJ. The closest method in AssetJ would be `containsExactlyInOrder` but these two methods are not *quite* the same. 

### Checklist


timtebeek:
I'm guessing the test will fail; that needs to updated as well.

timtebeek (approved):

PR #380, opened 2023-07-15
Restore now Scanning AnyToNullable, as it does not affect Mockito 2+

timtebeek:
## What's changed?
Restore AnyToNullable after 7b2c98bf2792ce59afc31ab649ea3c4c1b9db126 and reimplementation in #378 by @nmck257

## What's your motivation?
Recipe should no longer make excessive changes; meaning it can be restored to migrate from Mockito v1 to v2 without semantic changes and potential breaking tests.


timtebeek:
[Signed off on Slack](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1689350674815769?thread_ts=1689339763.545469&cid=C01VADFPJQZ).

PR #379, opened 2023-07-11
More Hamcrest Migrations for Number Matchers

AlekSimpson:
## What's changed?
This PR modifies the existing Hamcrest to AssertJ migration recipe to allow for more number matchers to be migrated. 

## What's your motivation?
[#357](https://github.com/openrewrite/rewrite-testing-frameworks/issues/357)

## Anything in particular you'd like reviewers to focus on?
So I added support for `closeTo` but when I was looking through the list of matchers that related to numbers I couldn't find any others that weren't already supported. Maybe I'm not looking at the right documentation. Let me know if there were any other number matchers you were maybe expecting to see here.

## Anyone you would like to review specifically?
@timtebeek 

## Have you considered any alternatives or workarounds?
At first I had created a dedicated recipe for `closeTo` but then realized the `HamcrestMatcherToAssertJ` recipe did most of what needed to be done already. 

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ auto-formatter on affected files
- [ ] I've updated the documentation (if applicable)

PR #381, opened 2023-07-17
ArrayHamcrestMigrations

AlekSimpson:
## What's changed?
This PR adds a test to test for migrating the `arrayContaing` Hamcrest matcher to AssertJ's contains exactly. This PR will also be updated with more Hamcrest array matchers as I work through them.  

## What's your motivation?
[#357](https://github.com/openrewrite/rewrite-testing-frameworks/issues/357)

It seems like the recipe should be able to pass the test without any extra modifications. However, it is not even running the recipe because the `assertThat` matcher won't recognize `assertThat` in the test. I'm not sure why this would be the case because the recipe is finding `assertThat` just fine with every other test except this one. It must be something wrong with the test but the test seems identical to the others in terms of its use of `assertThat`.

@timtebeek 

### Checklist


timtebeek:
@AlekSimpson your input test sample did not compile; sometimes it helps to make sure that the input compiles correctly when you get mismatches like you saw here. By swapping `int[]` out for `Integer[]` your input compiles, and the test runs successfully.

It's been proposed to actually fail the test if the input already has compilation errors, but that's not yet been implemented unfortunately.

AlekSimpson:
Ok this should be good to merge. Latest commit adds tests that cover more of the Hamcrest Array Matchers. The only things missing from this would be the matchers that have nested matchers as arguments.

timtebeek:
Before these recipes are actually available to users they would have to be [added to this recipeList](https://github.com/openrewrite/rewrite-testing-frameworks/blob/07d3c549974a9130b986b5b29bb6e8ce435078b4/src/main/resources/META-INF/rewrite/hamcrest.yml#L41), ideally under a separate heading for Array matchers like we also have for Object matchers, String matchers, List Matchers, etc.. 

It might be better to test these similar to how we test the other yaml-defined recipes, such as for instance for Lists:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/07d3c549974a9130b986b5b29bb6e8ce435078b4/src/test/java/org/openrewrite/java/testing/hamcrest/MigrateHamcrestToAssertJTest.java#L230-L268

The tests in `src/test/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToAssertJTest.java` are only ever necessary if you need to make changes to that recipe to accommodate a new class of recipes; looks like that might not have been needed here. I'd reported the same in a shorter version last week over on https://github.com/openrewrite/rewrite-testing-frameworks/pull/379#issuecomment-1632087570

Could you update the PR to make those changes?

src/main/resources/META-INF/rewrite/hamcrest.yml
```diff
@@ -87,6 +87,23 @@ recipeList:
       matcher: isA
       assertion: isInstanceOf
 
+  # Array Matchers
+  - org.openrewrite.java.testing.hamcrest.HamcrestMatcherToAssertJ:
+      matcher: arrayContaining
+      assertion: containsExactly
+  - org.openrewrite.java.testing.hamcrest.HamcrestMatcherToAssertJ:
+      matcher: arrayContainingInAnyOrder
+      assertion: containsOnly
```
timtebeek, line 96: As discussed:
> Hamcrest arrayContainingInAnyOrder
Creates an order agnostic matcher for arrays that matches when each item in the examined array is logically equal to one item anywhere in the specified items. For a positive match, the examined array must be of the same length as the number of specified items.

versus

> AssertJ containsOnly
Verifies that the actual group contains only the given values and nothing else, in any order and ignoring duplicates (i.e. once a value is found, its duplicates are also considered found).
> If you need to check exactly the elements and their duplicates use:
containsExactly(Object...) if the order does matter
containsExactlyInAnyOrder(Object...) if the order does not matter

which I'm not 100% sure, but sounds like maybe it should be contansExactlyInAnyOrder? or what do you think there?

We might want to similarly check `hasItemInArray` versus `contains` below for any semantic differences.

timtebeek (commented):
Request to double check the replacements such that we don't have any semantic changes that might reduce othe test fidelity or even break.

timtebeek (approved):

PR #382, opened 2023-07-17
Alek/remove try catch blocks fix

AlekSimpson:
## What's changed?
The new try catch recipe was not handling cases where binary expressions were being passed into `Assertions.fail` correctly. This PR fixes that.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchFailBlocks.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/RemoveTryCatchFailBlocks.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.testing.junit5;
 
+import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


timtebeek:
@AlekSimpson As you can see in cb80173db251be5e4c52ff8180506da1716ba51e and 6f2b19b329a5cfb4446b6589d57c692924971c80 I've swapped the logic around to only recognize and apply changes in known cases, rather than what we had previously where might inadvertently make changes by only returning early for known cases. This in my view improves the predictability of the recipe, such that we should be able to safely run it at scale. We now also cover some additional cases, and handle some cases differently, for instance by only retaining the left side of a binary expression, and covering `fail(.., Throwable)`. Hope you agree with those changes, and will apply the same pattern of only making changes for known cases to future recipes.

@joanvr would you be willing to review for a second pair of eyes on my changes?

PR #384, opened 2023-07-19
Migrate assertEquals with array parameter to assertArrayEquals

puskarpeter:
## What's changed?

Added a test case to cover the mentioned use case.

## What's your motivation?

- assertEquals in JUnit4 accepts Object[], to preserve the behaviour we need to rewrite it into assertArrayEquals
https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertEquals(java.lang.Object%5B%5D,%20java.lang.Object%5B%5D)
### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java
@@ -462,4 +462,35 @@ class AssertToAssertionsTest implements RewriteTest {
           )
         );
     }
+
+    // edge case for deprecated use of assertEquals
+    // https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertEquals(java.lang.Object%5B%5D,%20java.lang.Object%5B%5D)
+    @Issue("#383")
+    @Test
+    void assertEqualsWithArrayArgumentToAssertArrayEquals() {
+        //language=java
+        rewriteRun(
+          spec -> spec.typeValidationOptions(TypeValidation.none()),
+          java(
+            """
+              import org.junit.Assert;
+              
+              class MyTest {
+                  void test() {
+                       Assert.assertEquals(new Object[1], new Object[1]);
+                  }
+              }
+              """,
+            """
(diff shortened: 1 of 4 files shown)


src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java
```diff
@@ -462,4 +462,35 @@ void test() {
           )
         );
     }
+
+    // edge case for deprecated use of assertEquals
+    // https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertEquals(java.lang.Object%5B%5D,%20java.lang.Object%5B%5D)
+    @Issue("#383")
```
timtebeek, line 468:
```suggestion
    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/pull/384")
```

src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java
```diff
@@ -462,4 +462,35 @@ void test() {
...
+    @Test
+    void assertEqualsWithArrayArgumentToAssertArrayEquals() {
+        //language=java
+        rewriteRun(
+          spec -> spec.typeValidationOptions(TypeValidation.none()),
```
timtebeek, line 473:
```suggestion
```
Ideally this is not necessary; in rare cases where we want to make changes despite missing types we add that to some tests, but it could potentially break recipe chaining, as the next recipe will then not have the correct types available.

puskarpeter: Oh okay, I just copy-pasted it from the previous example without knowing what it really does exactly. Thanks for the explanation.

timtebeek: You're welcome; thought as much seeing as the one just above had the same; figured point it out with a bit of background. :)

timtebeek (approved):

Issue #348, opened 2023-06-05
Simplify chained AssertJ assertions


TobiX:
## What problem are you trying to solve?

After running OpenRewrite to move assertions to AssertJ "best practices", SonarLint/SonarQube complains about many assertions not using the "best" form of AssertJ assertions (Rule: https://rules.sonarsource.com/java/RSPEC-5838). It would be nice if OpenRewrite could apply these suggestions automatically.

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?

Maybe, but currently not enough time.

timtebeek:
Hi @TobiX ; thank you for reporting these issues here & the helpful kink to RSPEC-5838. Looking through the findings we appear to already cover some of those rewrites through [recent Assert recipe additions](https://github.com/openrewrite/rewrite-testing-frameworks/tree/main/src/main/java/org/openrewrite/java/testing/assertj):
1. [UseExplicitContains.java](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitContains.java)
2. [UseExplicitIsEmpty.java](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitIsEmpty.java)
3. [UseExplicitSize.java](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/java/org/openrewrite/java/testing/assertj/UseExplicitSize.java)

But we can always to better with new recipes to handle the remaining cases. Our [upcoming support for refaster templates](https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/lang/UseStringIsEmpty.java#L22-L30) should make it a whole lot easier to contribute such rewrite recipes, and we'll gladly coach anyone to such a contribution.

Are there any examples in RSPEC-5838 in particular that you would like to prioritize?

TobiX:
Hmm, interesting. Some notes:
-  `org.openrewrite.java.testing.assertj.Assertj` doesn't contain `UseExplicitIsEmpty` yet? (This is probably why I missed some of those changes)
[...]

PR #385, opened 2023-07-20
Simplify chained AssertJ assertions

AlekSimpson:
## What's changed?
This PR adds a recipe that finds chained AssertJ assertions that have equivalent dedicated assertions and replaces them with those dedicated assertions. 

## What's your motivation?
[#348](https://github.com/openrewrite/rewrite-testing-frameworks/issues/348)

I think I have the idea of generalizing all these replacements right. There is still more I need to add to make it more robust, this is just the initial version. However, I can't get the first `ASSERT_TO_REPLACE` matcher to pass even though in the test it should be. I have included `assertj-core-3.24` in the test resources and made sure the method patterns match with the import statements. Maybe I am just using the wrong method pattern?

@timtebeek 

### Checklist


src/test/java/org/openrewrite/java/testing/assertj/SimplifyChainedAssertJAssertionsTest.java
```diff
@@ -0,0 +1,52 @@
...
+          java(
+            """
+              import org.junit.jupiter.api.Test;
+              
+              import static org.assertj.core.api.Assertions.assertThat;
+              
+              class MyTest {
+                  @Test
+                  void testMethod() {
+                      assertThat(getString().isEmpty()).isTrue();
```
timtebeek, line 33: `getString()` is undefined here; which will lead to mismatches.

src/main/java/org/openrewrite/java/testing/assertj/SimplifyChainedAssertJAssertions.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new SimplifyChainedAssertJAssertionsVisitor();
+    }
+
+    private class SimplifyChainedAssertJAssertionsVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private final MethodMatcher ASSERT_THAT_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions assertThat(..)");
+        private final MethodMatcher CHAINED_ASSERT_MATCHER = new MethodMatcher("java..* " + chainedAssertion + "(..)");
+        private final MethodMatcher ASSERT_TO_REPLACE = new MethodMatcher("org.assertj.core.api.*" + assertToReplace + "(..)");
```
timtebeek, line 83: I think we're missing a space here to separate the class from the method.
```suggestion
        private final MethodMatcher ASSERT_TO_REPLACE = new MethodMatcher("org.assertj.core.api.* " + assertToReplace + "(..)");
```

src/main/java/org/openrewrite/java/testing/assertj/SimplifyChainedAssertJAssertions.java
```diff
@@ -0,0 +1,139 @@
...
+
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            //  assert has assertion
+            if (!ASSERT_TO_REPLACE.matches(mi) || mi.getSelect() == null) {
```
timtebeek, line 88: What breaks if we remove the first part of this conditional? I'd expect nothing.
```suggestion
            if (mi.getSelect() == null) {
```

src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
```diff
@@ -0,0 +1,83 @@
...
+          .recipe(Environment.builder()
+            .scanRuntimeClasspath("org.openrewrite.java.testing.assertj")
+            .build()
+            .activateRecipes("org.openrewrite.java.testing.assertj.SimplifyChainedAssertJAssertions"));
+    }
+
+    private static Stream<Arguments> stringReplacements() {
+        return Stream.of(
+          Arguments.arguments("getString().isEmpty()", "isTrue()", "isEmpty"),
+          Arguments.arguments("getString()", "hasSize(0)", "isEmpty()"),
```
timtebeek, line 47: Are we sure we're not inadvertently also  matching `assertThat(getString().hasSize(123)).isTrue()` ? Might be good to also add some parameterized tests where we should _not_ be making changes.

AlekSimpson:
Ok so I finished adding the parameterized tests. One thing I think was strange though was that in order for the tests to run I had to include the recipe in the test functions with `spec -> spec.recipe(...)` in order for them to run. I think maybe I am not linking the tests to the yaml file correctly?

src/main/resources/META-INF/rewrite/assertj.yml
```diff
@@ -45,6 +45,224 @@ recipeList:
       fullyQualifiedTargetTypeName: "org.assertj.core.api.Assertions"
   - org.openrewrite.java.UseStaticImport:
       methodPattern: "org.assertj.core.api.Assertions *(..)"
+
```
timtebeek, line 48:
```suggestion

---
type: specs.openrewrite.org/v1beta/recipe
name: org.openrewrite.java.testing.assertj.SimplifyChainedAssertJAssertions
displayName: Simplify AssertJ chained assertions
description: Replace AssertJ chained assertions with dedicated assertions that function the same.
tags:
  - testing
  - assertj
recipeList:
```

src/main/resources/META-INF/rewrite/assertj.yml
```diff
@@ -45,6 +45,224 @@ recipeList:
...
+  # Simplify Chained AssertJ Assertions
+  # String Assertions
+  - org.openrewrite.java.SimplifyChainedAssertJAssertion:
```
timtebeek, line 51: This and all below likely need the correct package & classname.
```suggestion
  - org.openrewrite.java.testing.assertj.SimplifyChainedAssertJAssertion:
```

src/main/java/org/openrewrite/java/testing/assertj/SimplifyChainedAssertJAssertions.java
```diff
@@ -0,0 +1,145 @@
...
+import org.openrewrite.java.tree.J;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+@AllArgsConstructor
+@NoArgsConstructor
+public class SimplifyChainedAssertJAssertions extends Recipe {
```
timtebeek, line 39: Let's rename this one to singular, such that we can have a grouping one that's plural.
```suggestion
public class SimplifyChainedAssertJAssertion extends Recipe {
```

src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
```diff
@@ -0,0 +1,320 @@
...
+        String formattedAssertBefore = assertBefore.formatted(chainedAssertion, firstArg, assertToReplace, secondArg);
+
+        String finalArgument = firstArg.equals("") && !secondArg.equals("0") ? secondArg : firstArg;
+        finalArgument = finalArgument.contains(".") ? finalArgument.split("\\.")[0] : finalArgument;
+
+        String before = String.format(template, formattedAssertBefore);
+        String after = String.format(template, assertAfter.formatted(dedicatedAssertion, finalArgument));
+
+        rewriteRun(
+          spec -> spec.recipe(new SimplifyChainedAssertJAssertions(chainedAssertion, assertToReplace, dedicatedAssertion)),
```
timtebeek, line 100: Should not be necessary here, nor below, with the changes above.
```suggestion
```

src/test/java/org/openrewrite/java/testing/assertj/SimplifyChainedAssertJAssertionsTest.java
timtebeek, line 1: We might just be able to lose this file completely; I imagine everything's covered through the yaml recipe tests.

AlekSimpson: Ya that's probably true. I originally had that file just to test specific things but now everything should be covered by the parameterized tests.

src/main/resources/META-INF/rewrite/assertj.yml
```diff
@@ -45,6 +45,233 @@ recipeList:
...
+displayName: Simplify AssertJ chained assertions
+description: Replace AssertJ chained assertions with dedicated assertions that function the same.
+tags:
+  - testing
+  - assertj
+recipeList:
+  # Simplify Chained AssertJ Assertions
+  # String Assertions
+  - org.openrewrite.java.testing.assertj.SimplifyChainedAssertJAssertion:
+      chainedAssertion: isEmpty
```
timtebeek, line 61: I think it's likely we'll want to move to something like matching `isEmpty(String)` or `isEmpty(CharSequence)` hear to prevent those too eager matches on `isEmpty(Path)`

AlekSimpson: So you think all of the tests that use `isEmpty` should be moved up to the top of the String assertions? I want to make sure I am understanding this correctly.

AlekSimpson:
@timtebeek I almost feel like maybe I should add one more parameter to the recipe that specifies what type the recipe should run on. This way if it finds a line of code that matches the rest of the parameters, it can check the type against what it knows it should only be running on. This way every recipe will only run on the correct types.

src/main/resources/META-INF/rewrite/assertj.yml
timtebeek: I'd meant more something along the lines of what you arrived at [in this comment](https://github.com/openrewrite/rewrite-testing-frameworks/pull/385#issuecomment-1659360505), but then perhaps (if possible) as part of the same input, unless that leads to less than optimal string wrangling. So for instance for this line:
```suggestion
      chainedAssertion: isEmpty(String)
```

AlekSimpson:
Ok so I ended up adding a fourth parameter that is for specifying which type the recipe should run on. It seems to have fixed all of the issues where some recipes would run to early. I also made it so that strings would not be directly passed into `startsWithRaw()` and are first wrapped with `Path.of()`.

src/main/resources/META-INF/rewrite/assertj.yml
```diff
@@ -24,6 +24,8 @@ tags:
 recipeList:
   - org.openrewrite.java.testing.assertj.StaticImports
   - org.openrewrite.java.testing.assertj.JUnitToAssertj
+  - org.openrewrite.java.testing.assertj.SimplifyChainedAssertJAssertions
+  # TODO Reevaluate and potentially remove these three explicit recipes
   - org.openrewrite.java.testing.assertj.UseExplicitContains
   - org.openrewrite.java.testing.assertj.UseExplicitIsEmpty
   - org.openrewrite.java.testing.assertj.UseExplicitSize
```
timtebeek, line 31: @AlekSimpson I'd think we can replace these three explicit recipes classes and their tests with yaml configuration and additional parameterized tests. Could you explore that to reduce how much we maintain?
```suggestion
```

src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
```diff
@@ -0,0 +1,439 @@
...
+          Arguments.arguments("isLessThanOrEqualTo", "hasSizeLessThanOrEqualTo", "expression"),
+          Arguments.arguments("isLessThan", "hasSizeLessThan", "expression"),
+          Arguments.arguments("isGreaterThan", "hasSizeGreaterThan", "expression"),
+          Arguments.arguments("isGreaterThanOrEqualTo", "hasSizeGreaterThanOrEqualTo", "expression")
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("arrayReplacements")
+    @Disabled("array.length is not a method invocation, so that's not matched or handled just yet")
```
timtebeek, line 406: As indicated `array.length` is not a method invocation, so it fails quite early in the checks. Not sure if we'd want to cover this within the same recipe already; perhaps best to leave this out and potentially later pick that up separately.

timtebeek (commented):
I think we're close to a merge; I'd like to cover/replace those previously explicit collection cases if we can; it'd be a shame not to now that we have generic replacement options.

I'd also defer handling arrays, and potentially move the recipe & test into a separate new issue that we could pick up later, as I think it might complicate the current implementation which assumes the assertThat argument is a method invocation.

src/main/resources/META-INF/rewrite/assertj.yml
AlekSimpson: Ya I can, I think I remember looking at those recipes when starting this one. If I remember correctly they were essentially doing the same thing as this recipe but for specific methods?

AlekSimpson: Ok I looked into those recipes and ran their tests against this new recipe and they all passed. I have commented them out in the yaml, I think its safe to remove them.

timtebeek: Great! Then let's remove those classes and associated tests, and ensure those tests are covered in the parameterized tests we already have.

src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
timtebeek: Probably best to remove these tests and the commented out recipes as well then to really clean up the PR before a merge.

PR #386, opened 2023-07-21
Use xmlunit-legacy instead of xmlunit

puskarpeter:
- xmlunit is not compatible with JUnit5 hence we need to migrate it

## What's changed?

Package migration for xmlunit, no code changes since xmlunit-legacy provides compatibility between versions 2.x and 1.x

## What's your motivation?

JUnit5 compatibility. XMLUnit 1.6 would not compile, XMLUnit legacy everything works with JUnit5 as expected.

## Any additional context

https://github.com/xmlunit/xmlunit/#requirements

### Checklist


timtebeek (approved):
Thanks a lot @puskarpeter

src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -266,3 +267,20 @@ recipeList:
...
+  - testing
+  - junit
+  - xmlunit
+recipeList:
+  - org.openrewrite.maven.ChangeDependencyGroupIdAndArtifactId:
+      oldGroupId: xmlunit
+      oldArtifactId: xmlunit
+      newGroupId: org.xmlunit
+      newArtifactId: xmlunit-legacy
+      newVersion: 2.9.1
```
timtebeek, line 285: Looking at https://github.com/xmlunit/xmlunit/releases it seems there's infrequent releases.
Would you think it makes sense to adopt [version selectors](https://docs.openrewrite.org/reference/dependency-version-selectors) here?
We could use for instance the following to get any future 2.9.x or 2.10+ release added:
```suggestion
      newVersion: 2.x
```

PR #390, opened 2023-07-23
Hamcrest to AssertJ dependency tests

timtebeek:
## What's changed?
Add tests to verify hamcrest to assertj dependencies are added with correct scope.

## What's your motivation?
When running through the platform, I saw Gradle dependencies added where they might not have been needed. 

Potentially limited value having these here. Should we still add them?

Might need to troubleshoot elsewhere, but this is at least a sanity check that dependencies are added with the right scope and only if using.


timtebeek:
Merging now to see if pipeline still passes

PR #392, opened 2023-08-01
Collapse consecutive `assertThat` statements (#373)

### LINKED ISSUE #373: Collapse consecutive AssertJ `assertThat(..)` statements that each use the same object to a single statement
## What precondition(s) should be checked before applying this recipe?
Only consecutive calls; not when interleaved with others.
Only when the asserted object does not change; so for instance no calls to `extracting("baz")`, as that would change the semantics.

## Describe the situation before applying the recipe
```java
assertThat(listA).isNotNull();
assertThat(listA).hasSize(3);
assertThat(listA).containsExactly("a", "b", "c");
```

## Describe the situation after applying the recipe
```java
assertThat(listA)
    .isNotNull()
    .hasSize(3)
    .containsExactly("a", "b", "c");
```

## Any additional context
Would help [clean up after the migration from Hamcrest `allOf(Matcher...)`](https://github.com/openrewrite/rewrite-testing-frameworks/issues/357) to individual statements.

srmalkan:
- Closes #373 

Hi All, 

I have added a recipe to collapse if consecutive AssertThat statements on the same parameter are found. It ignores cases where the parameter is a Method Invocation or lambda or existing chained assertThat statement.

I am facing a bit of difficulty in formatting the chained assertThat statement, can someone guide me on that?


timtebeek:
Really nice to see you picked this up @srmalkan! Impressed with how much of the recipe already works. I've explored locally if I could get the indentation to work, but no such luck yet either. I'd expected the fix to be in setting a prefix on each of the method invocations where we change the select, but I've not yet found a working combination. Will try to explore later again.

srmalkan:
Yes, i thought so too. After looking at few objects, now i feel its related to setting up 'after' correctly in JRightPadded.

srmalkan:
Hi @timtebeek,

I was able to fix formatting by setting `after` in the JRightPadded element. However, facing difficulty to set whitespace correctly.

timtebeek:
Glad to see how far this has come! I think we can just adjust the `after` values to include and additional level of indentation; That's the only remaining issue right?

src/main/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatements.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+
+        private List<List<J.MethodInvocation>> getConsecutiveAssertThatList(List<Statement> statements) {
+            List<List<J.MethodInvocation>> consecutiveAssertThatList = new ArrayList<>();
+            String prevArg = "";
+            int currListIndex = 0;
+            int statementListSize = statements.size();
+            List<J.MethodInvocation> currList = new ArrayList<>();
+            for (int currIndex = 0; currIndex < statementListSize; currIndex++) {
```
joanvr, line 84:
```suggestion
            List<J.MethodInvocation> currList = new ArrayList<>();
            for (int currIndex = 0; currIndex < statements.size(); currIndex++) {
```

src/main/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatements.java
```diff
@@ -0,0 +1,169 @@
...
+
+                consecutiveAssertThatList.add(new ArrayList<>());
+
+                Statement statement = statements.get(currIndex);
+                if (statement instanceof J.MethodInvocation) {
+                    Optional<J.MethodInvocation> assertThatMi = getAssertThatMi(statement);
+                    if (assertThatMi.isPresent()) {
+                        if (isAssertThatValid(statement)) {
+                            if (!getFirstArgumentName(assertThatMi.get()).equals(prevArg)) {
```
joanvr, line 93: Does it make sense to use org.openrewrite.java.search.SemanticallyEqual here instead of just checking the argument name?

src/main/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatements.java
```diff
@@ -0,0 +1,170 @@
...
+                    chainCount.set(chainCount.get() + 1);
+
+                    if (ASSERT_THAT.matches(mi)) {
+                        if (chainCount.get() > 2) {
+                            isValid.set(false);
+                        }
+
+                        J assertThatArgument = mi.getArguments().get(0);
+                        if (assertThatArgument instanceof J.MethodInvocation || assertThatArgument instanceof J.Lambda)
+                            isValid.set(false);
```
timtebeek, line 139:
```suggestion
                        if (assertThatArgument instanceof J.MethodInvocation || assertThatArgument instanceof J.Lambda) {
                            isValid.set(false);
                        }
```

timtebeek (approved):
Thanks for getting this started and your patience throughout @srmalkan ! Good to see this completed, finally, and I look forward to seeing this put to use.

PR #393, opened 2023-08-01
Fix problem with incorrectly removed Mockito imports

### LINKED ISSUE #280: CleanupMockitoImports incorrectly removes `import org.mockito.Mockito;`
**Steps to reproduce**

1. `git clone git@github.com:apache/wicket.git`
2. `cd wicket`
3. Run [Cleanup Mockito imports](https://docs.openrewrite.org/reference/recipes/java/testing/mockito/cleanupmockitoimports)
```shell
mvn org.openrewrite.maven:rewrite-maven-plugin:4.36.0:run   \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:1.30.0   \
  -DactiveRecipes=org.openrewrite.java.testing.junit5.UseMockitoExtension
```

This produces changes in various files
```
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/markup/html/internal/HeaderResponseTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehaviorTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/core/util/file/WebApplicationPathTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/core/request/cycle/RequestCycleUrlForTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/core/request/resource/ByteArrayResourceTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/core/request/handler/render/WebPageRendererTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/resource/ResourceUtilTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/validation/CompoundValidatorTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/protocol/http/WicketFilterTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/protocol/http/mock/MockHttpServletRequestTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING] Changes have been made to wicket-core/src/test/java/org/apache/wicket/protocol/http/servlet/ServletWebResponseTest.java by:
[WARNING]     org.openrewrite.java.testing.mockito.CleanupMockitoImports
```

Yet all of those changes remove imports that are still being used
```diff
a/wicket-core/src/test/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehaviorTest.java b/wicket-core/src/test/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehaviorTest.java
index a658eb064a..4b96f37e4d 100644
--- a/wicket-core/src/test/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehaviorTest.java
+++ b/wicket-core/src/test/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehaviorTest.java
@@ -34,7 +34,6 @@ import org.apache.wicket.markup.html.form.Form;
 import org.apache.wicket.markup.html.link.Link;
 import org.apache.wicket.util.tester.WicketTestCase;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
 
 /**
  * Tests for AbstractDefaultAjaxBehavior
```

As can be [seen in the original](https://github.com/apache/wicket/blob/master/wicket-core/src/test/java/org/apache/wicket/ajax/AbstractDefaultAjaxBehaviorTest.java#L65)
```java
		listener.onComplete("alert('Complete!');");
		attributes.getAjaxCallListeners().add(listener);

		Component component = Mockito.mock(Component.class);
		AbstractDefaultAjaxBehavior behavior = new AbstractDefaultAjaxBehavior()
		{
```

Note that this also happens when executed as part of a larger suite of Mockito related changes, such as [Use Mockito JUnit Jupiter extension](https://docs.openrewrite.org/reference/recipes/java/testing/junit5/usemockitoextension).

wakingrufus:
## What's changed?
Don't attempt to remove Mockito class import when we know method calls use it

## What's your motivation?
Without this change, in classes that use both Class-qualified method calls and static method calls, 
the recipe will attempt to remove the Mockito class import.
RemoveImports is smart enough to not do this, but it does remove the static method import instead,
even if it is used. I do not fully understand why, so I don't feel comfortable proposing a fix to RemoveImports.
Instead, I accept it as a limitation of RemoveImports and have added logic to the recipe to not attempt to remove the class import if we know it is directly referenced.

This might not be the most elegant solution, so please let me know if there is a better way to handle this.
Also, let me know if there are any other test cases that should be covered

The recipe _could_ reorganize imports to use _only_ static import or class imports, and not mix.
That would be a more significant change, and would inject some opinions that might not be expected by users of the recipe.

## Any additional context
fixes #280 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/mockito/CleanupMockitoImportsTest.java
+++ b/src/test/java/org/openrewrite/java/testing/mockito/CleanupMockitoImportsTest.java
@@ -98,6 +98,60 @@ class CleanupMockitoImportsTest implements RewriteTest {
     }
 
     @Test
+    void doNotRemoveImportsAssociatedWithAnUntypedMockitoMethodMixed() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.mockito.Mockito;
+              import static org.mockito.Mockito.when;
+              import static org.mockito.BDDMockito.given;
+              import static org.mockito.Mockito.verifyNoInteractions;
+
+              class MyObjectTest {
+                MyObject myObject;
+                MyMockClass myMock;
+                            
+                void test() {
+                  when(myObject.getSomeField()).thenReturn("testValue");
+                  given(myObject.getSomeField()).willReturn("testValue");
+                  verifyNoInteractions(myMock);
+                  Mockito.reset();
+                }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void doNotRemoveImportsAssociatedWithATypedMockitoMethodMixed() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.mockito.Mockito;
+              import static org.mockito.Mockito.when;
+              import static org.mockito.BDDMockito.given;
+              import static org.mockito.Mockito.verifyNoInteractions;
+
+              class MyObjectTest {
+                MyObject myObject;
+                MyMockClass myMock;
+                            
+                void test() {
+                  when(myObject.getSomeField()).thenReturn("testValue");
+                  given(myObject.getSomeField()).willReturn("testValue");
+                  verifyNoInteractions(myMock);
+                  Mockito.mock(OtherClass.class);
+                }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
     void doNotRemoveStartImportsPossiblyAssociatedWithAnUntypedMockitoMethod() {
         //language=java
         rewriteRun(
--- a/src/main/java/org/openrewrite/java/testing/mockito/CleanupMockitoImports.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/CleanupMockitoImports.java
@@ -29,13 +29,11 @@ import org.openrewrite.java.tree.TypeUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.regex.Pattern;
 
 /**
  * Removes unused "org.mockito" imports.
  */
 public class CleanupMockitoImports extends Recipe {
-    private static final Pattern MOCKITO_CLASS_PATTERN = Pattern.compile("org.mockito.Mockito");
 
     @Override
     public String getDisplayName() {
@@ -52,7 +50,7 @@ public class CleanupMockitoImports extends Recipe {
         return Preconditions.check(new UsesType<>("org.mockito.*", false), new CleanupMockitoImportsVisitor());
     }
 
-    public static class CleanupMockitoImportsVisitor extends JavaIsoVisitor<ExecutionContext> {
+    private static class CleanupMockitoImportsVisitor extends JavaIsoVisitor<ExecutionContext> {
         private static final List<String> MOCKITO_METHOD_NAMES = Arrays.asList(
                 "after",
                 "atLeast",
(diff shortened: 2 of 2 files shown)


src/test/java/org/openrewrite/java/testing/mockito/CleanupMockitoImportsTest.java
```diff
@@ -97,6 +97,33 @@ void test() {
...
+              import static org.mockito.BDDMockito.given;
+              import static org.mockito.Mockito.verifyNoInteractions;
+
+              class MyObjectTest {
+                MyObject myObject;
+                MyMockClass myMock;
+                            
+                void test() {
+                  when(myObject.getSomeField()).thenReturn("testValue");
+                  given(myObject.getSomeField()).willReturn("testValue");
```
timtebeek, line 117: These get methods are not defined, leading to a missing types and hence the incorrect conclusion that there are unused imports. Does the test still fail if these get methods are present?

wakingrufus: I think this is a deeper problem in the RemoveImports class. in the evaluation of maybeRemove the non-static import, it sees the static when import, and removes that instead

wakingrufus: I tried repro in the unit test for RemoveImport and it didn't repro, so there is another factor im missing. Back to debugging...

timtebeek: Much appreciated that you're having a detailed look!

wakingrufus: ok my unit test in RemoveImport CAN repro the issue when i put an "unknown" parameter into the static method call. so your hunch about that was right. But i think maybe the fix is that the mockito recipe should not even be attempting to remove the class import in this case

wakingrufus: OK I think I have a solution. what do you think?

timtebeek: Looks good! Thanks for working through this one. I've added a very minor change to use TypeUtils, such that it's more inline with what we use elsewhere, but otherwise happy to see this resolved!

timtebeek:
Agree with your assessment that there might be an limitation to how RemoveImports currently works with a mixture of static and non static imports. Harder to confidently resolve without any negative side effects, which is also why we already maintain that list of `MOCKITO_METHOD_NAMES` here with special handling. Fine to expand the handling to include qualified import handling as you've done here, as it's most prevalent with Mockito.

As for reorganizing the imports: we have [such recipes for JUnit & AssertJ](https://docs.openrewrite.org/?q=static+import), so it's not unprecedented. But indeed, that might be slightly past what people prefer, and then handling this case properly is the better approach, at least for now.

Thanks again!

timtebeek (approved):

Issue #400, opened 2023-08-09
Adopt AssertJ duration assertions


timtebeek:
## What problem are you trying to solve?
More expressive assertions around durations, using the methods defined in:
https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/DurationAssert.html

## What precondition(s) should be checked before applying this recipe?
Durations are used in the class.

## Describe the situation before applying the recipe
```java
assertThat(Duration.between(accessToken.getIssuedAt(), accessToken.getExpiresAt()).getSeconds()).isEqualTo(1);
```

## Describe the situation after applying the recipe
```java
assertThat(Duration.between(accessToken.getIssuedAt(), accessToken.getExpiresAt())).hasSeconds(1);
```

## Any additional context
We might also want to simplify the durations of the resulting assertions, similar to what we already do in:
https://docs.openrewrite.org/recipes/staticanalysis/simplifydurationcreationunits
```diff
-.hasSeconds(300)
+.hasMinutes(5)
```

timtebeek:
Once merged we can run this against spring-security, as seen previously in  https://github.com/spring-projects/spring-security/pull/13619

timtebeek:
Seeing some errors when [run against Spring Security](https://app.moderne.io/results/y6G9lqKyV); would you mind investigating @AlekSimpson ?

timtebeek:
# Error:
```
assertThat(returned).as("The returned value is not correct.").isEqualTo(Tag.EVAL_PAGE)
```

# Message:
```
class org.openrewrite.java.tree.J$Identifier cannot be cast to class org.openrewrite.java.tree.J$MethodInvocation (org.openrewrite.java.tree.J$Identifier and org.openrewrite.java.tree.J$MethodInvocation are in unnamed module of loader 'app')
```

# Detail:
```
java.lang.ClassCastException: class org.openrewrite.java.tree.J$Identifier cannot be cast to class org.openrewrite.java.tree.J$MethodInvocation (org.openrewrite.java.tree.J$Identifier and org.openrewrite.java.tree.J$MethodInvocation are in unnamed module of loader 'app')
org.openrewrite.java.testing.assertj.AdoptAssertJDurationAssertions$AdoptAssertJDurationAssertionsVisitor.simplifyMultipleAssertions(AdoptAssertJDurationAssertions.java:132)
[...]

PR #401, opened 2023-08-24
Adopt AssertJ duration assertions

AlekSimpson:
## What's changed?
This PR introduces a recipe that simplifies some AssertJ duration assertions with dedicated asserts and also simplifies some `Duration`s to easier to read forms.
Ex:
```
- .hasHours(48)
+ .hasDays(2)
```

## What's your motivation?
[#400](https://github.com/openrewrite/rewrite-testing-frameworks/issues/400)

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
@@ -133,9 +133,7 @@ public class AdoptAssertJDurationAssertions extends Recipe {
                 return m;
             }
 
-            Long isEqualToArgRaw = SimplifyDurationCreationUnits.getConstantIntegralValue(isEqualToArg);
-            boolean isRelatedToDuration = checkIfRelatedToDuration(assertThatArg);
-            if (isEqualToArgRaw == 0 && isRelatedToDuration) {
+            if (isZero(isEqualToArg) && checkIfRelatedToDuration(assertThatArg)) {
                 String formatted_template = formatTemplate("assertThat(#{any()}).%s();", m.getSimpleName(), asDescription);
                 templateParameters.set(0, assertThatArg);
                 return applyTemplate(ctx, m, formatted_template, templateParameters.toArray());
@@ -152,6 +150,14 @@ public class AdoptAssertJDurationAssertions extends Recipe {
             }
 
             return m;
+        }
+
+        private boolean isZero(Expression isEqualToArg) {
+            if (isEqualToArg instanceof J.Literal) {
+                J.Literal literal = (J.Literal) isEqualToArg;
+                return literal.getValue() instanceof Number && ((Number) literal.getValue()).longValue() == 0;
+            }
+            return false;
         }
 
         private J.MethodInvocation simplifyTimeUnits(J.MethodInvocation m, ExecutionContext ctx) {
--- a/src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
@@ -64,7 +64,7 @@ public class AdoptAssertJDurationAssertions extends Recipe {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return Preconditions.check(Preconditions.or(
-                new UsesMethod<>("org.assertj.core.api.AbstractDurationAssert has*(int)"),
+                new UsesMethod<>("org.assertj.core.api.AbstractDurationAssert has*(int)", true),
                 new UsesMethod<>("org.assertj.core.api.AbstractLongAssert isEqualTo(..)")
             ), new AdoptAssertJDurationAssertionsVisitor()
         );
--- a/src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
@@ -15,9 +15,39 @@
  */
 package org.openrewrite.java.testing.assertj;
 
+import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 public class AdoptAssertJDurationAssertions extends Recipe {
+    static final MethodMatcher ASSERT_THAT_MATCHER = new MethodMatcher("org.assertj.core.api.Assertions assertThat(..)");
+    static final MethodMatcher GET_NANO_MATCHER = new MethodMatcher("java.time.Duration getNano()");
+    static final MethodMatcher GET_SECONDS_MATCHER = new MethodMatcher("java.time.Duration getSeconds()");
+    static final MethodMatcher ISEQUALTO_MATCHER = new MethodMatcher("org.assertj.core.api.AbstractLongAssert isEqualTo(..)");
+    static final Map<String, String> methodMap = new HashMap<String, String>() {{
+        put("getSeconds", "hasSeconds");
+        put("getNano", "hasNanos");
+    }};
+    static List<MethodMatcher> timeUnitMatchers = Arrays.asList(
+            new MethodMatcher("org.assertj.core.api.DurationAssert hasMillis(..)"),
+            new MethodMatcher("org.assertj.core.api.DurationAssert hasSeconds(..)"),
+            new MethodMatcher("org.assertj.core.api.DurationAssert hasNanos(..)"),
+            new MethodMatcher("org.assertj.core.api.DurationAssert hasMinutes(..)"),
+            new MethodMatcher("org.assertj.core.api.DurationAssert hasHours(..)"),
+            new MethodMatcher("org.assertj.core.api.DurationAssert hasDays(..)")
+    );
+
     @Override
     public String getDisplayName() {
         return "Adopt AssertJ Duration assertions";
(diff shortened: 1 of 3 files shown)


timtebeek:
We might also want to cover these methods: [isNegative](https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractDurationAssert.html#isNegative()), [isPositive](https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractDurationAssert.html#isPositive()), [isZero](https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractDurationAssert.html#isZero()). 
Other than that a good start, after fixing the compilation errors in 99c123c.

You can [reuse this (now) public method to simplify the expressions to a constant value](https://github.com/openrewrite/rewrite-static-analysis/blob/ba080f5e7fb3698cc196e6199fbed22822e720f2/src/main/java/org/openrewrite/staticanalysis/SimplifyDurationCreationUnits.java#L116):.

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -27,4 +57,53 @@ public String getDisplayName() {
     public String getDescription() {
         return "Adopt AssertJ DurationAssert assertions for more expressive messages.";
     }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new AdoptAssertJDurationAssertionsVisitor();
```
timtebeek, line 63: Can we add a precondtion here on using either `org.assertj.core.api.DurationAssert has*(..)` or `org.assertj.core.api.AbstractLongAssert isEqualTo(..)`? That could help prevent the `stream().anyMatch` from being evaluated for each method invocation.

timtebeek: Even a matcher on `org.assertj.core.api.Assertions assertThat(..)` could limit execution quite a bit already, but more specific is better I think

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,154 @@
...
+    );
+
+    @Override
+    public String getDisplayName() {
+        return "Adopt AssertJ Duration assertions";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Adopt AssertJ DurationAssert assertions for more expressive messages.";
```
knutwannheden, line 61:
```suggestion
        return "Adopt AssertJ `DurationAssert` assertions for more expressive messages.";
```

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(Preconditions.or(
+                new UsesMethod<>("org.assertj.core.api.AbstractDurationAssert has*(int)"),
```
knutwannheden, line 67:
```suggestion
                new UsesMethod<>("org.assertj.core.api.AbstractDurationAssert has*(int)", true),
```

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,154 @@
...
+                new UsesMethod<>("org.assertj.core.api.AbstractLongAssert isEqualTo(..)")
```
knutwannheden, line 68:
```suggestion
                new UsesMethod<>("org.assertj.core.api.AbstractLongAssert isEqualTo(..)", true)
```

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,154 @@
...
+            Long argValue = SimplifyDurationCreationUnits.getConstantIntegralValue(arg); // note: guess my machine hasn't updated to the new commit yet, method should be public
+            if (argValue % timeLength == 0) {
+                // convert divided value to OpenRewrite LST type
+                int rawValue = (int)(argValue / timeLength);
+                J.Literal newArg = rawValueToExpression(rawValue);
+                // get new method name
+
+                // update method invocation with new name and arg
+                Expression methodSelect = m.getSelect();
+                return JavaTemplate.builder("#{any()}.#{any()}(#{int})")
```
knutwannheden, line 132: Just as a hint:
```suggestion
                return JavaTemplate.builder("#{any()}.#{any()}(#{any(int)})")
```

AlekSimpson: ah whoops, I don't know why that slipped my mind. Thanks for catching it.

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,154 @@
...
+    static final MethodMatcher ASSERT_THAT_MATCHER = new MethodMatcher("org.assertj.core.api.Assertions assertThat(..)");
+    static final MethodMatcher GET_NANO_MATCHER = new MethodMatcher("java.time.Duration getNano()");
+    static final MethodMatcher GET_SECONDS_MATCHER = new MethodMatcher("java.time.Duration getSeconds()");
+    static final MethodMatcher ISEQUALTO_MATCHER = new MethodMatcher("org.assertj.core.api.AbstractLongAssert isEqualTo(..)");
+    static final Map<String, String> methodMap = new HashMap<String, String>() {{
+        put("getSeconds", "hasSeconds");
+        put("getNano", "hasNanos");
+    }};
+    static List<MethodMatcher> timeUnitMatchers = Arrays.asList(
+            new MethodMatcher("org.assertj.core.api.AbstractDurationAssert hasMillis(..)", true),
```
knutwannheden, line 46: Please note that the following five lines should probably be changed in the same way. I just added the suggestion to the first one.

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,168 @@
...
+            int methodArg = (int)unitInfo.get(1);
+
+            if (!(m.getSimpleName().equals(methodName))) {
+                // convert divided value to OpenRewrite LST type
+                J.Literal newArg = rawValueToExpression(methodArg);
+                // update method invocation with new name and arg
+                Expression methodSelect = m.getSelect();
+                return JavaTemplate.builder("#{any()}.#{}(#{any(int)})")
+                        .contextSensitive()
+                        .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "junit-jupiter-api-5.9", "assertj-core-3.24"))
```
timtebeek, line 131: When would we need ` "junit-jupiter-api-5.9"` here and above?

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @SuppressWarnings("DataFlowIssue")
+    private static class AdoptAssertJDurationAssertionsVisitor extends JavaIsoVisitor<ExecutionContext> {
+        @Override
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            J.MethodInvocation m = super.visitMethodInvocation(mi, ctx);
+            if (timeUnitMatchers.stream().anyMatch(matcher -> matcher.matches(mi))) {
+                return simplifyTimeUnits(mi, ctx);
+            }else if (ISEQUALTO_INT_MATCHER.matches(mi) || ISEQUALTO_LONG_MATCHER.matches(mi)) {
```
timtebeek, line 84: Formatter not yet applied here.

src/test/java/org/openrewrite/java/testing/assertj/AdoptAssertJDuractionAssertionsTest.java
```diff
@@ -0,0 +1,426 @@
...
+                  void testMethod(Duration time) {
+                      assertThat(time).hasMinutes(5);
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    //millis to hours
```
timtebeek, line 202: I don't think these comments add much; can we remove these?

AlekSimpson: oh yes, I just had them there to help me when I was writing them all out

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
AlekSimpson: Not quite sure I am clear on this, do you mean that `maybeAutoFormat` should be used here?

timtebeek: I mean you've not run IntelliJ IDEA `Cmd + Alt + L`.

AlekSimpson:
@timtebeek Made all the revisions here except for the format one that I am unclear on. I also wanted to ask if you had specific replacements in mind for isPositive, isNegative and isZero. I assume you mean replacements like:

```
-.isEqualTo(0)
+.isZero()
```

timtebeek:
> I also wanted to ask if you had specific replacements in mind for isPositive, isNegative and isZero. I assume you mean replacements like:
> 
> ```
> -.isEqualTo(0)
> +.isZero()
> ```

Indeed such replacements are what I had in mind, as we go from comparing numbers to the [isNegative](https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractDurationAssert.html#isNegative()), [isPositive](https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractDurationAssert.html#isPositive()) & [isZero](https://www.javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractDurationAssert.html#isZero()) methods available on `AbstractDurationAssert`

timtebeek:
H @AlekSimpson ; I've added a few commits just now as I've gone over the implementation; I've also added some tests that highlight existing issues. Could you fix those before we merge?

AlekSimpson:
@timtebeek I am currently updating the code to handle the `.as()` test case. I've made it so that if the `as()` method is found then it gets the argument and saves it to add to the template parameters and then the template that is passed into the `JavaTemplate.builder()` is updated to retain the `as()` method. I keep getting the error:

knutwannheden:
There is a `%s` in that error message. I assume something didn't get substituted before you compiled the template.

AlekSimpson:
> There is a `%s` in that error message. I assume something didn't get substituted before you compiled the template.

Oh yep, that's definitely the issue

AlekSimpson:
@timtebeek The latest commit fixes two of the three tests you added. The last one that is not fixed is the retain whitespace test case. Do I need to use something like `maybeAutoformat` to fix this test? I considered trying to somehow manually add the whitespace in but that didn't seem like a very flexible solution to me.

src/main/java/org/openrewrite/java/testing/assertj/AdoptAssertJDurationAssertions.java
```diff
@@ -0,0 +1,208 @@
...
+                return Arrays.asList(name, argValue);
+            }
+        }
+
+        private J.MethodInvocation applyTemplate(ExecutionContext ctx, J.MethodInvocation m, String template, Object... parameters) {
+            return JavaTemplate.builder(template)
+                    .contextSensitive()
+                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "assertj-core-3.24"))
+                    .build()
+                    .apply(getCursor(), m.getCoordinates().replace(), parameters);
```
knutwannheden, line 180: @AlekSimpson What you want to do is to retain the whitespace here. Assuming you know that the result has a non-null `select` you can achieve that by doing something like this:
```java
            J.MethodInvocation invocation = JavaTemplate.builder(template)
                    .contextSensitive()
                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "assertj-core-3.24"))
                    .build()
                    .apply(getCursor(), m.getCoordinates().replace(), parameters);
            return invocation.getPadding()
                    .withSelect(
                            invocation.getPadding().getSelect()
                                    .withAfter(m.getPadding().getSelect().getAfter())
                    );
```
But it would probably be best if you guard this with null checks for the selects on both elements.

AlekSimpson: This works! Thanks for pointing it out, I'll be sure to add those null checks in.

timtebeek (approved):

Bananeweizen:
I would like to suggest moving the isZero, isPositive, isNegative out of this recipe. Here they apply just for duration, right? But they are valid for any assertion of numeric type, and would have far greater impact then.

knutwannheden:
> I would like to suggest moving the isZero, isPositive, isNegative out of this recipe. Here they apply just for duration, right? But they are valid for any assertion of numeric type, and would have far greater impact then.

Sounds good! Do you want to provide a PR for this? 😄

PR #411, opened 2023-10-01
fix: update used FindImports constructor to not match inherited

SimonVerhoeven:
## What's changed?

updated the used constructor

## What's your motivation?

To help fix the build

n/a

### Checklist


timtebeek (approved):
Cheers!

PR #416, opened 2023-10-19
feat: add mockito dependency if needed

SimonVerhoeven:
## What's changed?

Added addDependency for Mockito

## What's your motivation?

In case a rewrite from JMockito to Mockito happens, the build does not break.

### Checklist


kunli2 (approved):
thanks for the update!

PR #419, opened 2023-10-28
JMockit to Mockito Recipe - Handle Argument Matchers and Void Methods

tinder-dthomson:
## What's changed?

- Added handling for [Argument Matchers](https://jmockit.github.io/tutorial/Mocking.html#argumentMatching).
- Added handling for Void methods in Expectations blocks.

## What's your motivation?

Continuation of #418.

This recipe is still incomplete, pending handling of `times`, `minTimes`, and `maxTimes` fields of `Expectations` blocks in a subsequent PR.

### Checklist
- [x] I've added unit tests to cover both positive and negative cases
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #420, opened 2023-10-31
JMockit to Mockito Recipe - Handle Typed Class Argument Matching and Collections

tinder-dthomson:
## What's your motivation?

Continuation of #419.

Handle JMockit typed argument matchers like `mockInstance.anyMethod((List<String>) any)`.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockito.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockito.java
@@ -88,65 +88,70 @@ public class JMockitExpectationsToMockito extends Recipe {
             J.Block newBody = md.getBody();
             List<Statement> statements = md.getBody().getStatements();
 
-            // iterate over each statement in the method body, find Expectations blocks and rewrite them
-            for (int bodyStatementIndex = 0; bodyStatementIndex < statements.size(); bodyStatementIndex++) {
-                Statement s = statements.get(bodyStatementIndex);
-                if (!(s instanceof J.NewClass)) {
-                    continue;
-                }
-                J.NewClass nc = (J.NewClass) s;
-                if (!(nc.getClazz() instanceof J.Identifier)) {
-                    continue;
-                }
-                J.Identifier clazz = (J.Identifier) nc.getClazz();
-                if (!TypeUtils.isAssignableTo("mockit.Expectations", clazz.getType())) {
-                    continue;
-                }
-                // empty Expectations block is considered invalid
-                assert nc.getBody() != null && !nc.getBody().getStatements().isEmpty() : "Expectations block is empty";
-                // Expectations block should be composed of a block within another block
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockito.java
```diff
@@ -213,5 +267,17 @@ private static String getMockitoStatementTemplate(Expression result) {
             }
             return template;
         }
+
+        private static class ArgumentMatcherTemplateInfo {
```
timtebeek, line 271: Feel free to use `@Value` on this class; eventually we'll convert those to records directly.

timtebeek (commented):
Much appreciate the continued work on this! Let me know  when this is ready to review & merge.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockito.java
tinder-dthomson: Sounds good! I read that we should only assume Java 17 for tests and Java 8 for recipes so I did not use Records here.

tinder-dthomson:
> Much appreciate the continued work on this! Let me know when this is ready to review & merge.

Will do, I want to polish this a bit more - the code is getting tougher to read.

tinder-dthomson:
@timtebeek ready for review & merge

src/main/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockito.java
```diff
@@ -166,29 +173,80 @@ private J.Block applyTemplate(ExecutionContext ctx, List<Object> templateParams,
...
+                JavaType typeCastType = tc.getType();
+                if (typeCastType instanceof JavaType.Parameterized) {
+                    // strip the raw type from the parameterized type
+                    className = ((JavaType.Parameterized) typeCastType).getType().getClassName();
+                    fqn = ((JavaType.Parameterized) typeCastType).getType().getFullyQualifiedName();
+                } else if (typeCastType instanceof JavaType.FullyQualified) {
+                    className = ((JavaType.FullyQualified) typeCastType).getClassName();
+                    fqn = ((JavaType.FullyQualified) typeCastType).getFullyQualifiedName();
+                } else {
+                    throw new IllegalStateException("Unexpected value: " + typeCastType);
```
timtebeek, line 205: We prefer not to throw exception, and instead not to make a change; would that be possible here?
That way recipes are safe to run as part of a larger set, only making changes when safe to do so.

tinder-dthomson: @timtebeek 

I added a try/catch around the entire recipe to that if anything goes wrong we simply return the original LST.

Issue #360, opened 2023-06-19
PowerMockitoMockStaticToMockito removes mockStatic initialization


Laurens-W:
## What version of OpenRewrite are you using?

I am using

- Maven plugin v5.0.0
- rewrite-testing-frameworks v2.0.1

## How are you running OpenRewrite?

I am using the Maven plugin, from the command line:
mvn org.openrewrite.maven:rewrite-maven-plugin:5.0.0:run -Drewrite.activeRecipes=org.openrewrite.java.testing.mockito.PowerMockitoMockStaticToMockito -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-java-security:LATEST,org.openrewrite.recipe:rewrite-kubernetes:LATEST,org.openrewrite.recipe:rewrite-logging-frameworks:LATEST,org.openrewrite.recipe:rewrite-migrate-java:LATEST,org.openrewrite.recipe:rewrite-spring:LATEST,org.openrewrite.recipe:rewrite-testing-frameworks:LATEST -Drewrite.exclusions=\*\*api\*\*.yaml

## What is the smallest, simplest way to reproduce the problem?

```java
class A {
  @Test
  void testTryWithResource() throws IOException {
    try (MockedStatic<Files> mocked = mockStatic(Files.class)) {
      // test logic that uses mocked
    }
  }
}
```

## What did you expect to see?

```java
class A {
  @Test
  void testTryWithResource() throws IOException {
    try (MockedStatic<Files> mocked = mockStatic(Files.class)) {
      // test logic that uses mocked
    }
  }
}
```

## What did you see instead?

```java
class A {
  @Test
  void testTryWithResource() throws IOException {
    try (MockedStatic<Files> mocked) {
      // test logic that uses mocked, which now provides a warning: `Variable 'mocked' might not have been initialized`
    }
  }
}
```
Note that I didn't include the import in the examples, it remains intact

## What is the full stack trace of any errors you encountered?

None while the recipe ran

SanderKnauff:
I'm also encountering this.  It seems that this rewrite does not check if the used `mockStatic` import originates from PowerMock instead of Mockito Core

timtebeek:
[...]

PR #426, opened 2023-11-20
Don't remove `mockStatic()` calls in resources of try-with-resources

knutwannheden:
This commit does not yet close the issue, as there are other cases, where this can potentially go wrong.

It also appears strange that the recipe matches on Mockito's `mockStatic()` rather than on PowerMock's `mockStatic()`. Therefore, the recipe is for now excluded from `Mockito1to3Migration`, as this is part of `JUnit4to5Migration` which is widely used.

Issue: #360

### DIFF
--- a/src/main/resources/META-INF/rewrite/mockito.yml
+++ b/src/main/resources/META-INF/rewrite/mockito.yml
@@ -111,7 +111,7 @@ recipeList:
   - org.openrewrite.java.testing.mockito.CleanupMockitoImports
   - org.openrewrite.java.testing.mockito.MockUtilsToStatic
   - org.openrewrite.java.testing.junit5.MockitoJUnitToMockitoExtension
-  - org.openrewrite.java.testing.mockito.ReplacePowerMockito
+#  - org.openrewrite.java.testing.mockito.ReplacePowerMockito
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: org.mockito
       artifactId: mockito-junit-jupiter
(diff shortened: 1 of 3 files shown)


timtebeek (approved):

src/main/resources/META-INF/rewrite/mockito.yml
```diff
@@ -111,7 +111,7 @@ recipeList:
   - org.openrewrite.java.testing.mockito.CleanupMockitoImports
   - org.openrewrite.java.testing.mockito.MockUtilsToStatic
   - org.openrewrite.java.testing.junit5.MockitoJUnitToMockitoExtension
-  - org.openrewrite.java.testing.mockito.ReplacePowerMockito
+#  - org.openrewrite.java.testing.mockito.ReplacePowerMockito
```
timtebeek, line 114:
```suggestion
# https://github.com/openrewrite/rewrite-testing-frameworks/issues/360
#  - org.openrewrite.java.testing.mockito.ReplacePowerMockito
```

knutwannheden:
@timtebeek We should take a fresh look at this PowerMock recipe. The thing with the `mockStatic()` matcher looks really odd to me. I also don't understand why the recipe uses a precondition like https://github.com/openrewrite/rewrite-testing-frameworks/blob/de6b6a823b672b197195877c4c96d5196b81a421/src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoMockStaticToMockito.java#L45-L48

rather than only checking for PowerMock.

timtebeek:
> @timtebeek We should take a fresh look at this PowerMock recipe. The thing with the `mockStatic()` matcher looks really odd to me. I also don't understand why the recipe uses a precondition like
> 
> rather than only checking for PowerMock.

Agree that's odd; would only migrate when there's a confusing mix of libraries in use in a source file already? We can loosen that restriction as part of this PR, especially now that it's not part of the default set of recipes run anymore in larger composed migrations.

PR #430, opened 2023-11-22
Fixes broken testcontainer dependencies after JUnit4to5Migration

simonzn:
- Added a testcase for build errors due to missing Junit4 dependencies
- Added workaround to JUnit4to5Migration: Undo the exclusion

Fixes #429

PR #433, opened 2023-11-25
Rewrite testcontainers to use explicit image and version

timtebeek:
## What's changed?
Add recipes to rewrite testcontainers, as rewritten for Rewrite 8 from https://github.com/openrewrite/rewrite-testcontainers

## What's your motivation?
Fixes https://github.com/openrewrite/rewrite-testcontainers/issues/1

I've crudely stripped out individual recipes and tests where I saw duplication; rather we can just run an aggregate `ExplicitContainerImages` that will match whatever is present. I'll add the other recipes that were present still.

@eddumelendez @kiview

Integrated here such that these recipes can be used from a JUnit Jupiter migration, and themselves reuse other recipes from there.

## Any additional context
- https://github.com/openrewrite/rewrite-testcontainers/issues/1

### DIFF
--- a/src/main/resources/META-INF/rewrite/testcontainers.yml
+++ b/src/main/resources/META-INF/rewrite/testcontainers.yml
@@ -1,0 +1,96 @@
+#
+# Copyright 2023 the original author or authors.
+# <p>
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# <p>
+# https://www.apache.org/licenses/LICENSE-2.0
+# <p>
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+type: specs.openrewrite.org/v1beta/recipe
+name: org.testcontainers.openrewrite.ExplicitContainerImages
+displayName: Explicit container images and versions
+description: Replace implicit default container images and versions with explicit versions.
+recipeList:
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.containers.CassandraContainer
+      image: "cassandra:3.11.2"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.containers.ClickHouseContainer
+      image: "yandex/clickhouse-server:18.10.3"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.containers.CockroachContainer
+      image: "cockroachdb/cockroach:v19.2.11"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.couchbase.CouchbaseContainer
+      image: "couchbase/server:6.5.1"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.containers.Db2Container
+      image: "ibmcom/db2:11.5.0.0a"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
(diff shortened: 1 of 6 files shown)


timtebeek:
@eddumelendez & @kiview I'm strongly leaning towards only having the testcontainers recipes here and archiving the rewrite-testcontainers repository. I can grant you both contributor rights here all the same. What are your immediate thoughts to that?

I'd been wanting to revisit this for a while, but didn't quite get to it. Then I saw these two items this past week which spurred me into action now, to hopefully get migrations for testcontainers going.
- https://github.com/testcontainers/testcontainers-java/pull/5826
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/430

We can optionally (but ideally separately) pick up some more of the items in openrewrite/rewrite-testcontainers#1, such as 
- [ ]  Add recipe to add `@TestContainers(parallel = true)`
- [ ]  Add recipes to migrate to new JUnit-4-free API
- [ ] Recent [deprecations from the changelog](https://github.com/testcontainers/testcontainers-java/blob/main/CHANGELOG.md)

src/main/resources/META-INF/rewrite/testcontainers.yml
```diff
@@ -0,0 +1,118 @@
...
+      image: "rabbitmq:3.7.25-management-alpine"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.containers.SolrContainer
+      image: "solr:8.3.0"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.containers.ToxiproxyContainer
+      image: "shopify/toxiproxy:2.1.0"
+  - org.openrewrite.java.testing.testcontainers.ExplicitContainerImage:
+      containerClass: org.testcontainers.vault.VaultContainer
+      image: "vault:1.1.3"
```
timtebeek, line 118: These versions seem to line up well with what I found through `grep -r 'DEFAULT_TAG = ' modules/ | cut --delimiter=" " --fields=1,11 | sort` on testcontainers/testcontainers-java:
https://github.com/search?q=repo%3Atestcontainers%2Ftestcontainers-java%20%22DEFAULT_TAG%20%3D%20%22&type=code

There's been some new addition though such as PostGIS, but I haven't checked each of them to see if those new additions contain a default no arg constructor. As-is this is already valuable I suppose. New ones are also easily added after a merge.

pstreef (approved):

src/main/resources/META-INF/rewrite/testcontainers.yml
```diff
@@ -0,0 +1,118 @@
...
+recipeList:
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: org.testcontainers.containers.ContainerState getContainerIpAddress()
+      newMethodName: getHost
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.testing.testcontainers.ExplicitContainerImages
+displayName: Explicit container images and versions
+description: Replace implicit default container images and versions with explicit versions.
+recipeList:
```
eddumelendez, line 43: kafka, localstack, mockserver and pulsar should be excluded due to the `constructor(String)` uses the tag instead of the image. For those cases, it should use `DockerImageName.parse(<image>:<tag>)`

src/main/resources/META-INF/rewrite/testcontainers.yml
eddumelendez: we are making sure new ones don't provide a default constructor. A recommended practice is to declare the image and tag to use.

All versions are good 👍🏽 I compared to the [initial implementation](https://github.com/openrewrite/rewrite-testcontainers/blob/main/src/main/resources/META-INF/rewrite/testcontainers.yml)

src/main/resources/META-INF/rewrite/testcontainers.yml
timtebeek: Good to know! I've excluded these for now; likely need an explicit recipe that adds that `DockerImageName.parse`, which we can add separately after this first merge; should be relatively straightforward now there's no build issues.

timtebeek: Added an option and restored those four in https://github.com/openrewrite/rewrite-testing-frameworks/pull/433/commits/0ac879ac093c545c8677e74e7499fb033bd1312b

PR #442, opened 2023-12-18
Do not change visibility in `UpdateTestAnnotation`; use `TestsShouldNotBePublic` instead

### LINKED ISSUE #441: Make visibility modifier change by UpdateTestAnnotation optional
The recipe UpdateTestAnnotation doing the migration of the @Test annotation from JUnit 4 to 5 also removes the public modifier from the test method.
This seems arbitrary to me as this may not be desired.
It is true that JUnit 5 does not require test lifecycle methods to be public, but this is not a requirement.

With the current implementation, these changes must be reverted manually if the test methods should remain public
(or by writing an additional rule which does the opposite of TestsShouldNotBePublic).

I would propose to make the removal of the visibility modifier configurable by adding a respective @Option to the recipe.

rewrite-testing-frameworks-2.1.4

I could try to provide a PR for this.

magicwerk:
## What's changed?
Fixes #441

UpdateTestAnnotation does not change the visibility modifier anymore.
I updated the test as needed.
I removed the following 2 tests in UpdateTestAnnotationTest as they make no sense after the change:
	privateToPackageVisibility()
	protectedToPackageVisibility()

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
@@ -116,10 +116,6 @@ public class UpdateTestAnnotation extends Recipe {
             ChangeTestAnnotation cta = new ChangeTestAnnotation();
             J.MethodDeclaration m = (J.MethodDeclaration) cta.visitNonNull(method, ctx, getCursor().getParentOrThrow());
             if (m != method) {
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Thanks a lot for taking this on @magicwerk ! You can try it out from [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions) shortly; we'll follow up with a release in the next week or so.

Issue #443, opened 2023-12-18
Inherited lifecycle annotations are not migrated properly


magicwerk:
## What problem are you trying to solve?

JUnit 5 changed the way inherited lifecycle annotations are handled.
Lifecycle annotations are inherited from superclasses as long as they are not overridden,
i.e. if you override a lifecycle method, the annotation must be repeated.
This is different from JUnit 4 where the annotation had not to be repeated in the overriding class.

Reference:
https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/BeforeEach.html
@BeforeEach methods are inherited from superclasses as long as they are not overridden

The current implementation makes JUnit 4 such tests fail after running the JUnit4to5Migration recipe as the lifecycle methods are not executed anymore.

I propose to add a recipe AddMissingTestBeforeAfterAnnotations which will add
the missing annotations
@Test, @BeforeEach, @AfterEach
if they are found on a superclass and not present on the class analyzed.

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?
Yes, I would try to contribute a PR

timtebeek:
Perfect suggestion indeed @magicwerk . From #442 I think you've spotted the `TypeUtils.isOverride` that was used there, that should help inform where any such annotations might be missing. Feel free to immediately tie your recipe into the `JUnit4to5Migration`, as I expect this to be a required step there. And thanks a lot for the offer to help; look forward to your PR!

PR #444, opened 2023-12-19
`AddMissingTestBeforeAfterAnnotations` recipe for overridden JUnit test methods

magicwerk:
I created AddMissingTestBeforeAfterAnnotations with a test AddMissingTestBeforeAfterAnnotationsTest.

The recipe does not handle the JUnit4 annotations BeforeClass/AfterClass as they cannot only be used on static methods which cannot be overridden. I update the text in the issue accordingly
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/443

I added org.openrewrite.java.testing.junit5.AddMissingTestBeforeAfterAnnotations to the JUnit4to5Migration recipe. I therefore also created a new test migrateInheritedTestBeforeAfterAnnotations() in JUnit5MigrationTest.

Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/443

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/AddMissingTestBeforeAfterAnnotations.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AddMissingTestBeforeAfterAnnotations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2023 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,11 +15,10 @@
  */
 package org.openrewrite.java.testing.junit5;
 
-import java.util.Comparator;
-import java.util.Optional;
-import java.util.function.Predicate;
-
+import lombok.EqualsAndHashCode;
+import lombok.Value;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.AnnotationMatcher;
@@ -30,14 +29,15 @@ import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType.FullyQualified;
 import org.openrewrite.java.tree.JavaType.Method;
 import org.openrewrite.java.tree.TypeUtils;
+import org.openrewrite.marker.SearchResult;
 
-import lombok.EqualsAndHashCode;
-import lombok.Value;
+import java.util.Comparator;
+import java.util.Optional;
+import java.util.function.Predicate;
 
 @Value
 @EqualsAndHashCode(callSuper = false)
 public class AddMissingTestBeforeAfterAnnotations extends Recipe {
-
     @Override
     public String getDisplayName() {
         return "Add missing `@BeforeEach`, `@AfterEach`, `@Test` to overriding methods";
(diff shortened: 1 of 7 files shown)


src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -79,6 +79,7 @@ recipeList:
   - org.openrewrite.java.testing.junit5.UpdateBeforeAfterAnnotations
   - org.openrewrite.java.testing.junit5.LifecycleNonPrivate
   - org.openrewrite.java.testing.junit5.UpdateTestAnnotation
+  - org.openrewrite.java.testing.junit5.AddMissingTestBeforeAfterAnnotations
```
timtebeek, line 82: I worry that we might be sensitive to ordering issues here, where if the super class has it's annotations changed before the extending class, would we then fail to match given that our predicate only checks for the old annotations?

src/main/java/org/openrewrite/java/testing/junit5/AddMissingTestBeforeAfterAnnotations.java
```diff
@@ -0,0 +1,117 @@
...
+
+        String annotation;
+        String simpleAnnotation;
+        private Predicate<FullyQualified> oldAnnotationPredicate;
+        private AnnotationMatcher annotationMatcher;
+
+        LifecyleAnnotation(String annotation, String oldAnnotation) {
+            this.annotation = annotation;
+            this.simpleAnnotation = "@" + annotation.substring(annotation.lastIndexOf(".") + 1);
+            this.oldAnnotationPredicate = n -> TypeUtils.isOfClassType(n, oldAnnotation);
```
timtebeek, line 105: I wonder if we should expand this predicate to also check for super class methods annotated with the new annotations. One reason is because of the ordering issue pointed out on `junit5.yml`, and another reason is because folks might have migrated by hand, and missed adding those annotations. Those overridden tests would then today not run, whereas they very likely should be.

I don't suppose there's many folks that In JUnit Jupiter intentionally override a method, and do not add the test annotation to stop running a test, although it's certainly possible. Such cases would be negatively affected there, so something to keep in mind. Ideally those folks override the super method and add `@Disabled` instead of relying on the absence of a test annotation.

timtebeek (commented):
Solid work, thanks! I've pushed some small changes mostly to limit where this recipe is run, as Preconditions help keep our recipes performant.

I've got a couple questions called out below that I think we should be able to work through quickly. Let me know your thoughts and we'll see this through.

src/main/java/org/openrewrite/java/testing/junit5/AddMissingTestBeforeAfterAnnotations.java
magicwerk: To be honest, I was first checking for the new annotations because I thought that UpdateBeforeAfterAnnotations runs first and then we just have to handle the new annotations - however this seems not to be the case, I still see the unchanged annotations.
I then looked at the other recipes in the package and had the impression, that they all look for the old stuff and migrate it to the new, so I did it also this way.
So I think all other recipes will just work if the old JUnit stuff is still in place and AddMissingTestBeforeAfterAnnotations behaves the same way.

src/main/resources/META-INF/rewrite/junit5.yml
magicwerk: see below

magicwerk:
I think I answered your questions, but somehow I cannot see your questions or my answers any more - did you get them?

timtebeek:
> I think I answered your questions, but somehow I cannot see your questions or my answers any more - did you get them?

They're part of the resolved conversations above, which can be unfolded if needed; thanks for sharing what you had tried and found already. I'll trust that this works as expected then, and thanks for the addition here!

timtebeek (approved):

magicwerk:
I think I have to revise my statement regarding whether a check for the new annotations on the superclass should be done or not.
As said, I was originally also checking for the new annotations, but then thought it would be bad practice, but I think it is necessary so users get the expected behavior.

Currently I am now working on the migration of the large project and do this component per component. Now I run into the problem that my base classes are already migrated (so they have JUnit 5 annotations), so if I run the migration on another component the AddMissingTestBeforeAfterAnnotations is not applied. 

Regarding your comment about intentionally overriding a method, I still think we're on the safe side:
If you have written your tests already in JUnit 5 and rely on the fact that you override a test method without annotating it as @Test or @Disabled, then you are already on JUnit 5, so there is no need to run the JUnit4to5Migration recipe.

On the other hand if you try to do the migration step by step, looking for the new annotations will support this use case.
So yes, I would add the check for new annotations as discussed.

I would create a new PR to change to the behavior if you agree.

timtebeek:
Yes I agree with a PR that also checks for the new annotations, and thanks for taking that on!

The reason I saw some unlikely but potential issues with folks adding implicit overrides without test annotations on Junit Jupiter is that we have a suite of best practices recipes that we want them to be able to run repeatedly: https://docs.openrewrite.org/recipes/java/testing/junit5/junit5bestpractices

In the unlikely case that folks
_are_ already on Junit Jupiter 
_and_ override methods
_and_ intentionally not add `@Test` or `@Disabled` 
_and_ run [JUnit Jupiter best practices](https://docs.openrewrite.org/recipes/java/testing/junit5/junit5bestpractices)
_then_ and only then might we add those annotations until they explicitly add `@Disabled`.

Seems like a tolerable risk to me, but something to document perhaps on the next PR just in case, just above the line that looks for the new annotations too.

magicwerk:
Ok, I will create the PR.

Regarding the suite of recipes you mentioned:

Currently JUnit4to5Migration contains
org.openrewrite.java.testing.junit5.LifecycleNonPrivate
which makes no sense for me, as lifecycle annotations must be public in JUnit 4.

Should this probably be moved to 
JUnit5BestPractices
or be deleted altogether?

PR #445, opened 2023-12-21
AddMissingTestBeforeAfterAnnotations is looking for both old and new annotations

magicwerk:
As discussed:

Recipe AddMissingTestBeforeAfterAnnotations is looking for both old and new annotations to support incremental migration

Moved recipe org.openrewrite.java.testing.junit5.LifecycleNonPrivate from
JUnit4to5Migration to JUnit5BestPractices


timtebeek (approved):
Thanks again! Added a very small change for readability; otherwise perfect.

PR #449, opened 2024-01-04
AddMissingTestBeforeAfterAnnotations is looking for annotations in all ancestor classes

magicwerk:
This PR contains another enhancement to the AddMissingTestBeforeAfterAnnotations recipe.

see #443, #444, #445

The current implementation did not handle the case correctly where the test annotations where not on the superclass itself,
but on an ancestor class up in the hierarchy. With this fix, not only the superclass but all classes up in the hierarchy are checked.
See new test addMissingTestBeforeAfterAnnotationsIfExtended


timtebeek (approved):
Nice addition, thanks!

PR #455, opened 2024-01-12
JMockit to Mockito Recipe - Improved Expectations Handling

tinder-dthomson:
## What's changed?

`JMockitExpectationsToMockito` was largely a work-in-progress until this PR. This PR includes several functional improvements and bug fixes.

**New Functionality Supported:**
- Verifications of invocations using JMockit `times`.
- Expectations with different results using `returns`.
- `Injectable` and `Tested` annotations (see note on partial mocking).

**Bug Fixes:**
- Statements which do not stub any mock are now moved properly rather than removed.
- Argument matchers are now handled correctly. Mockito requires every method argument in a stub to use an argument matcher, or none at all.

Finally, the recipe underwent a major refactor to improve readability and maintainability.

## What's your motivation?

The current recipe is a WIP. This PR handles the majority of remaining common use cases (with the exception of spying/partial mocking).

### Not Yet Supported
The below features are not supported by the recipe yet, and will likely require manual intervention to achieve desired results.

**Partial Mocking** - Partial mocking using `Expectations` block initialization parameters (Spy in Mockito) is not yet handled
**Static Methods** - Static methods are treated like regular test statements (not JMockit-related) and are simply moved, not rewritten.

### Checklist


timtebeek (approved):
Thanks a lot! A thorough rewrite indeed which is much appreciated.

We don't often use exceptions and try/catch in recipes, but given the complexity of the recipe changes and how it's a one time migration I think it's valid to keep that in as is. 

Appreciate you taking the time to come back to this one to see it through; I hope that helps you there, and really neat to see you're contributing your recipe here.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
```diff
@@ -23,8 +23,7 @@
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
-class JMockitToMockitoTest implements RewriteTest {
-
+class JMockitExpectationsToMockitoTest implements RewriteTest {
```
tinder-dthomson, line 26: @timtebeek I named it `JMockitToMockitoTest` because we're actually running the `org.openrewrite.java.testing.jmockit.JMockitToMockito` recipe.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
```diff
@@ -43,7 +42,7 @@ public void defaults(RecipeSpec spec) {
     }
 
     @Test
-    void jMockitExpectationsToMockitoVoidResult() {
+    void voidResult() {
```
tinder-dthomson, line 45: 👍

tinder-dthomson:
> Thanks a lot! A thorough rewrite indeed which is much appreciated.
> 
> We don't often use exceptions and try/catch in recipes, but given the complexity of the recipe changes and how it's a one time migration I think it's valid to keep that in as is.
> 
> Appreciate you taking the time to come back to this one to see it through; I hope that helps you there, and really neat to see you're contributing your recipe here.

Much appreciated! This recipe is already coming in handy over here, but there's more to do! Many JMockit features are not handled (like `MockUp` and `Verifications`) so I will likely be adding more and more over time.

On the note of error handling, I am happy to change the recipe to be following best practices. In the case of missing type information, for example, what is the common way of handling that? Simply return the LST unchanged?

timtebeek:
Indeed returning the unchanged element is best if changes can't be made reliably. I had a briefly look at doing that on your PR, but given that we're pretty far down when that's encountered I figured keep it as is for now.

Once we have more complete coverage we can also look at removing the dependency if we haven't already. Thanks again!

tinder-dthomson:
@timtebeek I started a new PR to track this change (still in Draft). Let me know if this is moving in the right direction: https://github.com/openrewrite/rewrite-testing-frameworks/pull/461

PR #461, opened 2024-01-16
JMockit to Mockito Recipe - Bug Fixes and Improvements

tinder-dthomson:
## What's changed?

**Improvements:**

- Added support for `minTimes` and `maxTimes` JMockit features.
- Several pieces of logic depended on the instance of the LST element (e.g. `J.Literal` or `J.Identifier`) rather than the `JavaType` itself. Refactored to use the more appropriate type matching.
- Argument matcher convenience methods like `anyString` or `anyLong` were not uniformly applied and in some cases `any(java.lang.String)` would be used instead. Use the convenience APIs consistently.
- The recipe had several scenarios where an exception is intentionally thrown in order to prevent processing LSTs that contain either invalid or unhandled LST structures. Change this to instead simply return the LST elements unchanged and continue processing rather than short-circuiting the entire recipe.

**Bug Fixes:**

- Multiple Expectations Blocks in a single test method causing incorrect statement ordering.
- Fix handling of `this.result = <value>` in an `Expectations` block; previously only `result = <value>` was properly handled.

## What's your motivation?

Continue the recipe maturation!

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/jmockit/ArgumentMatchersRewriter.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/ArgumentMatchersRewriter.java
@@ -103,25 +103,27 @@ class ArgumentMatchersRewriter {
     }
 
     private Expression rewriteMethodArgument(Expression methodArgument) {
-        String argumentMatcher, template;
-        if (!isJmockitArgumentMatcher(methodArgument)) {
-            if (methodArgument instanceof J.Literal) {
-                return rewritePrimitiveToArgumentMatcher((J.Literal) methodArgument);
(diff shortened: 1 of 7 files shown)


timtebeek:
Thanks! That helps indeed; once you feel this sufficiently covers known and unknown scenarios we can consider removing the try/catch as well.

tinder-dthomson:
@timtebeek this is now ready for review. Let me know if you have any questions or concerns!

timtebeek (approved):
All goods very good to me; thanks for your continued involvement here! Hope being able to try it out on projects as you developed helped. :)

PR #472, opened 2024-01-25
Recipe to simplify assertions on Optionals

timo-abele:
## What's changed?

Adds first refaster recipe

## What's your motivation?

fix #466
### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [ ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

PR #475, opened 2024-02-06
Add refaster recipes to simplify expected value

timo-abele:
## What's changed?

3 simple refaster recipes added

## What's your motivation?

salvaging the changes in https://github.com/openrewrite/rewrite-testing-frameworks/pull/472
adding missing transformations I noticed.

These recipes are simpler that the ones in https://github.com/openrewrite/rewrite-testing-frameworks/pull/472: they don't need generics and should work in Java 8. But each test yields: 
> Recipe was expected to make a change but made no changes.

@timtebeek 

SimplifyChainedAssertJAssertion could probably be extended to cover these cases, but I think refaster is the appropriate level of complexity
### Checklist


src/main/java/org/openrewrite/java/testing/assertj/SimplifyAssertExpectations.java
```diff
@@ -0,0 +1,80 @@
...
+public class SimplifyAssertExpectations {
+
+    @RecipeDescriptor(
+            name = "Simplify `assertThat(int).isEqualTo(0)`",
+            description = "Simplify `assertThat(int).isEqualTo(0)` to `assertThat(int).isZero()`."
+    )
+    public class SimplifyToIsZero {
+
+        @BeforeTemplate
+        void before(Integer i) {
```
knutwannheden, line 37: I think you want to use primitive types in the method signatures.

timo-abele: I tried it for both "Integer" recipes, but still get "Recipe was expected to make a change but made no changes." for all tests.

timtebeek:
Hi @timo-abele ; I've started two new efforts to (better) leverage the work done by our friends at Picnic:
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/527

I believe that could replace the recipes you've added here, given that their are already similar:
- https://error-prone.picnic.tech/refasterrules/AssertJIntegerRules/#abstractintegerassertiszero
- https://error-prone.picnic.tech/refasterrules/AssertJObjectRules/#assertthatisisequalto

Would you agree with closing this effort and instead replying on the upstream EPS project for these improvements?

timtebeek:
Closing this effort for now, as we need an upstream fix in rewrite-templating first, and in parallel we also have this PR that might cover some of the same
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/556

Note that the work here is much appreciated though! It's just the generics that are tripping us up to merge this as-is without additional effort put into https://github.com/openrewrite/rewrite-templating/pull/91 first.

Hope to revisit as soon as support is there though, and linking this issue into that one to be notified.

timtebeek:
Upstream we've seen a fix introduced
- https://github.com/openrewrite/rewrite/pull/4427

That means we now should see results from the existing ErrorProne Support recipes we pull in
https://github.com/openrewrite/rewrite-testing-frameworks/blob/3412efbd62dd4d155dd633b1674e54a1cdd023db/src/main/resources/META-INF/rewrite/assertj.yml#L31-L45

That reduces the need to reimplement them here. Thanks again for uncovering that we had an issue which is now solved!

PR #484, opened 2024-02-20
don't use TempDirNonFinal on method parameters

### LINKED ISSUE #483: TempDirNonFinal wrongly modifies method parameters

```java
class A {
   @Test
   void fileTest(@TempDir final File tempDir) {
   }
}
```

no change

## What did you see instead?
final is removed

The restriction of being non-final is only valid for fields, not for parameters: https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/io/TempDir.html. It's not really a problem when the recipe removes the final modifier of a method parameter, unless you have some additional tooling or IDE formatter adding the final keyword to all method parameters by default. :)

Yes.

Bananeweizen:
Fixes #483.

## What's changed?
apply TempDirNonFinal only to fields, not method parameters
Without the main code fix, the modified test fails.

The method for detecting whether the variables are fields or method arguments has been copied from another recipe. There are multiple (different) implementations of this method all over the place, and I have no idea if the other implementations are better or not. Someone from the core team might want to introduce an actual API method for that to avoid these local duplicates.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/TempDirNonFinal.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/TempDirNonFinal.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.testing.junit5;
 
+import org.openrewrite.Cursor;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for the immediate fix!

PR #485, opened 2024-02-27
JMockit to Mockito Recipe - `eq` Support, Better Type Propagation, and Error Handling

tinder-dthomson:
## What's changed?

This PR includes these major changes:

- Expectations with raw values mixed with argument matchers should use `eq`
- Preserve the type information of Mockito `any(Class)` invocation arguments
- Fix infinite loop when rewriting `Expectations` block fails
- Major refactor of `ArgumentMatchersRewriter` to simplify logic and add clarity 

## What's your motivation?

Continue improving the recipe, make it easier to understand.

### Checklist


timtebeek (approved):
Nice improvements, thanks! Especially good to see the infinite loop resolved, as that had tripped up one or two users that I know. I also like the more specific `eq` matchers. Should help folks that want to migrate still.

Issue #497, opened 2024-03-13
AssertJ skips `is`


timo-abele:
## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a **multi** module project.
```xml
    <profile>
        <id>openrewrite</id>
        <build>
            <plugins>
                <plugin>
                    <groupId>org.openrewrite.maven</groupId>
                    <artifactId>rewrite-maven-plugin</artifactId>
                    <version>5.23.1</version>
                    <configuration>
                        <activeRecipes>
                            <recipe>org.openrewrite.java.testing.assertj.Assertj</recipe>
                        </activeRecipes>
                        <failOnDryRunResults>true</failOnDryRunResults>
                    </configuration>
                    <dependencies>
                        <dependency>
                            <groupId>org.openrewrite.recipe</groupId>
                            <artifactId>rewrite-testing-frameworks</artifactId>
                            <version>2.4.1</version>
                        </dependency>
                    </dependencies>
                </plugin>
            </plugins>
        </build>
    </profile>
```

## What is the smallest, simplest way to reproduce the problem?

```java
import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.core.Is.is;
import org.junit.jupiter.api.Test;

class DebugTest {

    class Foo {
        int i = 8;
    }

    @Test
    void ba() {
        assertThat(System.out, is(System.out));
        assertThat(new Foo(), is(new Foo()));
    }

}
```

## What did you expect to see?

```java
import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
import org.junit.jupiter.api.Test;

class DebugTest {

    class Foo {
        int i = 8;
    }

    @Test
    void ba() {
        assertThat(System.out).isEqualTo(System.out);
        assertThat(new Foo()).isEqualTo(new Foo());
    }

}
```

## What did you see instead?

no change

## Comments
I'm a bit surprised that this seemingly simple case is not covered. I had added a custom class suspecting that this was the issue but it fails even for jdk classes. The same execution that skipped the case listed above, has successfully converted `assertThats` with custom types in other test classes...
**Update**: some instances of `is` on primitives are skipped as well while others are converted. Is it about the imports?

timtebeek:
Hmm; I'd have expected this to work through the recipe that we have for that migration
[...]

PR #498, opened 2024-03-13
Hamcrest `Is` matcher from core

timo-abele:
this testcase uses a different import for the is matcher

## What's changed?

added a testcase where the is matcher is imported via `import static org.hamcrest.core.Is.is;`

## What's your motivation?

Fix https://github.com/openrewrite/rewrite-testing-frameworks/issues/497

### Checklist


timtebeek:
Thanks for the helpful test to reproduce & pointing out the likely cause. We can either change the Java recipe, or run a declarative recipe before this change to get this to work, as indicated here.

An addition of this recipe 
```yaml
  - org.openrewrite.java.ChangeMethodTargetToStatic:
      methodPattern: org.hamcrest.core.Is is(..)
      fullyQualifiedTargetTypeName: org.hamcrest.Matchers
```

As the very first recipe here would allow both `RemoveIsMatcher` and `HamcrestIsMatcherToAssertJ` to work cases that currently use `org.hamcrest.core.Is`.

https://github.com/openrewrite/rewrite-testing-frameworks/blob/bcb4449f913f43ca2820ce18214d4399f7617833/src/main/resources/META-INF/rewrite/hamcrest.yml#L41-L46

Might be worth exploring if there's similar cases that could benefit from such a standardization-ahead-of-migration as well, but optional for this PR.

Thanks for helping figure this one out!

timtebeek (approved):

timtebeek:
Thanks for getting this started @timo-abele! I've moved the test and applied the above suggestion such that it now passes. Good to have this covered!

Issue #491, opened 2024-03-07
JUnitAssertNotNullToAssertThat does not add import when custom classes are the arguments


timo-abele:
Probably same issue as https://github.com/openrewrite/rewrite-testing-frameworks/issues/479, probably worth investigating which other recipes are affected.

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
            <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>5.23.1</version>
                <configuration>
                    <activeRecipes>
                        <recipe>org.openrewrite.java.testing.assertj.Assertj</recipe>
                    </activeRecipes>
                </configuration>
                <dependencies>
                    <dependency>
                        <groupId>org.openrewrite.recipe</groupId>
                        <artifactId>rewrite-testing-frameworks</artifactId>
                        <version>2.4.1</version>
                    </dependency>
                </dependencies>
            </plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
import static org.junit.jupiter.api.Assertions.assertNotNull;

import org.junit.jupiter.api.Test;

public class TTest {

    class A {}

    @Test
    public void testClass() {
        assertNotNull(new A());
    }
}
```

## What did you expect to see?

```java
import static org.assertj.core.api.Assertions.assertThat;

import org.junit.jupiter.api.Test;

public class TTest {

    class A {}

    @Test
    public void testClass() {
        assertThat(new A()).isNotNull();
    }
}
```

## What did you see instead?

```java
import org.junit.jupiter.api.Test;

public class TTest {

    class A {}

    @Test
    public void testClass() {
        assertThat(new A()).isNotNull();
    }
}
```

## What is the full stack trace of any errors you encountered?

```
stacktrace output here
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Probably won't find the time

timtebeek:
Thanks for the helpful before/after samples. Replicated indeed as a unit test
```java
    @Test
    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/491")
    void importAddedForCustomArguments() {
        rewriteRun(
          //language=java
          java(
            """
              import static org.junit.jupiter.api.Assertions.assertNotNull;

              import org.junit.jupiter.api.Test;

[...]

Issue #479, opened 2024-02-16
JUnitAssertEqualsToAssertThat does not add import when custom classes are the arguments


timo-abele:
## What version of OpenRewrite are you using?

I am using

- Maven plugin v5.22.0
- rewrite-testing-frameworks v2.3.2

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
      <plugin>
        <groupId>org.openrewrite.maven</groupId>
        <artifactId>rewrite-maven-plugin</artifactId>
        <version>5.22.0</version>
        <configuration>
          <activeRecipes>
            <recipe>org.openrewrite.java.testing.assertj.JUnitAssertEqualsToAssertThat</recipe>
          </activeRecipes>
        </configuration>
        <dependencies>
          <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-testing-frameworks</artifactId>
            <version>2.3.2</version>
          </dependency>
        </dependencies>
      </plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
package org.example;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;

class ATest {

  @Test void testEquals() {
    Assertions.assertEquals(new OwnClass(), new OwnClass());
  }

  public record OwnClass(String a) {

    public OwnClass() {this("1");}
  }
}
```
It looks like the parameters have to be classes that the user has defined themselves.
(The issue can be reproduced with classes from the standard library if the second parameter (`actual`) is `null`, however then there is also a method call ambiguity after the user has manually added the missing import, and `null` has to be cast to something. But that is a separate issue)

## What did you expect to see?

```java
package org.example;

import static org.assertj.core.api.Assertions.assertThat;

import org.junit.jupiter.api.Test;

class ATest {

  @Test void testEquals() {
    assertThat(new OwnClass()).isEqualTo(new OwnClass());
  }

  public record OwnClass(String a) {

    public OwnClass() {this("1");}
  }
}
```

## What did you see instead?

```java
package org.example;

import org.junit.jupiter.api.Test;

class ATest {

  @Test void testEquals() {
    assertThat(new OwnClass()).isEqualTo(new OwnClass());
  }

  public record OwnClass(String a) {

    public OwnClass() {this("1");}
  }
}
```

## What is the full stack trace of any errors you encountered?

```
stacktrace output here
```

timtebeek:
[...]

PR #505, opened 2024-04-14
Fix performance regression in AssertJ recipes

philippe-granet:
## What's changed?
Since commit https://github.com/openrewrite/rewrite-testing-frameworks/commit/368384a174bd92d49880c19a879fb9e14afcfa27 , AssertJ recipes are really slower since calling `contextSensitive()` method  disable a cache on JavaTemplate.

Fixes:
- Make sure there is a static import org.assertj.core.api.Assertions.* , even if not referenced (see #491 and #479)
- reintroduced assertionsParser cache in JUnitAssertEqualsToAssertThat

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertNotNullToAssertThatTest.java
+++ b/src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertNotNullToAssertThatTest.java
@@ -246,8 +246,9 @@ class JUnitAssertNotNullToAssertThatTest implements RewriteTest {
     @Test
     @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/491")
     void importAddedForCustomArguments() {
+        //language=java
         rewriteRun(
-          //language=java
+          spec -> spec.typeValidationOptions(TypeValidation.none()),
           java(
             """
               import org.junit.jupiter.api.Test;
--- a/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertArrayEqualsToAssertThat.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertArrayEqualsToAssertThat.java
@@ -73,7 +73,8 @@ public class JUnitAssertArrayEqualsToAssertThat extends Recipe {
             Expression expected = args.get(0);
             Expression actual = args.get(1);
 
-            maybeAddImport("org.assertj.core.api.Assertions", "assertThat");
+            // Make sure there is a static import for "org.assertj.core.api.Assertions.assertThat" (even if not referenced)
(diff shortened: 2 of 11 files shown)


philippe-granet:
It dosen't fix the problem, close this PR

src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertNotNullToAssertThatTest.java
```diff
@@ -246,8 +246,9 @@ private String notification() {
     @Test
     @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/491")
     void importAddedForCustomArguments() {
+        //language=java
         rewriteRun(
-          //language=java
+          spec -> spec.typeValidationOptions(TypeValidation.none()),
```
knutwannheden, line 251: As it seems like this test passed the type validation before, I wonder why it doesn't anymore.

knutwannheden (commented):
I think we can indeed make most of these templates context-free, as they only depend on the template parameters.

src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertNotNullToAssertThatTest.java
timtebeek: That test links to this issue, which was the reason to add contextSensitive as a quick solution at the time, with unfortunate runtime performance impact.
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/491

Those custom classes are not available when we drop `.contextSensitive()`, and the missing types would then break chained recipe runs to for instance adopt more optimal assertion patterns with AssertJ, as part of the migration.

Not quite sure where that leaves this effort. For a one time migration I'd lean towards taking the performance penalty for a more complete / idiomatic migration, but I don't know what prompted the need to fix the performance issue here. Any context there @philippe-granet ?

knutwannheden: I see. The template parameter's type is defined on the source path. That is a rather special case of course. So the test would also fail for a context-sensitive recipe, if the `A` class were defined in another source file.

I am not sure, but I would probably merge the PR and live with this limitation for now.

Thoughts?

timtebeek: Ok;  I'll trust your judgement on that then and get this merged as is. Thanks a lot both!

timtebeek (approved):
Thanks a lot for noticing, diagnosing and fixing the performance issue with these AssertJ recipes @philippe-granet !

PR #510, opened 2024-04-21
Recipe to make tests throw at most a single Exception

Bananeweizen:
## What's changed?
New recipe for replacing exception subclasses in test method signatures by java.lang.Exception, to improve readability. I'm not sure if there is any static code analysis guideline demanding this. I _am_ sure that there are different opinions by developers whether to declare `void test throws MyVerySpecialException` or just `...Exception`. Nevertheless I do need this recipe at work, and maybe it's useful for others, too.

You may want to look at the DocumentExample in the test to see the main idea. Feel free to suggest a different name or description.
The "is this a test" detection has been copied from another recipe. I'm not sure if it would be useful to have some helper classes there to unify that.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/cleanup/SimplifyTestThrows.java
+++ b/src/main/java/org/openrewrite/java/testing/cleanup/SimplifyTestThrows.java
@@ -16,18 +16,13 @@
 package org.openrewrite.java.testing.cleanup;
 
 import org.openrewrite.*;
-import org.openrewrite.internal.ListUtils;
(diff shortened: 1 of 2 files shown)


timtebeek:
Nice addition! Still have to go over it in more details, but I'm thinking we can name this something like `TestMethodSingleThrow`, to signify what the recipe will do beyond `SimplifyTestThrows`. Any thoughts there? Open to suggestions; just wondering if simplify is perhaps too broad.

timtebeek (approved):
Great addition, thanks for contributing this!

PR #514, opened 2024-05-16
Added to notMatchers

kevin-scholten:
## What's changed?
I've added 2 'notMatcher's to HamcrestNotMatcherToAssertJ.

## What's your motivation?
 I've added two notMatcher's that were not automatically changed in my project. By adding these two it fixed my problem.

-

-

-

## Any additional context
-

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/hamcrest.yml
+++ b/src/main/resources/META-INF/rewrite/hamcrest.yml
@@ -275,6 +275,12 @@ recipeList:
   - org.openrewrite.java.testing.hamcrest.HamcrestNotMatcherToAssertJ:
       notMatcher: hasItem
       assertion: doesNotContain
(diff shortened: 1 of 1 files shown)


timtebeek:
Awesome; thanks a lot for the immediate fix! Good to see you worked out what to add where, and able to confirm that it actually worked.

On a wider note: did you already notice [Apache Wicket has some OpenRewrite migration recipes](https://github.com/apache/wicket/tree/master/wicket-migration) as well? I added those a while back, and think those might be interesting for Topicus as well, especially as that [also runs some lambda migration recipes](https://github.com/apache/wicket/blob/f379f9ae0a472339252f813d17b49f7e62842c72/wicket-migration/src/main/resources/META-INF/rewrite/wicket.yml#L7-L8). Hope that helps!

If you find any more issues or have other ideas let me know! Glad to see the project put to use there. :)

Issue #513, opened 2024-05-02
Convert AssertJ's `assertThat(String).isEqualTo("")` to `isEmpty()`


timtebeek:
## What problem are you trying to solve?
Use most expressive AssertJ assertions instead of sub optimal patterns seen in the wild.

## Additional context
Separate from the following issue which migrates, this can instead go into recipes that fit our best practices.
 - https://github.com/openrewrite/rewrite-testing-frameworks/issues/357

PR #527, opened 2024-06-15
Add Picnic AssertJ rules to AssertJ best practices

timtebeek:
## What's your motivation?
Leverage existing Refaster recipes to apply more small fixes for users of AssertJ

I'd tried to remove `AbstractStringAssertStringIsEmptyRecipe` to replace `IsEqualToEmptyString`, but that's not matching, possibly due to generics. (cc @knutwannheden)

## Additional context
- #513
- #525


knutwannheden:
Good question. I think we need to debug it.

timtebeek:
> Good question. I think we need to debug it.

Thanks! I've logged this one to explore further outside this module

I think we can already merge this PR such that we can continue to use Picnic recipes for
- #520

PR #528, opened 2024-06-17
Add Kotlin support for Junit 5 recipes 1: CleanupJUnitImports

amitojduggal:
## What's changed?
Some of the recipes were not working with Kotlin code bases, i would like to add this one as a first example and will change the rest of the recipes the same way after the feedback. More MRs to follow. 

The change is around the usage of visitCompilationUnit, this visitor doesn't work with Kotlin codebases , i made the change to use preVisit, and this seems to fix it.

## What's your motivation?
we would like to make the recipes work on Kotlin codebases

Please let me know if this is the correct way, especially to add tests for both Kotlin and Java together like the way I have done.

@timtebeek @knutwannheden 

Nope

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -43,6 +43,7 @@ dependencies {
 
     testImplementation("org.openrewrite:rewrite-java-17")
     testImplementation("org.openrewrite:rewrite-groovy")
+    testImplementation("org.openrewrite:rewrite-kotlin:$rewriteVersion")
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Great work here @amitojduggal ! I've pushed two small changes to separate commits just to polish this up, but otherwise good to merge. Thanks a lot for getting this started!

amitojduggal:
Thanks a lot for the quick feedback @timtebeek :), I will continue working on similar changes.

PR #530, opened 2024-06-19
Jmockit Expectations with no times or result should be transformed to Mockito verify statement

shivanisky:
Currently Jmockit Expectation statements when they do not have a times or result, when migrated to mockito disappear. However, when migrated to mockito, it should have a corresponding verify statement as per following documentation:
1. https://www.baeldung.com/jmockit-expectations section 2.4 "Then we used the default behavior (if no repetition constraint is given minTimes = 1; is used) to define that at least one invocation will be done to methodForTimes2()"
2. https://jmockit.github.io/tutorial/Mocking.html section 6.2

From my experience, when people specify this in the Expectation, it almost always corresponds to a mockito verify with no explicit times, basically saying that this statement occurs exactly once. 
In mockito usually times is used to specify the exact number of invocations and this gives more finegrained verification, whereas the JMockit Expectation the developer typically intends "exactly once" when specifying no times even though it really means at least once.

If we prefer to use at least once, I can modify it as such, but a simple verify with no times is usually what the original author of the JMockit Expectation intends.

### Checklist below done

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/jmockit/ExpectationsBlockRewriter.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/ExpectationsBlockRewriter.java
@@ -17,12 +17,12 @@ package org.openrewrite.java.testing.jmockit;
 
 import org.openrewrite.Cursor;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.tree.*;
 
-import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -128,15 +128,20 @@ class ExpectationsBlockRewriter {
             removeExpectationsStatement();
         }
 
-        // TODO doesn't cover cases with multiple times eg minTimes and maxTimes - this is typically rare scenario
+        boolean hasTimes = false;
         if (mockInvocationResults.getTimes() != null) {
+            hasTimes = true;
             writeMethodVerification(invocation, mockInvocationResults.getTimes(), "times");
-        } else if (mockInvocationResults.getMinTimes() != null) {
+        }
+        if (mockInvocationResults.getMinTimes() != null) {
+            hasTimes = true;
             writeMethodVerification(invocation, mockInvocationResults.getMinTimes(), "atLeast");
-        } else if (mockInvocationResults.getMaxTimes() != null) {
+        }
+        if (mockInvocationResults.getMaxTimes() != null) {
+            hasTimes = true;
             writeMethodVerification(invocation, mockInvocationResults.getMaxTimes(), "atMost");
-        } else if (!hasExpectationsResults) {
-            // no times, no results
+        }
+        if (!hasExpectationsResults && !hasTimes) {
             writeMethodVerification(invocation, null, null);
         }
     }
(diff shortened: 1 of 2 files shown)


timtebeek (commented):
Changes look great to me already; indeed seems better not to complete drop `Expectations` without adding a corresponding `verify` that the method is at least invoked. Even makes me slightly worry for replacements done previously(!). I don't have prior exposure to JMockit though, so I appreciate the links to attached to the description.

We'll give @tinder-dthomson a little while to chime in, and if no response is received than I'd say we can merge this as is. Thanks a lot!

timtebeek (approved):

tinder-dthomson:
Hey folks! Thanks for introducing this change.

This behavior was not an oversight but rather a conscious decision on my part. My thought was that if a test wants to verify an invocation for a method without results, it belongs in the `Verifications` block. However as @shivanisky points out, `Expectations` block can be used for the same purpose. We should not remove these verifications.

src/main/java/org/openrewrite/java/testing/jmockit/ExpectationsBlockRewriter.java
```diff
@@ -118,21 +119,25 @@ private void rewriteMethodBody(List<Statement> expectationStatements) {
...
+        boolean hasExpectationsResults = !mockInvocationResults.getResults().isEmpty();
+        if (hasExpectationsResults) {
             // rewrite the statement to mockito if there are results
             rewriteExpectationResult(mockInvocationResults.getResults(), invocation);
         } else if (nextStatementCoordinates.isReplacement()) {
             // if there are no results and the Expectations block is not yet replaced, remove it
             removeExpectationsStatement();
         }
+
+        // TODO doesn't cover cases with multiple times eg minTimes and maxTimes - this is typically rare scenario
```
tinder-dthomson, line 131: Do we need to change this to `else if` for these statements? IIRC using sequential `if` statements as before would write each verification without issue.

src/main/java/org/openrewrite/java/testing/jmockit/ExpectationsBlockRewriter.java
```diff
@@ -22,6 +22,7 @@
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.tree.*;
 
+import javax.annotation.Nullable;
```
tinder-dthomson, line 25: I would avoid `javax.annotation` due to the Java EE / Jakarta EE debacle. No other recipes in this repo use `javax.annotation.Nullable`, I'd recommend we keep it that way.

src/main/java/org/openrewrite/java/testing/jmockit/ExpectationsBlockRewriter.java
shivanisky: You know I had the same thought last night a few hours before your message and I messaged Tim to hold off on merge to add a commit. Will fix.

src/main/java/org/openrewrite/java/testing/jmockit/ExpectationsBlockRewriter.java
shivanisky: I don't know about that debacle - more details? Do you you also prefer not using null as a valid argument? Perhaps it's better as I remember having a discussion around this with a reviewer l last year and ended up not using null, I forgot about it!

tinder-dthomson: It's a big pain related to the ownership of the Java EE transferring from Oracle to the Eclipse Foundation. Oracle apparently required them to change the naming to differentiate between Oracle and non-Oracle owned software.

See the [original Oracle post](https://blogs.oracle.com/javamagazine/post/transition-from-java-ee-to-jakarta-ee). Also see how the Spring community made the big decision to [move to Jakarta EE](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#jakarta-ee) in Spring Boot 3.

In practice, this means that the `javax.annotation` package is no longer maintained and has been effectively replaced by `jakarta.annotation` (along with the dependencies themselves). It's also apparently not a good idea to mix the two together. So either you should be using all Java EE (effectively tech debt), or use all Jakarta EE in your project.

In my software I avoid Java EE unless the project I'm working in already has them. `@Nullable` is nice but it's not functionally necessary.

timtebeek: Thanks for pointing this out @tinder-dthomson ; I've now codified this check in https://github.com/openrewrite/rewrite-recommendations/commit/3d724859e0137bc4ea3d7be845e631f5ed9eddfd, such that it will run on all PR going forward.

PR #532, opened 2024-06-20
Rewrite both JMockit  `@Mocked` and `@Injectable` annotated arguments

shivanisky:
See https://jmockit.github.io/tutorial/Mocking.html#injectable
As we are already doing with @Mocked annotation, also doing the same with @Injectable ie migration of method params with this annotation and fields with this annotation

### Checklist


timtebeek:
Great to see `Injectable` covered too; thanks @shivanisky ! I've changed the implementation a bit, at first to add an option to remove the extending classes, but right after settling on just having one recipe that replaces both annotations as once. Hope you agree!

timtebeek (approved):

PR #533, opened 2024-06-20
Add Kotlin support for JUnit recipes: UpdateBeforeAfterAnnotations

amitojduggal:
## What's changed?

- Adding support for Kotlin to the UpdateBeforeAfterAnnotations recipe.
- Adding Kotlin tests for AddParameterizedTestAnnotation
- More to come

## What's your motivation?
Would like to make the recipes to work both with Java and Kotlin.

@timtebeek @knutwannheden 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
@@ -18,12 +18,13 @@ package org.openrewrite.java.testing.junit5;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.InMemoryExecutionContext;
-import org.openrewrite.Issue;
 import org.openrewrite.java.JavaParser;
+import org.openrewrite.kotlin.KotlinParser;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.kotlin.Assertions.kotlin;
 
 class AddParameterizedTestAnnotationTest implements RewriteTest {
     @Override
@@ -31,10 +32,11 @@ class AddParameterizedTestAnnotationTest implements RewriteTest {
         spec
           .parser(JavaParser.fromJavaVersion()
             .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9", "junit-jupiter-params-5.9"))
+          .parser(KotlinParser.builder()
+            .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9", "junit-jupiter-params-5.9"))
           .recipe(new AddParameterizedTestAnnotation());
     }
 
-    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/314")
     @Test
     @DocumentExample
     void replaceTestWithParameterizedTest() {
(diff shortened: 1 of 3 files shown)


src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
```diff
@@ -18,23 +18,25 @@
...
     public void defaults(RecipeSpec spec) {
         spec
           .parser(JavaParser.fromJavaVersion()
             .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9", "junit-jupiter-params-5.9"))
+          .parser(KotlinParser.builder()
+            .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9", "junit-jupiter-params-5.9"))
           .recipe(new AddParameterizedTestAnnotation());
     }
 
-    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/314")
```
amitojduggal, line 37: The issue was fixed, the tag can be removed.

timtebeek: Thanks! We usually leave these issue marker in though, as they provide context as to why a test was added in a quickly accessible manner.

amitojduggal: makes sense i will bring it back.

src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
```diff
@@ -102,6 +135,35 @@ void testIsOdd(int number) {
...
+              import org.junit.jupiter.api.Assertions.assertTrue
+
+              class NumbersTest {
+                  @ParameterizedTest
+                  @ValueSource(ints = [1, 3, 5, -3, 15, Int.MAX_VALUE])
+                  fun testIsOdd(number: Int) {
+                      assertTrue(number % 2 != 0)
+                  }
+              }
+              """
```
timtebeek, line 166: I appreciate how thorough you've been to add all these Kotlin test blocks @amitojduggal , although I think going forward we'd only need to prove Kotlin works with a single test, perhaps the `@DocumentExample`, without a need to duplicate the assertions for every test scenario. Then later we can only add explicit Kotlin tests where we see issues, or want different handling as compared to the Java implementation. That would better allow us to see where there might be something of note by having a Kotlin test, as opposed to having them for all tests. What are your thoughts there?

src/test/java/org/openrewrite/java/testing/junit5/UpdateBeforeAfterAnnotationsTest.java
```diff
@@ -92,10 +119,34 @@ void after() {
...
+            """
+              import org.junit.jupiter.api.AfterEach
+
+              class Test {
+
+                  @AfterEach
+                  fun after() {
+                  }
+              }
+              """
```
timtebeek, line 144: For this class in particular we're only using the ChangeType recipes, which are extensively tested elsewhere, such that I think we can remove all but one of these Kotlin assertions from the tests.

src/test/java/org/openrewrite/java/testing/junit5/UpdateBeforeAfterAnnotationsTest.java
```diff
@@ -148,12 +223,31 @@ void afterClass() {
...
           )
         );
     }
 
-    @Issue("https://github.com/openrewrite/rewrite/issues/150")
-    @Disabled
     @Test
```
timtebeek, line 251: Ideally we retain this historical information, such that we have access to more context from the tests without having to annotate using the git history.
```suggestion
    @Issue("https://github.com/openrewrite/rewrite/issues/150")
    @Test
```

src/test/java/org/openrewrite/java/testing/junit5/AddParameterizedTestAnnotationTest.java
amitojduggal: Thanks @timtebeek , it makes a lot of sense, and it would help us speed it up. I can make these changes for both of the test classes.

src/test/java/org/openrewrite/java/testing/junit5/UpdateBeforeAfterAnnotationsTest.java
amitojduggal: Done @timtebeek

amitojduggal:
Its done @timtebeek , i addressed the review comments. Thanks for the feedback

timtebeek (approved):
Great to see @amitojduggal ! Thanks for applying that review feedback so quickly too.

PR #539, opened 2024-06-26
Add a NoMissingTypes Precondition to CleanupMockitoImports

gideon-sunbit:
## What's changed?
Add a NoMissingTypes Precondition to CleanupMockitoImports, to prevent changes if types are missing

## What's your motivation?
Fix issue where `org.mockito.kotlin` imports were being removed although they were in use.
Following suggestion by @timtebeek in Slack channel.

Possibly the unit test should use a Kotlin code block and not a Java one? not sure about this.

@timtebeek 

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,6 +55,7 @@ dependencies {
     testRuntimeOnly("net.datafaker:datafaker:latest.release") {
         exclude(group = "org.yaml", module = "snakeyaml")
     }
+    testRuntimeOnly("org.mockito.kotlin:mockito-kotlin:latest.release")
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks for getting this started! Didn't realize in Slack that it was Mockito Kotlin imports that were incorrectly removed. With that knowledge I've now added special handling, as we kind of expect missing types when it comes to converting Mockito test classes. Hope this works for your real use case too; do let me know when you try [our snapshot versions](https://docs.openrewrite.org/reference/snapshot-instructions).

PR #540, opened 2024-06-27
Migrate JMockit Verifications to Mockito

shivanisky:
Tests are failing, need to fix and also add tests once they pass. The import for verify is not being added. I don't know why this is not working as it is working for the when statements and the same code is being used.

https://jmockit.github.io/tutorial/Mocking.html#verification
https://www.javadoc.io/doc/org.jmockit/jmockit/1.45/mockit/Verifications.html

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
@@ -35,6 +35,7 @@ class JMockitBlockRewriter {
     private static final String THROW_TEMPLATE_PREFIX = "thenThrow(";
     private static final String LITERAL_TEMPLATE_FIELD = "#{}";
     private static final String ANY_TEMPLATE_FIELD = "#{any()}";
+    private static final String MOCKITO_IMPORT_FQN_PREFX = "org.mockito.Mockito";
 
     private static String getObjectTemplateField(String fqn) {
         return "#{any(" + fqn + ")}";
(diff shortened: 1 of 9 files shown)


tinder-dthomson:
> @tinder-dthomson Any thoughts on this PR from your experience with the frameworks & recipes?

I would not recommend removing the type validation, frustrating as it is. Usually this happens when you're trying to reference types that are not properly loaded, which shouldn't be a major problem in a unit test (just use standard java classes).

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
```diff
@@ -255,7 +274,8 @@ private static String getMockitoStatementTemplate(List<Expression> results) {
         return templateBuilder.toString();
     }
 
-    private static void appendToTemplate(StringBuilder templateBuilder, boolean buildingResults, String templatePrefix,
+    private static void appendToTemplate(StringBuilder templateBuilder, boolean buildingResults, String
+            templatePrefix,
```
timtebeek, line 278:
```suggestion
    private static void appendToTemplate(StringBuilder templateBuilder, boolean buildingResults, String templatePrefix,
```

timtebeek:
The type validation is only disabled selectively for method invocations on this one test that indeed uses a custom object.
https://github.com/openrewrite/rewrite-testing-frameworks/blob/be42e84677121c48d31eebfbfd0f0fd3f71bdb56/src/test/java/org/openrewrite/java/testing/jmockit/JMockitVerificationsToMockitoTest.java#L197-L200

We could modify this test to indeed use some class from the JDK itself to get around that, even if we would similarly lack type information in actual cases as well. That makes me inclined to keep this as it is, seeing how it's use is limited to a case we indeed can't seem to cover as easily. We've [discussed this in some more detail on Discord](https://discord.com/channels/996548279191998535/1245944114156146739/1257568743253606420); there's one option left to restore `.imports(fqn)` and add `.contextSensitive()`, but overall I'd say this is great work already, and need not be help up over this small imperfection that's common with other mocking frameworks as well.

timtebeek (approved):
Approved already from my side; let me know if either of you feels we should further explore the missing types for `verify` before we merged.

I'd say the missing types are acceptable here given the limited scope of these migration recipes, and hard to work around (in a performant manner) for limited added value, as there's no method or recipe chaining expected here.

tinder-dthomson (approved):
Great contribution!

PR #546, opened 2024-07-04
Add migration of JMockit NonStrictExpectations blocks as well as JMockit junit 4 runner

shivanisky:
Add migration of JMockit NonStrictExpectations blocks as well as JMockit junit 4 runner. Info on NonStrictExpectations:
https://javadoc.io/doc/org.jmockit/jmockit/1.22/mockit/NonStrictExpectations.html

@tinder-dthomson @timtebeek 

### Checklist


src/test/java/org/openrewrite/java/testing/jmockit/JMockitNonStrictExpectationsToMockitoTest.java
```diff
@@ -0,0 +1,1097 @@
...
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion()
+            .logCompilationWarningsAndErrors(true)
+            .classpathFromResources(new InMemoryExecutionContext(),
+              "junit-4.13.2",
+              "jmockit-1.22",
+              "mockito-core-3.12",
+              "mockito-junit-jupiter-3.12"
```
shivanisky, line 36: I will remove this line, not needed, as doing legacy JMockit migration which will be using junit 4 or lower

build.gradle.kts
```diff
@@ -20,6 +20,7 @@ recipeDependencies {
     parserClasspath("org.mockito:mockito-all:1.10.19")
     parserClasspath("org.mockito:mockito-core:3.+")
     parserClasspath("org.jmockit:jmockit:1.49")
+    parserClasspath("org.jmockit:jmockit:1.22")
```
tinder-dthomson, line 23: You can't add another version of the same library on the classpath. JMockit 1.49 is the last published version from 2019 so it should be the most appropriate choice.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
```diff
@@ -172,6 +178,9 @@ private void rewriteResult(J.MethodInvocation invocation, List<Expression> resul
             return;
         }
         visitor.maybeAddImport(MOCKITO_IMPORT_FQN_PREFX, "when");
+        if (this.blockType == NonStrictExpectations) {
+            visitor.maybeAddImport(MOCKITO_IMPORT_FQN_PREFX, "*");
```
tinder-dthomson, line 182: What's the thought behind this wildcard import?

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
```diff
@@ -211,9 +219,11 @@ private void rewriteVerify(J.MethodInvocation invocation, @Nullable Expression t
 
         // do this last making sure rewrite worked and specify hasReference=false because in verify case it cannot find
         // the static reference in AddImport class, and getSelect() returns not null
-        visitor.maybeAddImport(MOCKITO_IMPORT_FQN_PREFX, "verify", false);
-        if (verificationMode != null) {
-            visitor.maybeAddImport(MOCKITO_IMPORT_FQN_PREFX, verificationMode);
+        if (!rewriteFailed) {
```
tinder-dthomson, line 222: I don't think this check is necessary, as `maybeAddImport` will only add the import if the type is actually used.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
```diff
@@ -242,9 +251,13 @@ private J.Block rewriteTemplate(String verifyTemplate, List<Object> templatePara
                 );
     }
 
-    private static @Nullable String getWhenTemplate(List<Expression> results) {
+    private @Nullable String getWhenTemplate(List<Expression> results) {
```
tinder-dthomson, line 254: Can we remove `@Nullable`? I mentioned this in a previous PR but it's essentially tech debt unless you go straight to jakarta.

I noticed that we are using openrewrite.lang.internal package, but internal typically means it's not intended for users but rather the framework itself and is at risk to breaking changes.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
```diff
@@ -242,9 +251,13 @@ private J.Block rewriteTemplate(String verifyTemplate, List<Object> templatePara
...
         boolean buildingResults = false;
-        final StringBuilder templateBuilder = new StringBuilder(WHEN_TEMPLATE_PREFIX);
+        StringBuilder templateBuilder = new StringBuilder();
+        if (this.blockType == NonStrictExpectations) {
+            templateBuilder.append("lenient().");
```
tinder-dthomson, line 258: You need to add the import for lenient as well here.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
```diff
@@ -41,7 +41,7 @@ static Optional<JMockitBlockType> getJMockitBlock(Statement s) {
         }
 
         for (JMockitBlockType blockType : JMockitBlockType.values()) {
-            if (TypeUtils.isAssignableTo(blockType.getFqn(), clazz.getType())) {
+            if (TypeUtils.isOfClassType(clazz.getType(), blockType.getFqn())) {
```
tinder-dthomson, line 44: Why this change to `isOfClassType`?

src/main/resources/META-INF/rewrite/jmockit.yml
```diff
@@ -36,6 +36,9 @@ recipeList:
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: mockit.integration.junit5.JMockitExtension
       newFullyQualifiedTypeName: org.mockito.junit.jupiter.MockitoExtension
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: mockit.integration.junit4.JMockit
```
tinder-dthomson, line 40: I think this fqn is wrong, isn't it `JUnitRunner`?

src/test/java/org/openrewrite/java/testing/jmockit/JMockitNonStrictExpectationsToMockitoTest.java
tinder-dthomson: Shouldn't this be compatible with both JUnit 4 and 5?

src/main/resources/META-INF/rewrite/jmockit.yml
timtebeek: Luckily we support [using multiple versions of a library in a project](https://docs.openrewrite.org/authoring-recipes/multiple-versions), as we needed that for rewrite-spring as well. Ideally you'll want to use a single version if you can get away with it, but in the case of removed classes that's not always an option.

tinder-dthomson: Fantastic! In that case, this change should be good. I'll go ahead and update my other comment.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitNonStrictExpectationsToMockitoTest.java
shivanisky: For NonStrictExpectations, they would be using junit 4 or older as NonStrictExpectations is not there in jmockit 1.23 upwards. Last version was 1.22. Hence, would like to support this situation.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
shivanisky: Is because, it was treating NonStrictExpectations as an Expectations because NonStrictExpectations inherit from Expectations. So isAssignable includes inherited types.
https://javadoc.io/doc/org.jmockit/jmockit/1.22/mockit/NonStrictExpectations.html

Let me know if you have a better idea.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: I'm hitting an issue with the framework, where when I add lenient, the when keyword static reference can't be found. I've put in a workaround and added comments now. This was just a first draft attempt.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: Is an expensive check, especially if not needed, hence would like to skip it if not needed

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: Don't like adding constants if not needed in multiple places, or needed to easily modify, prefer keeping things in context within the context. What is reason for static final String?

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: I don't want to use Nullable, I just can't think of what to replace Expression times with. Any suggestions?

src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
tinder-dthomson: Got it, just was curious. Although there is a potential disadvantage - if a test uses a class that extends `Expectations`. I imagine this is a rare case though, so we can probably not worry about handling it.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
tinder-dthomson: Fair enough, I wasn't aware of the expensive nature of `maybeAddImport`, but we also already know that the rewrite failed at this point so I see no harm in this check.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
tinder-dthomson: No problem, if you want to keep working on it we can revisit when you move the PR out of Draft.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
tinder-dthomson: It's already a convention in this recipe to use static final String for template strings, so let's stay consistent.

It also makes it easier to identify what templates we are using, rather than forcing users to read through all the code.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitNonStrictExpectationsToMockitoTest.java
tinder-dthomson: I believe there's nothing preventing the usage of older JMockit versions with JUnit 5, so in my opinion we should support JUnit 5 migration as well.

JUnit 5 documentation seems to imply there are no compatibility concerns: https://jmockit.github.io/tutorial/Introduction.html#:~:text=JMockit%20supports%20(and%20requires)%20the,%2Fjmockit.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
```diff
@@ -131,6 +132,11 @@ private void rewriteMethodInvocation(List<Statement> statementsToRewrite) {
             rewriteResult(invocation, mockInvocationResults.getResults());
         }
 
+        if (blockType == NonStrictExpectations) {
```
tinder-dthomson, line 135: What happens in JMockit if `NonStrictExpectations` block contains a `times = 2` statement? Is that not verified?

build.gradle.kts
tinder-dthomson: @timtebeek do you know how this version will get picked in the `ChangeType` recipe added in jmockit.yml? I can easily see how using `classpathFromResources` allows us to explicitly choose the classpath dependencies, but I'm unclear on this case.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: Yes - all ready now. I think I am pretty much done with this PR

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: I need to add dependency for jakarta as not there, happy with this? Tried to reduce dependency on @Nullable as much as possible, but there is one case where it is being used as an argument. I could convert it to an Optional, let me know if there's a better option.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
shivanisky: Yes this is only for stubbing, there is link https://javadoc.io/doc/org.jmockit/jmockit/1.22/mockit/NonStrictExpectations.html

timtebeek (approved):
Great to see this once again! I'll give @tinder-dthomson a bit of time to go over this as well, before we merge and see this through.

src/main/java/org/openrewrite/java/testing/jmockit/JMockitBlockRewriter.java
tinder-dthomson: It's a bit unclear what the behavior is, due to this statement:

"The difference from [regular expectations](https://javadoc.io/static/org.jmockit/jmockit/1.22/mockit/Expectations.html) is that a non-strict recorded expectation is allowed to have no matching invocations **(unless otherwise specified).**"

I wonder if this means that if `times` is specific, does it get honored? I'm fine with making this assumption for now but I wouldn't be surprised if JMockit has the ability to mix both.

tinder-dthomson (approved):
This looks good to me overall! I will follow-up with another PR to completely remove `@Nullable`, I'd rather not block you any longer.

Thanks again for the contribution!

build.gradle.kts
timtebeek: We do a prefix match, that you can extend to include the version number like `jmockit-1.22` if you want to be specific. Rewrite-spring has the most variation in acceptable patterns if you want to [see some examples](https://github.com/search?q=repo%3Aopenrewrite%2Frewrite-spring%20classpathFromResources&type=code).

timtebeek:
> This looks good to me overall! I will follow-up with another PR to completely remove `@Nullable`, I'd rather not block you any longer.

We can keep the internal nullable annotations; I've restored a few in these classes already with https://github.com/openrewrite/rewrite-testing-frameworks/pull/546/commits/e906d82c4643e69e47039c7b1ea114ec3396cd79.
The idea is that we add non null by default to `package-info.java`, and then only have to be explicit using our internal nullable annotation as needed.

Thanks both! Great to see this come along.

shivanisky:
This is a very good point - perhaps times is honoured as an Expectation -
it seems from the documentation this is what would happen.

https://javadoc.io/doc/org.jmockit/jmockit/1.22/mockit/NonStrictExpectations.html

“
A non-negative value assigned to this field will be taken as the exact
number of times that invocations matching the current expectation should
occur during replay.”

I haven’t seen a times in a NonStrictExpectation but should support it … it
would get wiped for now. I’ll look at it next week.

I guess the clause to ignore times should be removed, and then if has
times, don’t add the lenient() and also add corresponding verify with the
times, as is already being done with Expectations.

I’ll try to add this next week.

tinder-dthomson:
> > This looks good to me overall! I will follow-up with another PR to completely remove `@Nullable`, I'd rather not block you any longer.
> 
> We can keep the internal nullable annotations; I've restored a few in these classes already with [e906d82](https://github.com/openrewrite/rewrite-testing-frameworks/commit/e906d82c4643e69e47039c7b1ea114ec3396cd79). The idea is that we add non null by default to `package-info.java`, and then only have to be explicit using our internal nullable annotation as needed.
> 
> Thanks both! Great to see this come along.

Thanks for chiming in! I was unaware of your previous touches related to `@Nullable`.
It's unfortunate the package contains `internal` if we're open to it being used by recipe authors, but this is a better alternative to using javax or jakarta package roots imo.

PR #549, opened 2024-07-10
JMockit NonStrictExpectations with times should be migrated to default when without lenient(), refactor

shivanisky:
## What's changed?
JMockit NonStrictExpectations with times should be migrated to default mockito when without lenient so it is strict, and should also have a corresponding verify with times. Also refactor so code is looking very nice and clean.

See below times method in:
https://javadoc.io/doc/org.jmockit/jmockit/1.22/index.html

## What's your motivation?

See below we can see that NonStrictExpectations become strict when a times is specified ie they behave just like Expectations

Screenshot 2024-07-10 235533

@tinder-dthomson @timtebeek 

### Checklist


timtebeek (approved):
Great work once again! I'll approve and merge already; if @tinder-dthomson has any feedback we can pick that up separately.

PR #550, opened 2024-07-11
Quick attempt at idomatic AssertJ assertions through Refaster

timtebeek:
## What's changed?
* Added rewrite-templating for Refaster recipe support
* A recipe that aims to rewrite `assertThat(actual).isEqualTo(0)` to assertThat(actual).isZero()`
 
## What's your motivation?
Get more idiomatic AssertJ assertions developed more easily, from having a first example here.

Currently fails as the before template isn't matched
```java
                final JavaTemplate before = JavaTemplate
                        .builder("org.assertj.core.api.Assertions.assertThat(#{actual:any(int)}).isEqualTo(0);")
                        .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()))
                        .build();
```
image

`org.assertj.core.api.AbstractIntegerAssert<Generic{SELF extends }>{name=isEqualTo,return=Generic{?},parameters=[int]}`

Explicit recipe would work for this case, but is less easily developed and maintained, whereas we expect more similar recipes to be welcome here.

## Any additional context
- https://github.com/openrewrite/rewrite-templating/issues/47

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,6 +33,7 @@ dependencies {
     implementation("org.openrewrite:rewrite-java")
     implementation("org.openrewrite:rewrite-gradle")
     implementation("org.openrewrite:rewrite-maven")
+    implementation("org.openrewrite:rewrite-templating:$rewriteVersion")
     implementation("org.openrewrite.recipe:rewrite-java-dependencies:$rewriteVersion")
     implementation("org.openrewrite.recipe:rewrite-static-analysis:$rewriteVersion")
     runtimeOnly("org.openrewrite:rewrite-java-17")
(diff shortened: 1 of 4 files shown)


timtebeek:
Likely a similar issue to what's explored in

timtebeek:
Closing this effort for now, as we need an upstream fix in rewrite-templating first, and in parallel we also have 
- https://github.com/openrewrite/rewrite-testing-frameworks/pull/556

timtebeek:
We should now already get this recipe applied
https://github.com/PicnicSupermarket/error-prone-support/blob/d94f65a1f03f16fe3d46198ef9658bc4eb41f9cf/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJIntegerRules.java#L42-L52

After the recent fix in
- https://github.com/openrewrite/rewrite/pull/4427

knutwannheden:
Indeed.
Screenshot_20240828-044626

PR #554, opened 2024-07-24
Simplify toString assertion

velo:
Instead of using:
```
assertThat(cat.toString()).isEqualTo("animal");
```

Could use:
```
assertThat(cat).hasToString("animal");
```

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
+++ b/src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
@@ -471,7 +471,7 @@ class MigrateChainedAssertToAssertJTest implements RewriteTest {
                   class MyTest {
                       void testMethod(Object argument) {
                           String s = "hello world";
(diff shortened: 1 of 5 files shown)


src/test/java/org/openrewrite/java/testing/assertj/MigrateChainedAssertToAssertJTest.java
```diff
@@ -455,4 +455,40 @@ void test(Iterator<String> iterator, Iterator<String> otherIterator) {
             );
         }
     }
+    
+    @Nested
+    class Objects {
+        
+        void objectoToStringReplacement() {
```
timtebeek, line 462:
```suggestion
        @Test
        void objectoToStringReplacement() {
```

timtebeek (approved):
Nice addition, thanks! Love how you keep turning up with more of these replacements, and how easy they've become to add and improve tests suites.

velo:
> Nice addition, thanks! Love how you keep turning up with more of these replacements, and how easy they've become to add and improve tests suites.

You never know when one of these might land me a role at open rewrite.

And I used it a lot on my day job

PR #555, opened 2024-07-24
ExpectedExceptionToAssertThrows should not wrap single statement

timtebeek:
As [discussed on Slack](https://rewriteoss.slack.com/archives/C01A843MWG5/p1721792132436429) with @shivanisky .


shivanisky:
Awesome thanks!

Issue #522, opened 2024-06-13
Convert JMockit `@Tested` and `Deencapsulation` when migrating to Mockito


timtebeek:
## What problem are you trying to solve?
We want to achieve a complete migration coming from JMockit going to Mockito.
It seems we don't yet cover either the `@Tested` annotation, nor the `Deencapsulation` class.

## Describe the situation before applying the recipe
_TODO gather clear before and after examples_

## Describe the situation after applying the recipe
_TODO gather clear before and after examples_

## Have you considered any alternatives or workarounds?
As a first step we could identify such cases and suggest to migrate these manually.

## Any additional context
This was a reported via email.

Issue #325, opened 2023-03-23
JMockit to Mockito


nmck257:
[JMockit](https://jmockit.github.io/) remains popular, but has not had a release since 2019, and has [various compatibility issues](https://github.com/jmockit/jmockit1/pull/736) with Java 11+.

As Java 8 loses LTS status, developers would likely benefit from some recipe(s) to facilitate migration from JMockit to Mockito

klauerm:
From my first research I found the following, that the recipe shall do:
1. Replace JMockit imports with Mockito imports:
Replace JMockit imports in your test classes with Mockito imports. For example, replace:
```java
import mockit.*;
```
with
```java
import org.mockito.Mockito;
import org.mockito.junit.MockitoJUnitRunner;
```

2. Refactor test class annotations:
Replace JMockit's `@RunWith(JMockit.class)` with Mockito's `@RunWith(MockitoJUnitRunner.class)`.

3. Refactor mocks:
Replace `@Mocked` with `@Mock` for mock objects. For example, replace:
```java
@Mocked
private MyClass myClassMock;
```
with
```java
@Mock
private MyClass myClassMock;
```

4. Refactor test method setup:
In JMockit, you would typically use `new Expectations()`, `new StrictExpectations()`, or `new Verifications()` blocks to set up and verify mock behaviors. In Mockito, you'll use methods like `when()`, `thenReturn()`, `verify()`, and `times()`.

For example, if you had a JMockit expectation like this:
```java
new Expectations() {{
    myClassMock.myMethod(input);
    result = output;
}};
```
You would refactor it to use Mockito like this:
```java
Mockito.when(myClassMock.myMethod(input)).thenReturn(output);
```

5. Refactor verifications:
If you had a JMockit verification like this:
```java
new Verifications() {{
    myClassMock.myMethod(input);
    times = 1;
}};
```
You would refactor it to use Mockito like this:
```java
Mockito.verify(myClassMock, Mockito.times(1)).myMethod(input);
```

@nmck257 Is there anything else, that is missing from your point of view?

nmck257:
Honestly, I've never used JMockit myself, but that looks like an excellent scope to start with. :)

Note that RunWith is JUnit4 syntax, so we may want to be mindful about how this recipe should interact with the JUnit 5 migration recipe(s)

hazendaz:
[...]

PR #557, opened 2024-07-28
Add unit tests for JMockit Delegate to mockito migration. Also add test case for comments

shivanisky:
At the moment, JMockit Delegates are not migrated to mockito as mentioned in issue
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/522. 

What I'm seeing is that they are being trashed with the template being printed out. These tests were written to try to replicate this issue, however I was unable to.
They may help anyone adding the feature for Delegate migration.

https://javadoc.io/doc/org.jmockit/jmockit/latest/mockit/Delegate.html

Also added one test case showing comments are not preserved.

I am wondering if it's better to have the test cases failing with the expected output and to disable them. Please let me know if that's preferred.

@timtebeek @tinder-dthomson 

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/jmockit/JMockitDelegateToMockitoTest.java
+++ b/src/test/java/org/openrewrite/java/testing/jmockit/JMockitDelegateToMockitoTest.java
@@ -17,6 +17,7 @@ package org.openrewrite.java.testing.jmockit;
 
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
 import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
@@ -36,6 +37,7 @@ class JMockitDelegateToMockitoTest implements RewriteTest {
         setDefaultParserSettings(spec);
     }
 
+    @DocumentExample
     @Test
     void whenNoArgsVoidMethod() {
         //language=java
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the continued work on this! Indeed probably best to have the tests fail with the actual expected output, as opposed to documenting the status quo. No need to mark them `@Disabled` just yet, we can do that just before a merge if we want to delay any work towards this, and then have the tests reference an open issue with more information.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitDelegateToMockitoTest.java
```diff
@@ -0,0 +1,364 @@
...
+class JMockitDelegateToMockitoTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        setDefaultParserSettings(spec);
+    }
+
+    @DocumentExample
+    @Test
+    void whenNoArgsVoidMethod() {
```
timtebeek, line 42: I'm wondering what the right issue to tag these tests with is; would that be 
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/325
```suggestion
    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/325")
    void whenNoArgsVoidMethod() {
```

timtebeek (commented):
Great to have these examples documented and runnable! One last question before we merge, but other than that good to go.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitDelegateToMockitoTest.java
shivanisky: I am thinking this one: https://github.com/openrewrite/rewrite-testing-frameworks/issues/522 - have also mentioned it in PR description (just added it today). But ... 325 works well too - I'll let you decide.

Oh thanks - I wasn't sure how to tag an issue in this repo, in the past I've done it in a different way, thanks for showing it to me.

timtebeek (approved):

Issue #464, opened 2024-01-18
convert `AssertTrue(false)` to `fail`


timo-abele:
## What problem are you trying to solve?

I have in my code some (Junit5) `AssertTrue(false, "this shouldn't happen")`, because someone didn't use `fail` when they should have.

## Describe the solution you'd like

The Junit5 cleanup recipe should convert such cases to `fail("this shouldn't happen")`.
The AssertJ best pratices recipe might add a similar recipe.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

no

timtebeek:
Hi @timo-abele ; Yes would fit right in with some of the other AssertTrue migration recipes we already have in
https://github.com/openrewrite/rewrite-testing-frameworks/tree/a426e409ca3b8e04c6470e86cc09d61346e7c0f4/src/main/java/org/openrewrite/java/testing/cleanup

We could even implement this as a Refaster recipe, similar to what we recently added to rewrite-static-analysis in
- https://github.com/openrewrite/rewrite-static-analysis/pull/231

Given that there's good examples to work from already I'm going to mark this one as good first issue; anyone welcome to pick this up!

PR #558, opened 2024-08-02
Convert `assertTrue(false, String)` and `assertFalse(true, String)` with `fail(String)`

Laurens-W:
## What's changed?

Add recipe to rewrite assertTrue(false, "message") and assertFalse(true, "message") to fail("message")

## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/464

Test coverage

@timtebeek @timo-abele 

### Checklist


timtebeek:
Awesome start, thanks! Did you notice this comment in the issue?

> We could even implement this as a Refaster recipe, similar to what we recently added to rewrite-static-analysis in
> https://github.com/openrewrite/rewrite-static-analysis/pull/231

I think we can keep the tests, and replace the recipes with just a couple lines in Refaster recipes.

Laurens-W:
> Awesome start, thanks! Did you notice this comment in the issue?
> 
> > We could even implement this as a Refaster recipe, similar to what we recently added to rewrite-static-analysis in
> > [openrewrite/rewrite-static-analysis#231](https://github.com/openrewrite/rewrite-static-analysis/pull/231)
> 
> I think we can keep the tests, and replace the recipes with just a couple lines in Refaster recipes.

Must admit I skimmed over that part :) I will continue working on this next week and apply the proposed changes!

timtebeek (changes requested):
Great to see! Let's merge the two separate test classes into a single `AssertLiteralBooleanToFailRecipeTest`, and within those I think we write a single test that covers all four cases from a single unit test that just has the four statements in one.

timtebeek (approved):
Neat addition; thanks @Laurens-W !

Issue #563, opened 2024-08-07
Removed unused checked exceptions, as discovered in JUnit 4 to 5 migration


shivanisky:
When running the junit 4 to 5 migration recipe, in the assertThrows case checked exceptions are not removed. This results in unused exceptions left in the method signature which are handled by the assertThrows case, and results in intellij warnings.

## Any additional context
@timtebeek has mentioned that this could be fixed in a similar way to the issue where unecessary catch blocks exist to catch exceptions not thrown by try blocks, and could be a separate recipe to fix all unecessary checked exceptions:
https://docs.openrewrite.org/recipes/staticanalysis/unnecessarycatch

Here is the code to detect unecessary exceptions:
https://github.com/openrewrite/rewrite-static-analysis/blob/6d04813574c6b02c8bf7d6e2f66b2bdaa881854c/src/main/java/org/openrewrite/staticanalysis/UnnecessaryCatch.java#L79-L94

PR #564, opened 2024-08-08
Remove throws from method that adopts `assertThrows`

timtebeek:
## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/563

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
@@ -26,6 +26,7 @@ import org.openrewrite.java.search.FindImports;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.*;
 import org.openrewrite.marker.Markup;
+import org.openrewrite.staticanalysis.UnnecessaryThrows;
 
 import java.util.Comparator;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java
```diff
@@ -149,6 +150,7 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
                                 .staticImports("org.junit.jupiter.api.Assertions.assertThrows")
                                 .build()
                                 .apply(updateCursor(m), m.getCoordinates().replaceBody(), cta.expectedException, lambda);
+                        m = (J.MethodDeclaration) new UnnecessaryThrows().getVisitor().visit(m, ctx, getCursor().getParentOrThrow());
```
timtebeek, line 153: As indicated; quite likely that we can not use `UnnecessaryThrows` directly, since that limits itself to avoid doing harm. We could create a separate visitor in rewrite-testing-frameworks that's a little more aggressive as we have a smaller scope to work with.

timtebeek:
Turns out we were overthinking this one; as we already replace the entire method body with `assertThrows`, there are then no exception left to be thrown that are not already caught.

shivanisky:
I'm just thinking with the above, what if the method throws multiple checked exceptions, and assertThrows only catches one, even if it is wrapping the entire method body. I think in that case, the remaining checked exception are still valid, or still not? From my memory I remember only removing one checked exception from the method signature, and keeping one when doing the junit 4 to 5 migration.

Issue #547, opened 2024-07-04
RemoveTryCatchFailBlocks should not rewrite try blocks that return


protocol7:
## What version of OpenRewrite are you using?
* org.openrewrite:rewrite-core:jar:8.28.1
* org.openrewrite.recipe:rewrite-testing-frameworks:jar:2.12.2

## What is the smallest, simplest way to reproduce the problem?
`org.openrewrite.java.testing.junit5.RemoveTryCatchFailBlocks` should not try to rewrite in the case where the `try {}` contains a `return`, see the test case for an example. This will create non-compiling code since the return is now in the `assertDoesNotThrow` lambda. We ran into this in our code base.

```java
  @Test
  void testRemoveTryCatchFailBlocksWithReturningTry() {
    rewriteRun(
            spec -> spec.recipe(new RemoveTryCatchFailBlocks()),
            java(
                    """
                    package com.helloworld;
        
                    import static org.junit.jupiter.api.Assertions.fail;
        
                    public class Foo {
        
                        String getFoo() {
                          try {
                            return "foo";
                          } catch (RuntimeException e) {
                            fail();
                          }
                        }
                    }"""));
  }
```

timtebeek:
Thanks yet again for the very clear runnable example; sounds like you're having fun over there cleaning up your tests. Much appreciated that you're reporting any issues that you come across, as that really helps us to improve.

In this case I think the fix is going to be very similar to [what we added yesterday](https://github.com/openrewrite/rewrite-testing-frameworks/commit/30f35fe919b8166c19ddf22e2f1ea1c0d918487a): a local visitor that in this case detects any `J.Return`, and if found we just return the unmodified input tree element.

protocol7:
Thanks for the pointer, I will see if I can come up with a fix for this.

PR #566, opened 2024-08-08
Dont convert try catch fail when return

Laurens-W:
## What's changed?

Add check for try bodies that contain a return statement

## What's your motivation?

Solves https://github.com/openrewrite/rewrite-testing-frameworks/issues/547
- Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/547

@timtebeek @protocol7 

### Checklist


timtebeek (approved):
Great work already; one case was missed for nested elements. I've added a quick pattern we use more often when we want to search downwards from a particular elements to handle that correctly. See if you can work out why and how that works. :)

PR #570, opened 2024-08-12
Context free SimplifyAssertJAssertion

timtebeek:
Should be quicker due to caching of templates.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/SimplifyAssertJAssertion.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/SimplifyAssertJAssertion.java
@@ -95,8 +95,7 @@ public class SimplifyAssertJAssertion extends Recipe {
             }
 
             // Assume zero argument replacement method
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/testing/assertj/SimplifyAssertJAssertion.java
```diff
@@ -95,11 +95,11 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation methodInvocat
             }
 
             // Assume zero argument replacement method
-            return JavaTemplate.builder(dedicatedAssertion + "()")
-                    .contextSensitive()
+            String receiverType = mi.getSelect().getType().toString().replaceAll("<.*>", "");
+            return JavaTemplate.builder("#{any(" + receiverType + ")}." + dedicatedAssertion + "()")
```
knutwannheden, line 99: I would instead try with this:
```suggestion
            return JavaTemplate.builder("#{any()}." + dedicatedAssertion + "()")
```
This should work (while keeping the `apply()` call as-is), since it should then retrieve the type from the method's select.

Issue #327, opened 2023-03-28
`org.junit.Assume.assumeThat()` is not migrated properly


knutwannheden:
It gets migrated to `org.junit.jupiter.api.Assumptions.assumeThat()` even though this doesn't exist in JUnit 5. It should instead either use JUnit 5's `org.junit.jupiter.api.Assumptions.assumeTrue()` or alternatively AssertJ's `assumeThat()`.

knutwannheden:
Going the [route of jSparrow](https://jsparrow.github.io/rules/replace-j-unit4-assumptions-with-hamcrest-j-unit.html#code-changes) via Hamcrest's `MatcherAssume`.

knutwannheden:
> Going the [route of jSparrow](https://jsparrow.github.io/rules/replace-j-unit4-assumptions-with-hamcrest-j-unit.html#code-changes) via Hamcrest's `MatcherAssume`.

Even though that would be easier, it feels backwards. Going for AssertJ's `Assumptions` instead.

knutwannheden:
Using Hamcrest is in a first steop probably better after all, because we don't yet have any migration from Hamcrest's matchers to AssertJ and JUnit's `assumeThat()` is based on Hamcrest matchers.

PR #577, opened 2024-08-14
Only add `hamcrest-junit` dependency when using `assertThat` or `assumeThat`

FieteO:
## What's changed?

Remove the `hamcrest-junit` dependency when running the `JUnit4to5Migration` recipe.

## What's your motivation?

The dependency is not needed in the Junit 5 context (AFAIK)


timtebeek:
How does this interact with this recipe a little further up/down? As we seem to add that explicitly in a JUnit 5 upgrade
https://github.com/openrewrite/rewrite-testing-frameworks/blob/b7d3b84556a165c07e3aa5e5bb71382beeda1eac/src/main/resources/META-INF/rewrite/junit5.yml#L92
https://github.com/openrewrite/rewrite-testing-frameworks/blob/b7d3b84556a165c07e3aa5e5bb71382beeda1eac/src/main/resources/META-INF/rewrite/junit5.yml#L166-L172

What makes that the dependency is not or no longer needed in your case?

FieteO:
Hmm, you are right. Probably some more investigation is needed as to when the dependency is needed/when it can be removed.
The `hamcrest-junit` seems to have been [the result of decoupling the the Hamcrest portions in Junit 4 into a separate dependency](https://github.com/hamcrest/JavaHamcrest/issues/92).
This dependency hasn't received an update since 9 years and it's hard to say what use it serves. The PR that added it to the `junit5.yml` was titled `Migrate JUnit 4 `Assume#assumeThat()` to `hamcrest-junit`` to fix issue #327 .

As far as I can tell, Junit 5 [also has support for Assumptions](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/Assumptions.html).
Maybe @knutwannheden can comment on this?

timtebeek:
This was a fun one @FieteO ; thanks for raising this for visibility. Turns out we were a little too eager adding that `hamcrest-junit` dependency on _any_ use of `org.junit.Assume`, as opposed to only when using `assertThat` or `assumeThat`. Solved now by creating a new scanning recipe that looks at methods in use and only if found adds the dependency.

timtebeek (approved):
Ended up removing the declarative `AddDependency` that was there before, and implementing a scanning recipe that's a little more selective. Thanks again for calling out this issue with over eager dependency addition!

FieteO:
Great, that there indeed was something to improve here. Thanks @timtebeek for working on this.

PR #582, opened 2024-08-15
Convert lambda block to expression after adopting `assertThrows`

shivanisky:
Test case for inlining valid single line assertthrows

## What's changed?
Inline valid single line assert throws test case

## What's your motivation?
This recipe creates intellij suggestion/warning to inline this code and I much prefer it inlined, saves 2 lines of useless code.

@timtebeek 

Yes - maybe a separate recipe which achieves this is better, or maybe both - change this recipe, and have a separate recipe. Can do separate recipe first as it addresses at a global level.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.java
@@ -114,7 +114,7 @@ class UpdateTestAnnotationTest implements RewriteTest {
             """
               import org.junit.Test;
 
-              public class MyTest {
+              class MyTest {
 
                   @Test(expected = IllegalArgumentException.class)
                   public void test() {
@@ -130,7 +130,7 @@ class UpdateTestAnnotationTest implements RewriteTest {
 
               import static org.junit.jupiter.api.Assertions.assertThrows;
 
-              public class MyTest {
+              class MyTest {
 
                   @Test
                   public void test() {
@@ -153,9 +153,9 @@ class UpdateTestAnnotationTest implements RewriteTest {
           java(
             """
               import org.junit.Test;
-                                
+
               public class MyTest {
-                                
+
                   @Test(expected = IndexOutOfBoundsException.class)
                   public void test() {
                       int arr = new int[]{}[0];
@@ -601,7 +601,7 @@ class UpdateTestAnnotationTest implements RewriteTest {
                       // Second call shows why we wrap the entire method body in the lambda
                       foo();
                   }
-                  
+
                   void foo() throws IOException {
                       throw new IOException();
                   }
@@ -624,7 +624,7 @@ class UpdateTestAnnotationTest implements RewriteTest {
                           foo();
                       });
                   }
-                  
+
                   void foo() throws IOException {
                       throw new IOException();
                   }
--- a/src/test/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.java
@@ -106,6 +106,45 @@ class UpdateTestAnnotationTest implements RewriteTest {
         );
     }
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.java
```diff
@@ -106,23 +106,62 @@ public void test() {
         );
     }
 
+    @Test
```
shivanisky, line 109: In test case above, can't inline even though single line ... and need to leave like that. So have to consider these situations.

timtebeek:
Thanks for yet another case to cover; think indeed best here to have a separate recipe, as that then keeps the logic cleaner here, and allows us to pick up cases already converted.

shivanisky:
Should the recipe be in rewrite static analysis? Maybe not just for
assertThrows, but for everything?

I think it’s just the removing parantheses bit … I will exclude throws
statements. Any other single statement cases which could break will be very
rare. I hope that this approach is fine.

Sorry to bother you when you are so busy. No rush on this one.

timtebeek:
I think we already have this one; we just need to tie it in here. Nice to be able to leverage existing logic there! :)

https://github.com/openrewrite/rewrite-static-analysis/blob/f1e53d203de026938615fee8b5175cba379dc279/src/test/java/org/openrewrite/staticanalysis/LambdaBlockToExpressionTest.java#L35-L58

timtebeek:
Did a quick check: we're running up against this safe guard where any overload invalidates the visit of a lambda argument.
https://github.com/openrewrite/rewrite-static-analysis/blob/6d04813574c6b02c8bf7d6e2f66b2bdaa881854c/src/main/java/org/openrewrite/staticanalysis/LambdaBlockToExpression.java#L74

Not quite sure if we can safely change that here. 🤔

timtebeek:
Specifically, the methods were seeing overloaded at that particular point are these.
```
org.junit.jupiter.api.Assertions{
    name=assertThrows,
    return=Generic{T extends java.lang.Throwable},
    parameters=[java.lang.Class<Generic{T extends java.lang.Throwable}>,org.junit.jupiter.api.function.Executable,java.util.function.Supplier<java.lang.String>]
}
org.junit.jupiter.api.Assertions{
    name=assertThrows,
    return=Generic{T extends java.lang.Throwable},
    parameters=[java.lang.Class<Generic{T extends java.lang.Throwable}>,org.junit.jupiter.api.function.Executable,java.lang.String]
}
org.junit.jupiter.api.Assertions{
    name=assertThrows,
    return=Generic{T extends java.lang.Throwable},
    parameters=[java.lang.Class<Generic{T extends java.lang.Throwable}>,org.junit.jupiter.api.function.Executable]
}
```
We could try this out in `LambdaBlockToExpressionTest` to  see if we can safely recognize these cases and allow the replacement, but it'll be tricky not the break any other cases there. I'll leave it at this for now. :)

timtebeek:
If you want here's a test for LambdaBlockToExpression in rewrite-static-analysis
```java
@Test
@Issue("https://github.com/openrewrite/rewrite-testing-frameworks/pull/582")
void simplifyAssertThrows() {
    rewriteRun(
      spec-> spec.recipe(new LambdaBlockToExpression()).parser(JavaParser.fromJavaVersion().classpath("junit").logCompilationWarningsAndErrors(true)),
      //language=java
      java(
        """
        import static org.junit.jupiter.api.Assertions.assertThrows;
        
        class Test {
            void test() {
                assertThrows(IllegalArgumentException.class, () -> {
                    foo();
                });
            }
            void foo() {
                throw new IllegalArgumentException("boom");
            }
        }
        """,
        """
        import static org.junit.jupiter.api.Assertions.assertThrows;
        
        class Test {
            void test() {
                assertThrows(IllegalArgumentException.class, () -> foo());
            }
            void foo() {
                throw new IllegalArgumentException("boom");
            }
        }
        """
      )
    );
}
```

Issue #572, opened 2023-12-06
`ExpectedExceptionToAssertThrows` leads to compilation issue when variables from lambda are referenced after `assertThrows`


TomChatt2:
## What version of OpenRewrite are you using?
I am using openrewrite-maven-plugin 5.11.0 with rewrite-spring 5.1.1

## How are you running OpenRewrite?
I am using the Maven plugin. Apologies, my project is on an offline network with no Internet access, so I'm not able to share or copy anything. 

## This is the gist of the initial code
```java
import my.MyAppException;
import org.junit.Rule;
import org.junit.Test;
import org.junit.rules.ExpectedException;

public class ApplicationTest {

  @Rule
  public ExpectedException expectedException = ExpectedException.none();

  @Test
  public void sampleTest() throws Exception {
    final String EXPECTED_MSG = "this is what i expected";
    expectedException.expect(MyAppException.class);
    expectedException.expectMessage(EXPECTED_MSG);
    doThrow(new MyAppException(EXPECTED_MSG)).when(...)...
    // some app code expected to throw exception here
  }
}
```

## What did you expect to see?
```java
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertThrows;

import my.MyAppException;
import org.junit.jupiter.api.Test;

class ApplicationTest {

  @Test
  void sampleTest() throws Exception {
    final String EXPECTED_MSG = "this is what i expected";
    ThrowableException ex = assertThrows(MyAppException.class, () -> {
        doThrow(new MyAppException(EXPECTED_MSG)).when(...)...
        // some app code expected to throw exception here
    });
    assertTrue(ex.getMessage().contains(EXPECTED_MSG));
  }
}
```

## What did you see instead?
```java
import static org.junit.jupiter.api.Assertions.assertEquals;
// the import for assertThrows was missing

import my.MyAppException;
import org.junit.jupiter.api.Test;

class ApplicationTest {

  @Test
  void sampleTest() throws Exception {
    ThrowableException ex = assertThrows(MyAppException.class, () -> {  // ERROR: assertThrows is undefined (missing import)
        final String EXPECTED_MSG = "this is what i expected";  // this shouldn't have been put inside the lambda expression
        doThrow(new MyAppException(EXPECTED_MSG)).when(...)...
        // some app code expected to throw exception here
    });
[...]

Issue #194, opened 2022-03-26
JUnit 5 Assertions.assertThrows should only contain a single statement


timtebeek:
Hi! Noticed a somewhat surprising pattern after migrating from JUnit 4 to JUnit 5, which is [captured in the unit test `assertThrowsMultiLine`](https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/test/kotlin/org/openrewrite/java/testing/junit5/UpdateTestAnnotationTest.kt#L126):
```java
@Test
fun assertThrowsMultiLine() = assertChanged(
    before = """
        import org.junit.Test;
        
        public class A {
        
            @Test(expected = IllegalArgumentException.class)
            public void test() {
                String foo = "foo";
                throw new IllegalArgumentException("boom");
            }
        }
    """,
    after = """
        import org.junit.jupiter.api.Test;
        
        import static org.junit.jupiter.api.Assertions.assertThrows;
        
        public class A {
        
            @Test
            void test() {
                assertThrows(IllegalArgumentException.class, () -> {
                    String foo = "foo";
                    throw new IllegalArgumentException("boom");
                });
            }
        }
    """
)
```

While technically equivalent to the JUnit 4 model, where any of the two statements might throw an exception caught in the assertion, it's nonetheless perhaps too broad, particularly when there are more than a handful of statements, all of which might throw an exception. Ideally, in my opinion, one would only assert that the last statement is used within the `assertThrows` invocation.

```java
@Test
void test() {
    String foo = "foo";
    assertThrows(IllegalArgumentException.class, () -> throw new IllegalArgumentException("boom"));
}
```

timtebeek:
Generalized, and separated from the JUnit 4 to 5 migration, this could even be a recipe in and of itself:
```java
package org.openrewrite.java.testing.junit5

import org.junit.jupiter.api.Test
import org.openrewrite.java.JavaRecipeTest

class AssertThrowsOnSingleStatementTest : JavaRecipeTest {

    override val recipe = AssertThrowsOnSingleStatement()

    @Test
    fun assertThrowsOnSingleStatement() = assertChanged(
        before = """
            package org.openrewrite.java.testing.junit5;
            
            import static org.junit.jupiter.api.Assertions.assertThrows;
            
            public class SimpleAssertThrowsTest {
            
                public void throwsExceptionWithSpecificType() {
                    assertThrows(NullPointerException.class, () -> {
                        Object foo = null;
[...]

PR #583, opened 2024-08-16
Prevent `ExpectedExceptionToAssertThrows` from pulling variable declarations into lambda

Laurens-W:
## What's changed?

Added a testcase to prove them problem exists

@timtebeek 

## Any additional context
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/572
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/194

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrowsTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrowsTest.java
@@ -469,4 +469,51 @@ class ExpectedExceptionToAssertThrowsTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void variableDeclarationsShouldNotBePulledIntoAssertThrowsLambda() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.junit.Rule;
+              import org.junit.Test;
+              import org.junit.rules.ExpectedException;
+
+              class MyTest {
+
+                  @Rule
+                  public ExpectedException thrown = ExpectedException.none();
+
+                  @Test
+                  public void testMethod() {
+                      int num = 1;
+                      String message = "message";
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrowsTest.java
```diff
@@ -471,4 +471,60 @@ void foo() throws IOException {
...
+
+              class MyTest {
+
+                  @Test
+                  public void testMethod() {
+                      int num = 1;
+                      String message = "message";
+                      Throwable exception = assertThrows(RuntimeException.class, () -> {
+                          foo();
+                      }, "Using vars" + num + message);
```
timtebeek, line 519: This style of assertion is new, and a likely better replacement for the default case as well; that up to now had used:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/da544f83adab0975be2d4cd841fb1a51d42306c8/src/test/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrowsTest.java#L109-L111

Whereas if we can prevent the `Throwable exception` assignment that would be even better.

timtebeek: We do have a potential other issue that would tie in with this one, without an immediately desirable solution just yet:
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/194#issuecomment-1082191671

timtebeek: This was wrong; the third argument is the failure message for the test, not the expected exception message.

timtebeek:
Closed after discussion, with summary posted on https://github.com/moderneinc/customer-requests/issues/285#issuecomment-2298474812

PR #584, opened 2024-08-20
Retain Mockito Strictness WARN when switching to JUnit 5 through `@MockitoSettings(strictness = Strictness.WARN)`

Laurens-W:
## What's changed?

Add a recipe to add `@MockitoSettings(strictness = Strictness.LENIENT)`  so that the default behavior does not change

@timtebeek 

### Checklist


src/main/resources/META-INF/rewrite/mockito.yml
```diff
@@ -147,3 +147,4 @@ recipeList:
       oldGroupId: org.mockito
       oldArtifactId: mockito-all
       newArtifactId: mockito-core
+  - org.openrewrite.java.testing.junit5.MockitoJUnitAddMockitoSettingsLenientStrictness
```
timtebeek, line 150: As indicated, we'd only want to run this recipe if folks are in the process of moving from Mockito 1 or 2 on JUnit 4 to JUnit 5, as that's where the default looks to have changed [according to this answer](https://stackoverflow.com/a/53234137/53444). In general the docs are still a bit vague as to strictness, as seen from the many issues linked here:
- https://github.com/mockito/mockito/issues/2532

Here's an example PR that adds conditional logic for running a recipe only when still on an older version of a dependency; that should help restrict this run if we stick to a detached recipe.
- https://github.com/openrewrite/rewrite-spring/pull/555

src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitAddMockitoSettingsLenientStrictness.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+
+public class MockitoJUnitAddMockitoSettingsLenientStrictness extends Recipe {
```
timtebeek, line 26: As indicated I wonder if we should make this [part of `MockitoJUnitToMockitoExtension`](https://github.com/openrewrite/rewrite-testing-frameworks/blob/efcc1e4904e90d0c7a1cbd6b1a230ff698a3917d/src/main/java/org/openrewrite/java/testing/junit5/MockitoJUnitToMockitoExtension.java#L53) instead of having a separate recipe. That'll depend on whether Mockito 1 and 2 could be used on JUnit 4 without the use of that test MockitoRule or MockitoTestRule.

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitAddMockitoSettingsLenientStrictness.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2021 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitAddMockitoSettingsLenientStrictness.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final String EXTEND_WITH_MOCKITO_EXTENSION = "@" + EXTEND_WITH_FQ + "(" + MOCKITO_EXTENSION_FQ + ".class)";
+
+    @Override
+    public String getDisplayName() {
+        return "Add `@MockitoSettings(strictness = Strictness.LENIENT)` when migration to JUnit 5";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Add `@MockitoSettings(strictness = Strictness.LENIENT)` when migration to JUnit 5.";
```
timtebeek, line 49: It'd be good to describe _why_ we're making this change here, as it's likely to raise a few eye brows.

PR #588, opened 2024-08-28
`PowerMockitoMockStaticToMockito` duplicates variables and methods when static nested class is involved

Laurens-W:
## What's changed?

Add minimalistic testcase to show that when a static nested class is mocked, it tries to add a field declaration but the JavaTemplate breaks duplicating fields and methods.

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoMockStaticToMockito.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoMockStaticToMockito.java
@@ -446,7 +446,6 @@ public class PowerMockitoMockStaticToMockito extends Recipe {
                                 new Cursor(getCursor().getParentOrThrow(), classDecl),
                                 classDecl.getBody().getCoordinates().firstStatement(),
                                 classlessTypeName,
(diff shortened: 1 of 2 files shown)


timtebeek:
Great to see the troublesome case replicated in a small way; would be good to see this resolved as well.

timtebeek (commented):
Great to see this replicated; would be good to see this resolved as well.

src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoMockStaticToMockito.java
```diff
@@ -446,7 +446,8 @@ private J.ClassDeclaration addFieldDeclarationForMockedTypes(J.ClassDeclaration
                                 new Cursor(getCursor().getParentOrThrow(), classDecl),
                                 classDecl.getBody().getCoordinates().firstStatement(),
                                 classlessTypeName,
-                                classlessTypeName
+                                // Not sure if replacing `.` with `_` is a structural solution, but it works for inner classes for now
```
timtebeek, line 449: This seems more like something to post as a self-review comment as opposed to a comment to carry forward in the code.
```suggestion
```

PR #590, opened 2024-08-30
Move Junit 5 assertThrows to last statement in lamdba

shivanisky:
## What's changed?
In multi line assertThrows move assertThrows to last statement only

## What's your motivation?
Currently the Junit 4 to 5 recipe migrates the @Test junit 4 annotation to wrap the entire method to assertThrows when typically we want it on the last statement.
 
@timtebeek 

Yes - may be modifying the existing recipe is better? Rather than providing an extra recipe to move existing assertThrows with multi line lamdbas to the last statement only, but we can hook this recipe in optionally, so ... maybe not. Maybe that can be an optional argument in the existing recipe - move applyAssertThrowsToLast and the user can set it to true if they are ok to take a hit of compilation errors in favour of better usage of junit 5 features and correctness.

## Any additional context
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/194

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
@@ -43,8 +43,8 @@ public class AssertThrowsOnLastStatement extends Recipe {
 
     @Override
     public String getDescription() {
-        return "Applies Junit 5 assertThrows on last statement in lambda block only, in rare cases may cause compilation " +
-                "errors if lambda uses non final variables.";
+        return "Applies Junit 5 assertThrows on last statement in lambda block only, in extremely rare cases may cause " +
+                "compilation errors if lambda uses non final variables.";
     }
 
     @Override
--- a/src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
@@ -100,6 +100,8 @@ public class AssertThrowsOnLastStatement extends Recipe {
                         return methodStatement;
                     }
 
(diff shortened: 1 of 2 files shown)


timtebeek:
Hi @shivanisky ; Great start! I've pushed a few quick changes to get you on a path to get this mostly working; then from here we'd have to guard against the cases where we might do harm, which would be the more interesting challenge of this recipe. There's a few such cases described in https://github.com/openrewrite/rewrite-testing-frameworks/issues/194 that we might want to cover here. Thanks again for exploring if we an actually see this through in a mostly reliable way.

shivanisky:
> Hi @shivanisky ; Great start! I've pushed a few quick changes to get you on a path to get this mostly working; then from here we'd have to guard against the cases where we might do harm, which would be the more interesting challenge of this recipe. There's a few such cases described in #194 that we might want to cover here. Thanks again for exploring if we an actually see this through in a mostly reliable way.

Thank you so much for your optimisations and feature adds! Its looking very close now :)

Regarding the final/non final variable maybe can find all variables used in last statement, find the declaration and see if declared as final. Maybe the LST holds this information - sometimes this is what dreams are made of hehe :)

shivanisky:
Added case for variable declaration ie Throwable ex = assertThrows(....) seems to be working well. However, LamdaBlockToExpression doesn't seem to be removing curly braces for single statement lamdba and inlining it. Any idea why? This is generated intellij warning for the test java file.

shivanisky:
I think no point in putting a lot of effort for non final variables case - I am not seeing it across multiple repos with thousands of test cases and over 500 files. Have put in warning in description regarding this rare case, and I think better that people can use it if they want to. Eventually, can hook in to the main junit 5 migration recipe with argument to optionally use this, so that scope is limited to migrating the junit 4 @Test annotation only rather than touching existing assertThrows lamdas.

PR #592, opened 2024-09-05
Do not perform `JUnit4to5Migration` where TestNG is used

Laurens-W:
## What's changed?

Precondition added to `JUnit4to5Migration` that checks whether testNG is used

@timtebeek 

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/junit5.yml
+++ b/src/main/resources/META-INF/rewrite/junit5.yml
@@ -51,6 +51,9 @@ description: Migrates JUnit 4.x tests to JUnit Jupiter.
 tags:
   - testing
   - junit
+preconditions:
+  - org.openrewrite.java.search.DoesNotUseType:
+      fullyQualifiedTypeName: org.testng..*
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -51,6 +51,9 @@ description: Migrates JUnit 4.x tests to JUnit Jupiter.
 tags:
   - testing
   - junit
+preconditions:
+  - org.openrewrite.java.search.DoesNotUseType:
+      fullyQualifiedTypeName: org.testng..*
```
timtebeek, line 56: Did you consider adding `includeImplicit` ? Any pros/cons you see?
```suggestion
      fullyQualifiedTypeName: org.testng..*
      includeImplicit: true
```

Laurens-W: I thought it might not be necessary, but after running a findtype with both options it seems to almost always result in more finding!

PR #596, opened 2024-09-10
Handle multiple blocks in JMockit Expectations

yurii-yu:
## What's changed?

Cope with a corner case in refactoring JMockit Expectations.
Sometimes programmers will use more than one curly braces in one Expectation block. This looks bizarre and confusing, but it is syntactically correct. It will be ideal if such code can also be refactored automatically.

usually people write code like this:
```java
new Expectations() {{
    MethodComparatorTest.this.method.getName(); result = "NoTest";
    MethodComparatorTest.this.method2.getName(); result = "Test";
}};
```
while some programmers prefer grouping related statements with redundant curly braces
```java
new Expectations() {
    {
        MethodComparatorTest.this.method.getName();
        this.result = "NoTest";
    }
    {
        MethodComparatorTest.this.method2.getName();
        this.result = "Test";
    }
};
```
I made changes in two places.

On `line 39` in `JMockitUtils.java`, this criterion need to be changed, or else OpenRewrite will just ignore such blocks. 
```java
if (nc.getBody() == null || nc.getBody().getStatements().size() != 1)
```

On `line 53` in `SetupStatementsRewriter.java`, this rewrite logic need to be able to loop over all statements, instead of only the first one,  in the Expectations block.
```java
J.Block expectationsBlock = (J.Block) nc.getBody().getStatements().get(0);

// statement needs to be moved directly before expectations class instantiation
JavaCoordinates coordinates = nc.getCoordinates().before();
List<Statement> newExpectationsBlockStatements = new ArrayList<>();
for (Statement expectationStatement : expectationsBlock.getStatements()) {
    if (!isSetupStatement(expectationStatement, spies)) {
        newExpectationsBlockStatements.add(expectationStatement);
        continue;
    }
    rewriteBodyStatement(expectationStatement, coordinates);
    // subsequent setup statements are moved in order
    coordinates = expectationStatement.getCoordinates().after();
}
// the new expectations block has the setup statements removed
J.Block newExpectationsBlock = expectationsBlock.withStatements(newExpectationsBlockStatements);
nc = nc.withBody(nc.getBody().withStatements(Collections.singletonList(newExpectationsBlock)));
```

I also added one test case in `JMockitExpectationsToMockitoTest.java`

## What's your motivation?

We encountered problem in dealing with several legacy projects. 
Unfortunately, OpenWrite just kept most of those test cases untouched. It took us a couple of days to locate the problem.
[...]

PR #603, opened 2024-09-16
Bugfix/multiple curly braces in one verification block

yurii-yu:
## What's changed?

This is a change similar to [the one I made before: Pull Request 596](https://github.com/openrewrite/rewrite-testing-frameworks/pull/596)

Cope with a corner case in refactoring JMockit Verifications.
Sometimes programmers will use more than one curly braces in one Verification block. 
This is bizarre and confusing, because it looks like there are multiple "anonymous class" here while in fact these curly braces do not make any effective difference. It will be ideal if such code can also be refactored automatically.

Usually people write code like this:

```java
new Verifications() {{
    myObject.wait();
    myObject.wait(anyLong, anyInt);
    myObject.wait(anyLong);
    times = 2;
}};
```

while some programmers prefer grouping related statements with redundant curly braces

```java
new Verifications() {
    {
    myObject.wait();
    myObject.wait(anyLong, anyInt);
    }
    {
    myObject.wait(anyLong);
    times = 2;
    }
};
```

Such code snippet can not be refactored automatically.

## What's your motivation?
We encountered problem in dealing with several legacy projects.
Some codes can not be refactored automatically, while manually changing them seems tedious, since they are not complex at all. We located and solved this problem, which can save us quite a lot of time.
Benefited a lot from the OpenRewrite project previously, we believe that more people and more legacy projects should be able to take advantage of it.

This change is similar to the one I made for "multiple blocks in one Expectations block", I only modified the IF statement to cope with not only Expectation block but also Verifications block. Moreover, I adhere to the convention of the changes made by @timtebeek for my last pull request.
I read the best practice so I kept the LST untouched.
The change has been tested by the entire test suite. As far as I can see, everything works fine.
Please note that the new code might need to be formatted to comply with the existing convention.

@timtebeek I would be very happy if you can review this, it is an easy fix.

Maybe changing the LST is a better solution, but I do not think it is worthy since this is not a common case.
Users can also use a script to remove the curly braces before applying this recipe, but that is an invasive way.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
@@ -36,10 +36,10 @@ class JMockitUtils {
         J.Identifier clazz = (J.Identifier) nc.getClazz();
 
         // JMockit block should be composed of a block within another block
-        if (nc.getBody() == null
-                || (nc.getBody().getStatements().size() != 1 &&
-                        !TypeUtils.isAssignableTo("mockit.Expectations", clazz.getType()) &&
-                        !TypeUtils.isAssignableTo("mockit.Verifications", clazz.getType()))) {
+        if (nc.getBody() == null ||
+            (nc.getBody().getStatements().size() != 1 &&
+             !TypeUtils.isAssignableTo("mockit.Expectations", clazz.getType()) &&
+             !TypeUtils.isAssignableTo("mockit.Verifications", clazz.getType()))) {
             return empty();
         }
 
--- a/src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/JMockitUtils.java
@@ -36,8 +36,10 @@ class JMockitUtils {
         J.Identifier clazz = (J.Identifier) nc.getClazz();
 
         // JMockit block should be composed of a block within another block
-        if (nc.getBody() == null || (nc.getBody().getStatements().size() != 1 &&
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Great to see this reuse; thanks for the fix again!

yurii-yu:
@timtebeek It's so nice working with you, since the experience is wonderful.

To me, there are still some mysterious problems when applying OpenRewrite to our legacy project. For example, in the verification block, `anyLong` can be refactored to `anyLong()` correctly, whereas `this.anyLong` (`anyLong` is a member of the `Verifications` class)can not, because the former is granted as a `J$Identifier` while the latter is granted as a `J$FieldAccess`.
We used a bash script to remove all `this.` which cause problems. But I guess it would be ideal if other users can apply OpenRewrite in such scenario without any extra work.
I am still looking into this problem but I can not say how long it takes. I will come back to you if I (hopefully) solve this problem.
Regards

timtebeek:
Thanks for the kind words and continued improvements @yurii-yu ! It's great to see how quickly you've taken to adding improvements here.

For the `J$Identifier` vs `J$FieldAccess` it might be best to start a draft PR with a unit test, and then step through with the debugger & [TreeVisitingPrinter](https://docs.openrewrite.org/concepts-explanations/tree-visiting-printer#using-treevisitingprinter.printtree) to inspect and handle each individual case. I hope that helps!

yurii-yu:
Yes, thank you for this insightful suggestion. That's exactly what I am thinking about.
I guess the problem lies not in rewriting-testing-framework but in the rewrite project. I have some initial clue, but I still need more time since the issue is much complicated than I thought before because there are numerous scenarios concerning `this.`, I must be cautious. And, after all, this is a "non-official project" for me. 
I swear "I will be back". You have my words :)

PR #604, opened 2024-09-21
AssertJ assertions not collapsed on Integer

timo-abele:
## What's your motivation?

I expected the recipe to collapse assertThats on any type, but in my project it didn't work on an `Integer`. 
This PR verifies my observations and serves as a starting point for 
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/605

### Checklist


timtebeek:
Hi @timo-abele ! Thanks for notifying us with a runnable test. :) It looks to fail on this particular line:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/3412efbd62dd4d155dd633b1674e54a1cdd023db/src/main/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatements.java#L104

method | type
-- | --
`assertThat.getType()` | `org.assertj.core.api.AbstractIntegerAssert<Generic{?}>`
`assertion.getType()` | `Generic{? extends org.assertj.core.api.AbstractIntegerAssert<Generic{? extends }>}`

src/test/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatementsTest.java
```diff
@@ -191,6 +191,42 @@ private int[] notification() {
...
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import static org.assertj.core.api.Assertions.assertThat;
+
+              import org.junit.jupiter.api.Test;
+
+              class MyTest {
+                  @Test
```
timtebeek, line 205:
```suggestion
              class MyTest {
```

src/test/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatementsTest.java
```diff
@@ -191,6 +191,42 @@ private int[] notification() {
...
+                  }
+              }
+              """,
+            """
+              import static org.assertj.core.api.Assertions.assertThat;
+
+              import org.junit.jupiter.api.Test;
+
+              class MyTest {
+                  @Test
```
timtebeek, line 219:
```suggestion
              class MyTest {
```

src/test/java/org/openrewrite/java/testing/assertj/CollapseConsecutiveAssertThatStatementsTest.java
```diff
@@ -191,6 +191,42 @@ private int[] notification() {
...
+                  void test() {
+                      Integer i = 1+1;
+                      assertThat(i).isNotNull().isEqualTo(2);
```
timtebeek, line 222:
```suggestion
                      assertThat(i)
                              .isNotNull()
                              .isEqualTo(2);
```

timtebeek (commented):
Some minor suggestions to get the test going: there's no need to add the JUnit dependency here. :)

timtebeek:
You'll notice that the types are different; we have that prevent methods like `extracting` from changing the type, as those can't be safely chained.

We can likely relax the type comparison a little though, to allow for parameterized types with a bound that includes the same class, even if the generic bound is slightly different (`Generic{?}` vs `Generic{? extends }`)

timtebeek:
Thanks for getting this started! We can keep this open until the moment we figure out the exact change needed to get this going, as indeed we're a little too conservative right now, not chaining all the cases we could potentially.

timtebeek:
Explored this briefly again, but couldn't yet find an easy solution with the generics mismatch between the two method invocations. We'll merge this with the test disabled such that it's easier to pick up #605 in the future.

timtebeek (approved):

PR #606, opened 2024-09-21
`assertInstanceOf` not converted from JUnit5 to assertJ

timo-abele:
## What's your motivation?

Reproducing an observed conversion miss.

## Any additional context

I'm not sure with sub recipe should convert this so I am using the best practices recipe as a catch-all.

### Checklist


timtebeek:
Thanks for the suggestion! This likely needs a dedicated recipe, similar to the other examples seen here:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/8fb11c8f0f116a87883fc30008b008eeba6568cc/src/main/resources/META-INF/rewrite/assertj.yml#L459-L468

Anyone welcome to fill in the missing piece here to make it a reality.

src/test/java/org/openrewrite/java/testing/assertj/AssertJBestPracticesTest.java
```diff
@@ -38,10 +38,36 @@ class AssertJBestPracticesTest implements RewriteTest {
     public void defaults(RecipeSpec spec) {
         spec.parser(
             JavaParser.fromJavaVersion()
-              .classpathFromResources(new InMemoryExecutionContext(), "assertj-core-3.24"))
+              .classpathFromResources(new InMemoryExecutionContext(), "assertj-core-3.24", "junit-jupiter-api-5.9"))
```
timtebeek, line 41:
```suggestion
              .classpathFromResources(new InMemoryExecutionContext(), "assertj-core-3.24"))
```

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright 2021 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
...
+
+            List<Expression> args = method.getArguments();
+
+            if (args.size() == 2) {
+                Expression expectedType = args.get(0);
+                Expression actualValue = args.get(1);
+
+                method = JavaTemplate.builder("assertThat(#{any()}).isInstanceOf(#{any()});")
+                        .staticImports("org.assertj.core.api.Assertions.assertThat")
+                        .javaParser(assertionsParser(ctx))
```
timtebeek, line 78: This should just be this simple; drop the `assertionsParser` field and method; those stemmed from an earlier time when we did not have caching on the JavaParsers created.
```suggestion
                        .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "assertj-core-3.24"))
```

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
...
+        private JavaParser.@Nullable Builder<?, ?> assertionsParser;
+
+        private JavaParser.Builder<?, ?> assertionsParser(ExecutionContext ctx) {
+            if (assertionsParser == null) {
+                assertionsParser = JavaParser.fromJavaVersion()
+                        .classpathFromResources(ctx, "assertj-core-3.24");
+            }
+            return assertionsParser;
+        }
+
```
timtebeek, line 61:
```suggestion
```

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
...
+        return "Convert JUnit-style `assertInstanceOf()` to AssertJ's `assertThat().isInstanceOf()`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("org.junit.jupiter.api.Assertions", false),
+                new AssertInstanceOfToToAssertThatVisitor());
+    }
+
+    public static class AssertInstanceOfToToAssertThatVisitor extends JavaIsoVisitor<ExecutionContext> {
```
timtebeek, line 51: There is no need for this to be a public static class; just create an inline anonymous class instead.

src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThatTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec
+          .parser(JavaParser.fromJavaVersion()
+            .classpathFromResources(new InMemoryExecutionContext(), "junit-jupiter-api-5.9"))
+          .recipe(new JUnitAssertInstanceOfToAssertThat());
+    }
+
+    @Test
```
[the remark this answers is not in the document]
timtebeek: We try to have exactly one example unit test tagged with this annotation at the top of every unit test file. That way it's clear which one to copy when you want to create a variant.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
...
+
+        private JavaParser.Builder<?, ?> assertionsParser(ExecutionContext ctx) {
+            if (assertionsParser == null) {
+                assertionsParser = JavaParser.fromJavaVersion()
+                        .classpathFromResources(ctx, "assertj-core-3.24");
+            }
+            return assertionsParser;
+        }
+
+        private static final MethodMatcher JUNIT_ASSERT_NULL_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions" + " assertInstanceOf(..)");
```
timtebeek, line 62:
```suggestion
        private static final MethodMatcher JUNIT_ASSERT_NULL_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions assertInstanceOf(..)");
```

jevanlingen (commented):
Apply bot comments

jevanlingen (approved):
Apply bot comments

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,105 @@
...
+        return "Convert JUnit-style `assertInstanceOf()` to AssertJ's `assertThat().isInstanceOf()`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("org.junit.jupiter.api.Assertions", false),
+                new AssertInstanceOfToToAssertThatVisitor());
+    }
+
+    private static class AssertInstanceOfToToAssertThatVisitor extends JavaIsoVisitor<ExecutionContext> {
```
timtebeek, line 52: I'd intended my earlier comment as doing this instead; having nested named visitors is again an outdated pattern.
```suggestion
                new JavaIsoVisitor<ExecutionContext>() {
```

timtebeek: You'll want a matching change further down to add the missing `);}`

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Convert JUnit-style `assertInstanceOf()` to AssertJ's `assertThat().isInstanceOf()`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("org.junit.jupiter.api.Assertions", false),
```
timtebeek, line 47: Can we change the precondition to `UsesMethod` instead? That can reuse a `MethodMatcher` variable. That make the precondition more specific, and then runs the recipe visitor on fewer classes for a performance improvement.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -0,0 +1,94 @@
...
+                new JavaIsoVisitor<ExecutionContext>() {
+                    private final MethodMatcher JUNIT_ASSERT_NULL_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions assertInstanceOf(..)");
+
+                    @Override
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
```
timtebeek, line 52: We lack a call to `super.visitMethodInvocation` here, and assign that to a variable `md` you use further down. Otherwise you miss out on assertions that might be nested inside other method invocations, as you could have with assertAll, assertSoftly, and other method invocations that could take a lambda.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
jevanlingen: Yups, good idea!

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
jevanlingen: Yups, good idea!

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
jevanlingen: Yeah, you are right. Actually I thought about it, but then I went in line with the other AssertJ recipe's (which don't have the super invocation either). So let's do a follow up PR where we improve the other recipe's as well.

PR #610, opened 2024-09-25
Migrate JMockit block to Mockito which contains redundant "this" prefix for arg matcher and migrate array as parameter for any 

yurii-yu:
## What's changed?

Be able to cope with corner cases in refactoring JMockit Verifications.

1. Normally, `anyString` in JMockit should be refactored to `anyString()` in Mockito, so as to `anyInt` or `anyLong`. But when there are redundant `this` before `anyString`, the code will not be processed because OpenWrite treat `anyString` as an `J.Identifier` object where as `this.anyString` as a `J.FieldAccess` object. My patch will do a type cast for this exact case.
2. Normally, `(Object) any` in Mockito should be refactored to `any(Object.class)` in Mockito. This works for simple Object. But when `Object` is an array such as `(byte[]) any`, OpenRewrite will not change it to `any(byte[].class)` because the class type of `byte[]` is not `JavaType.FullyQualified`. I added processing branch for this case.

Corresponding unit tests were also added.

## What's your motivation?

We encountered problem in dealing with several legacy projects.
Some codes can not be refactored automatically, while manually changing them seems tedious, since they are not complex at all. We located and solved this problem, which can save us quite a lot of time.
Benefited a lot from the OpenRewrite project previously, we believe that more people and more legacy projects should be able to take advantage of it.

My solution for the second question is obviously not the perfect one, because the element type of array could be another array so theoretically  this could be a infinite loop. Solving such problem will take me much more time and energy. 
But I guess that is a rare case which is almost impossible to encounter, so my code simply assume that the element type of the array is plain object.

@timtebeek Please have a look again on my code, amigo :)

Yes, as I mentioned above, I think a perfect solution for the second question is unworthy.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/jmockit/ArgumentMatchersRewriter.java
+++ b/src/main/java/org/openrewrite/java/testing/jmockit/ArgumentMatchersRewriter.java
@@ -154,9 +154,9 @@ class ArgumentMatchersRewriter {
             // ((int) any) to anyInt(), ((long) any) to anyLong(), etc
             argumentMatcher = PRIMITIVE_TO_MOCKITO_ARGUMENT_MATCHER.get(type);
             template = argumentMatcher + "()";
-        } else if (type instanceof JavaType.FullyQualified) {
-            // ((<type>) any) to any(<type>.class)
-            return rewriteFullyQualifiedToArgumentMatcher(methodArgument, (JavaType.FullyQualified) type);
+        } else if (type instanceof JavaType.FullyQualified || type instanceof JavaType.Array) {
+            // ((<type>) any) to any(<type>.class), type can also be simple array
+            return rewriteAnyWithClassParameterToArgumentMatcher(methodArgument, type);
         }
         if (template == null || argumentMatcher == null) {
             // unhandled type, return argument unchanged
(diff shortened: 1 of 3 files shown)


timtebeek:
Many thanks for the continued improvements here! I'm traveling right now so it's a bit harder to do a detailed review. Perhaps @shivanisky can chime in? (No requirement of course!)
Otherwise I should be back Monday, for one day only, and can try to fit it in then.

shivanisky:
> Many thanks for the continued improvements here! I'm traveling right now so it's a bit harder to do a detailed review. Perhaps @shivanisky can chime in? (No requirement of course!)
> 
> Otherwise I should be back Monday, for one day only, and can try to fit it in then.

Weekend is looking busy but I'll try to look if I get free. Glad to see this work is helping ☺️ others.

These code changes affect not just JMockit Verifications but Expectations and NonStrictExpectations migration as well, essentially the entire JMockit to Mockito migration.

src/test/java/org/openrewrite/java/testing/jmockit/JMockitVerificationsToMockitoTest.java
```diff
@@ -777,4 +775,122 @@ void test() {
...
+                      verify(myObject).wait(anyLong(), anyInt());
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void whenArrayArgumentMatcher() {
```
shivanisky, line 824: As this change affect expectations as well, may be best to add a similar test case in JMockitExpectationsToMockitoTest. Instead of verify(object).method(any(byte[].class), would be when(object.method(any(byte[]).class).thenReturn ...

src/test/java/org/openrewrite/java/testing/jmockit/JMockitVerificationsToMockitoTest.java
```diff
@@ -777,4 +775,122 @@ void test() {
...
+              
+              import static org.mockito.Mockito.*;
+              
+              @ExtendWith(MockitoExtension.class)
+              class MyTest {
+                  @Mock
+                  Object myObject;
+              
+                  void test() {
+                      verify(myObject).wait(anyLong(), anyInt());
```
shivanisky, line 815: Looks good. Could you add another test case which contains a mix of this and no this for the same method ? eg myOject.wait(this.anyString, anyInt)

src/main/java/org/openrewrite/java/testing/jmockit/ArgumentMatchersRewriter.java
```diff
@@ -100,16 +101,24 @@ private J.MethodInvocation rewriteMethodInvocation(J.MethodInvocation invocation
             invocation = invocation.withSelect(rewriteMethodInvocation((J.MethodInvocation) invocation.getSelect()));
         }
         // in mockito, argument matchers must be used for all arguments or none
-        boolean hasArgumentMatcher = false;
-        List<Expression> arguments = invocation.getArguments();
-        for (Expression methodArgument : arguments) {
-            if (isJmockitArgumentMatcher(methodArgument)) {
-                hasArgumentMatcher = true;
-                break;
+        List<Expression> arguments = new ArrayList<>(invocation.getArguments().size());
```
shivanisky, line 104: Could simplify this to:
       List<Expression> arguments = invocation.getArguments();
       // replace this.matcher with matcher, otherwise it's ignored
       arguments.replaceAll(arg -> { 
               if (methodArgument instanceof J.FieldAccess) {
                    J.FieldAccess fieldAccess = (J.FieldAccess) methodArgument;
                     if (fieldAccess.getTarget() instanceof J.Identifier &&
                            ((J.Identifier) fieldAccess.getTarget()).getSimpleName().equals("this")) {
                               return fieldAccess.getName();
                    }
               }
               return arg;
      });

src/test/java/org/openrewrite/java/testing/jmockit/JMockitVerificationsToMockitoTest.java
yurii-yu: Insightful! Of course I take your suggestions

src/test/java/org/openrewrite/java/testing/jmockit/JMockitVerificationsToMockitoTest.java
yurii-yu: Also brilliant. I added a "mixture" and fortunately my code works well with it :)

src/main/java/org/openrewrite/java/testing/jmockit/ArgumentMatchersRewriter.java
yurii-yu: Personally I prefer the "functional programming" way, but I just followed the "traditional" ways in this file.
After a rough search I found that there are already plenty of code using functional programming, so I take your suggestion

yurii-yu:
@timtebeek Could you please kindly have a look at this pull-request? Thank you.

timtebeek (approved):
Thanks a lot for these fixes @yurii-yu ! Good to have those `this.` cases covered as well, and thanks for the help reviewing @shivanisky 🙏🏻

PR #617, opened 2024-10-14
Jmockit mockito full verifications and Refactor tests

shivanisky:
## What's changed?
Migrate JMockit FullVerifications to Mockito
https://www.javadoc.io/doc/org.jmockit/jmockit/latest/mockit/FullVerifications.html
https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html <- for verifyNoMoreInteractions

Replacing FullVerifications with standard mockito verify and then add one line of verifyNoMoreInteractions(mocks. ..)

Also do a refactor of tests to remove some unnecessary code - add a new JMockitTestBase which sets the default parser settings and other tests can inherit from that, hope it's ok to use inhertance here, if not please let me know. I think it makes a lot of sense here.

## What's your motivation?
Just wanted to add this feature to enhance the current migration scope, InOrderVerifications feature was requested here https://github.com/jmockit/jmockit1/issues/729#issuecomment-2406723779 and this will also help pave the way for that

@timtebeek @tinder-dthomson as FYI, and if wants to review also (I know he is busy)

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/jmockit/JMockitTestBase.java
+++ b/src/test/java/org/openrewrite/java/testing/jmockit/JMockitTestBase.java
@@ -1,0 +1,14 @@
+package org.openrewrite.java.testing.jmockit;
+
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.testing.jmockit.JMockitTestUtils.setDefaultParserSettings;
+
+public class JMockitTestBase implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        setDefaultParserSettings(spec);
+    }
+}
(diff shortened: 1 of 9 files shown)


src/test/java/org/openrewrite/java/testing/jmockit/JMockitTestBase.java
```diff
@@ -0,0 +1,29 @@
...
+
+import static org.openrewrite.java.testing.jmockit.JMockitTestUtils.setDefaultParserSettings;
+
+class JMockitTestBase implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        setDefaultParserSettings(spec);
+    }
+}
```
sambsnyd, line 29: I would prefer not to create a base class just to share a single method invocation's worth of configuration. Single inheritance of `JMockitTestBase` limits the ability of tests to extend other base classes for little code reuse. I would prefer to have `setDefaultParserSettings()` called explicitly in the `defaults` method of each individual test class.

sambsnyd (commented):
This is great!

PR #627, opened 2024-10-31
Initial support for DbRider migration from junit4 to junit5

### LINKED ISSUE #624: DBRider support
## What problem are you trying to solve?
Internally, we're using dbrider for integration test that assert db state. It's an extension on top of dbunit that allows for csv datasource annotations on top of test cases. 
Between junit4 and 5, there are some changes to the usage. 
We could try and create support for this. 

## Describe the situation before applying the recipe

```java
@AutoConfigureMockMvc
@ActiveProfiles(INTEGRATION_TEST)
@SpringBootTest
@TestExecutionListeners(mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS, listeners = {DBRiderTestExecutionListener.class})
@DBUnit(mergeDataSets = true, mergingStrategy = DataSetMergingStrategy.CLASS)
@DataSet(cleanBefore = true, cleanAfter = true, value = "")
public abstract class AbstractIntegrationTest {
    ...
}
```

## Describe the situation after applying the recipe

```java
@AutoConfigureMockMvc
@ActiveProfiles(INTEGRATION_TEST)
@SpringBootTest
@DBRider
@DBUnit(mergeDataSets = true, mergingStrategy = DataSetMergingStrategy.CLASS)
@DataSet(cleanBefore = true, cleanAfter = true, value = "")
public abstract class AbstractIntegrationTest {
    ...
}
```

Manual changes? 🤢 

## Any additional context
[see docs on dbrider here](https://github.com/database-rider/database-rider)

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?
I will try to come up with certain aspects, but additional support is always welcome as this is completely greenfielded topic from open rewrite perspective.

Jenson3210:
- fixes #624


Jenson3210:
@timtebeek , Used a similar workaround like you did in the other one to resolve the test cases. 
I consider this one as "ready for review" so removed draft!

src/main/java/org/openrewrite/java/testing/dbrider/ExecutionListenerToDbRiderAnnotation.java
```diff
@@ -0,0 +1,275 @@
...
+            public J.Annotation visitAnnotation(final J.Annotation annotation, final ExecutionContext ctx) {
+                if (EXECUTION_LISTENER_ANNOTATION_MATCHER.matches(annotation)) {
+                    acc.testExecutionListenersFound(annotation);
+                } else if (DBRIDER_ANNOTATION_MATCHER.matches(annotation)) {
+                    acc.dbriderFound = true;
+                }
+                return super.visitAnnotation(annotation, ctx);
+            }
+        };
+    }
```
timtebeek, line 71: I'm a bit puzzled why this is a scanning recipe, and even more so when this has a boolean that might be set by visiting any of dozens or more test classes, and then affect all of those test classes in turn.

Can we rework this to not be a scanning recipe? You can extract information all the same in a regular visitor by having an inline visitor in the visitCompilationUnit before you call super.visitCompilationUnit for instance. I hope that's clear, or the risk of the current approach is at least clear.

Jenson3210: Makes total sense... I was thinking of our specific usecase where we typicall have a slice of our tests which have a superclass. But totally ignored the fact that that's absolutely incorrect. Well spotted!

Jenson3210: Pushed a refactor for this.

timtebeek (approved):
Great to see! I've added some very light polish but with that we're good to merge. Thanks again!

Jenson3210:
Thank you for the patience in these ones and the great insights!

PR #633, opened 2024-11-05
Update AssertJ recipes to current recipe code style

jevanlingen:
## What's changed?
The AssertJ recipes use an older recipe code style; replaced all of 'm with an anonymous class + remove boilerplate code. The class `JUnitAssertInstanceOfToAssertThat` (created in https://github.com/openrewrite/rewrite-testing-frameworks/pull/606) i is used as the base template for all changed recipes.

## What's your motivation?
Code convergence

## Any additional context
Recipes in other packages I have left out of the scope of this PR. For example, the cleanup recipes also use the “old” recipe style.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertArrayEqualsToAssertThat.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertArrayEqualsToAssertThat.java
@@ -23,7 +23,7 @@ import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.MethodMatcher;
-import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
@@ -32,7 +32,10 @@ import org.openrewrite.java.tree.TypeUtils;
 import java.util.List;
 
 public class JUnitAssertArrayEqualsToAssertThat extends Recipe {
-    private static final String JUNIT_QUALIFIED_ASSERTIONS_CLASS_NAME = "org.junit.jupiter.api.Assertions";
+
+    private static final String JUNIT = "org.junit.jupiter.api.Assertions";
+    private static final String ASSERTJ = "org.assertj.core.api.Assertions";
+    private static final MethodMatcher ASSERT_ARRAY_EQUALS_MATCHER = new MethodMatcher(JUNIT + " assertArrayEquals(..)", true);
 
     @Override
     public String getDisplayName() {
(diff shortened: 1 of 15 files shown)


src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertArrayEqualsToAssertThat.java
```diff
@@ -32,7 +32,10 @@
...
 
 public class JUnitAssertArrayEqualsToAssertThat extends Recipe {
-    private static final String JUNIT_QUALIFIED_ASSERTIONS_CLASS_NAME = "org.junit.jupiter.api.Assertions";
+
+    private static final String JUNIT = "org.junit.jupiter.api.Assertions";
+    private static final String ASSERTJ = "org.assertj.core.api.Assertions";
+    private static final MethodMatcher ASSERT_ARRAY_EQUALS_MATCHER = new MethodMatcher(JUNIT + " assertArrayEquals(..)", true);
 
     @Override
     public String getDisplayName() {
```
timtebeek, line 41: Are you seeing these warnings in your IDE as well?
image

As a general rule we try to completely remove warnings on recipes we update, such that there's no nagging about warnings when we go to commit (enable that same check in your IDE). Would be a good moment to clear these warnings out as well by using a lower case `t` on `To`.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertFalseToAssertThat.java
```diff
@@ -44,66 +46,37 @@ public String getDescription() {
...
-                                actual
-                        );
-            } else {
                 Expression message = args.get(1);
                 JavaTemplate.Builder template = TypeUtils.isString(message.getType()) ?
                         JavaTemplate.builder("assertThat(#{any(boolean)}).as(#{any(String)}).isFalse();") :
                         JavaTemplate.builder("assertThat(#{any(boolean)}).as(#{any(java.util.function.Supplier)}).isFalse();");
-
-                method = template
+                return template
```
timtebeek, line 74: As seen on JUnitAssertInstanceOfToAssertThat there's no need for a ternary here or explicit type in that `any()` when reusing existing LST elements. The type will be deduced and can reduce the logic needed here.

https://github.com/openrewrite/rewrite-testing-frameworks/blob/152617c99daaca8384f3c4376ce183fdfd3a4268/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java#L67-L68

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
```diff
@@ -46,25 +46,25 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                J.MethodInvocation md = super.visitMethodInvocation(method, ctx);
-                if (!ASSERT_INSTANCE_OF_MATCHER.matches(md)) {
-                    return md;
+                J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
+                if (!ASSERT_INSTANCE_OF_MATCHER.matches(mi)) {
+                    return mi;
                 }
 
-                maybeAddImport("org.assertj.core.api.Assertions", "assertThat", false);
+                maybeAddImport("org.assertj.core.api.Assertions", "assertThat");
```
timtebeek, line 54: Appreciate this fix and the rename of the variable here to meet the conventions, thanks!

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertNotEqualsToAssertThat.java
```diff
@@ -45,117 +49,68 @@ public String getDescription() {
...
+                    return JavaTemplate.builder("assertThat(#{any()}).isNotEqualTo(#{any()});")
+                            .staticImports(ASSERTJ + ".assertThat")
+                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "assertj-core-3.24"))
+                            .build()
+                            .apply(getCursor(), mi.getCoordinates().replace(), actual, expected);
+                } else if (args.size() == 3 && !isFloatingPointType(args.get(2))) {
+                    Expression message = args.get(2);
+                    JavaTemplate.Builder template = TypeUtils.isString(message.getType()) ?
+                            JavaTemplate.builder("assertThat(#{any()}).as(#{any(String)}).isNotEqualTo(#{any()});") :
+                            JavaTemplate.builder("assertThat(#{any()}).as(#{any(java.util.function.Supplier)}).isNotEqualTo(#{any()});");
```
timtebeek, line 76: Same as above; no need for the ternary here if we use `#{any()}` instead of being explicit about the type. Please apply everywhere this pattern is repeated.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertArrayEqualsToAssertThat.java
jevanlingen: Yeah, I saw it too. I will go scroll through the code, in case I missed more of those.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertTrueToAssertThat.java
```diff
@@ -23,14 +23,17 @@
...
+import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.TypeUtils;
 
 import java.util.List;
 
 public class JUnitAssertTrueToAssertThat extends Recipe {
+
+    private static final MethodMatcher ASSERT_TRUE_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions" + " assertTrue(boolean, ..)");
```
timtebeek, line 35: No need to concatenate Strings here; please apply throughout; ideally we enforce this with a recipe as well.
```suggestion
    private static final MethodMatcher ASSERT_TRUE_MATCHER = new MethodMatcher("org.junit.jupiter.api.Assertions assertTrue(boolean, ..)");
```

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
jevanlingen: Well to be honest, I forgot to apply this everywhere. Once I did, there was one test that failed: `importAddedForCustomArguments` (JUnitAssertNotNullToAssertThatTest). As that test also has an issue number assigned to it (#419), it's probably important... 

So, do we want the `onlyIfReferenced` flag on or not?

src/main/java/org/openrewrite/java/testing/assertj/JUnitFailToAssertJFail.java
```diff
@@ -40,118 +45,86 @@ public String getDescription() {
...
+                    StringBuilder templateBuilder = new StringBuilder(ASSERTJ + ".fail(");
+                    for (int i = 0; i < args.size(); i++) {
+                        templateBuilder.append("#{any()}");
+                        if (i < args.size() - 1) {
+                            templateBuilder.append(", ");
+                        }
                     }
+                    templateBuilder.append(");");
+
+                    mi = JavaTemplate.builder(templateBuilder.toString())
```
timtebeek, line 87: This for loop and StringBuilder is another outdated pattern that could be replaced with just:
```java
                    String anyArgs = String.join(",", Collections.nCopies(args.size(), "#{any()}"));
                    mi = JavaTemplate.builder(ASSERTJ + ".fail(" + anyArgs + ");")
```
Please replace that as well wherever you see that, for simplicity.

timtebeek (commented):
Great to see this initial work already; helpful to have some examples to point out best practices (as opposed to what you were seeing). I've left a few more comments on patterns we can clear out; please apply those not just where I've left my remarks, but anywhere in these affected files.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertInstanceOfToAssertThat.java
timtebeek: Hmm yes let's keep that; we need it because our templates here are intentionally not context sensitive (would take a lot longer if we can not cache the templates for each time we make a change), and it's fine here to force add the import as we are guaranteed to use `assertThat` after the template returns.

jevanlingen: 👍🏻, I will put it back in for all other recipes (the other recipe's do simply not have such a test, but I guess the same rule applies to all).

timtebeek (approved):
Great simplification, thanks a lot! Should help to propagate the proper patterns.

Issue #636, opened 2024-11-06
AssertionsArgumentOrder does not work with Junit 4 assertions


adambir:
AssertionsArgumentOrder does not work with Junit 4 assertions, even though the main documentation mentions it : 
> Assertions such as org.junit.Assert.assertEquals expect the first argument to be the expected value and the second argument to be the actual value [...]

## What version of OpenRewrite are you using?
I am using

- Maven/Gradle plugin v5.43.2
- rewrite-testing-frameworks v2.21.0

## How are you running OpenRewrite?
Using the mvn plugin with a multi project module

## What is the smallest, simplest way to reproduce the problem?
```java
              import static org.junit.Assert.assertEquals;
                          
              class MyTest {
                  void someMethod() {
                      assertEquals(result(), "result");
                      assertEquals(result(), "result", "message");
                      assertEquals(0L, 1L);
                  }
                  String result() {
                      return "result";
                  }
              }
```

## What did you expect to see?

```java
              import static org.junit.Assert.assertEquals;
                          
              class MyTest {
                  void someMethod() {
                      assertEquals("result", result());
                      assertEquals("result", result(), "message");
                      assertEquals(0L, 1L);
                  }
                  String result() {
                      return "result";
                  }
              }
```

## What did you see instead?
```java
```java
              import static org.junit.Assert.assertEquals;
                          
              class MyTest {
                  void someMethod() {
                      assertEquals(result(), "result");
                      assertEquals(result(), "result", "message");
                      assertEquals(0L, 1L);
                  }
                  String result() {
                      return "result";
                  }
              }
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes, I'd like to propose a PR to generalize the behaviour to Junit 4. I would first need confirmation that this is something that could be considered.

adambir:
Added a PR  https://github.com/openrewrite/rewrite-testing-frameworks/pull/637

PR #637, opened 2024-11-06
Add support for Junit 4 matchers on AssertionsArgumentOrder

adambir:
- fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/636

## What's changed?
Added support for Junit 4 matchers in the `AssertionsArgumentOrder` recipe : 
- assertEquals
- assertNotEquals
- assertSame
- assertNotSame

## What's your motivation?
Junit 5 migration is sometimes complicated so supporting Junit 4 is a plus.

Did not add support for assertIterableEquals (does not exist in Junit 4), for assertArrayEquals and assertNull as the signature differs from Junit 5.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/cleanup/AssertionsArgumentOrder.java
+++ b/src/main/java/org/openrewrite/java/testing/cleanup/AssertionsArgumentOrder.java
@@ -32,13 +32,16 @@ import java.util.*;
 
 public class AssertionsArgumentOrder extends Recipe {
 
-    private static final MethodMatcher[] jupiterAssertionMatchers = new MethodMatcher[]{
+    private static final MethodMatcher[] junitAssertionsMatchers = new MethodMatcher[]{
             new MethodMatcher("org.junit.jupiter.api.Assertions assertArrayEquals(..)"),
             new MethodMatcher("org.junit.jupiter.api.Assertions assertEquals(..)"),
             new MethodMatcher("org.junit.jupiter.api.Assertions assertNotEquals(..)"),
             new MethodMatcher("org.junit.jupiter.api.Assertions assertSame(..)"),
             new MethodMatcher("org.junit.jupiter.api.Assertions assertNotSame(..)"),
-            new MethodMatcher("org.junit.jupiter.api.Assertions assertArrayEquals(..)")
+            new MethodMatcher("org.junit.Assert assertEquals(..)"),
+            new MethodMatcher("org.junit.Assert assertNotEquals(..)"),
+            new MethodMatcher("org.junit.Assert assertSame(..)"),
+            new MethodMatcher("org.junit.Assert assertNotSame(..)")
     };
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/cleanup/AssertionsArgumentOrderTest.java
```diff
@@ -54,7 +55,45 @@ String result() {
...
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import static org.junit.Assert.assertEquals;
+
+              class MyTest {
+                  void someMethod() {
+                      assertEquals(result(), "result");
+                      assertEquals(result(), "result", "message");
```
timtebeek, line 86: This line is converted into `assertEquals("result", result(), "message");`,
whereas argument order is: `assertEquals(message, expected, actual)`.

In this case we're making an incorrect change for JUnit 4 because the message used to be the first argument there, as opposed to the last argument with JUnit Jupiter. That's especially problematic for assertions on Strings, as the compiler does not warn you.

timtebeek (changes requested):
hi @adambir ; thanks for the initiative here! I'd originally left out JUnit 4 as the message argument handling there is different from JUnit Jupiter. I've placed a note on the new test to explain.

I think we could still support this provided we treat JUnit 4 and Jupiter separately, much like we also do for TestNG. Then we can separately check which indexed argument to use for the message, and which expected/actual arguments might need to be swapped. Perhaps good to update this PR to adjust the tests first, and only then look at separating those two.

src/main/java/org/openrewrite/java/testing/cleanup/AssertionsArgumentOrder.java
```diff
@@ -37,8 +37,21 @@ public class AssertionsArgumentOrder extends Recipe {
             new MethodMatcher("org.junit.jupiter.api.Assertions assertEquals(..)"),
             new MethodMatcher("org.junit.jupiter.api.Assertions assertNotEquals(..)"),
             new MethodMatcher("org.junit.jupiter.api.Assertions assertSame(..)"),
-            new MethodMatcher("org.junit.jupiter.api.Assertions assertNotSame(..)"),
-            new MethodMatcher("org.junit.jupiter.api.Assertions assertArrayEquals(..)")
```
adambir, line 41: Redundant with https://github.com/openrewrite/rewrite-testing-frameworks/pull/637/files#diff-932e6acf7d6226f1d48eee1a82ceaa83b9808e6791fae5b5cda5c4ebf42c9657R36

adambir:
Hi @timtebeek , thanks for the review.

I reworked the PR to handle the Junit 4 assertion starting with a String message. The message should now stay in first place. 
As suggested I separated the Junit5 and 4 matchers in the code. Support for assertArrayEquals and assertNull was added.

PR #641, opened 2024-11-17
don't exclude junit4 for testcontainers jupiter

sullis:
## What's changed?

JUnit5 migration recipe

## What's your motivation?

I have an application that uses Testcontainers junit-jupiter

We don't want to exclude JUnit4 because Testcontainers needs it.

### Checklist


timtebeek (approved):
Thanks! I've turned it into a wildcard for simplicity; Let's hope they get around to removing the dependency.

Issue #260, opened 2022-09-19
Transform assertTrue to assertSame or  assertNotSame


yeikel:
```java

org.junit.jupiter.api.Assertions.assertTrue(a == b);
org.junit.jupiter.api.Assertions.assertSame(a,b);

org.junit.jupiter.api.Assertions.assertTrue(a != b);
org.junit.jupiter.api.Assertions.assertNotSame(a, b);

```

yeikel:
That sounds fun. Hit me up in Slack :)

PR #642, opened 2024-11-17
Use JUnit5's `assertSame` or `assertNotSame` instead of `assertTrue(... == ...)`

mccartney:
## What's changed?
Adding a new recipe for converting:
```
assertTrue(a == b);
```
to
```
assertSame(a, b);
```
for JUnit5 test code as suggested in 
- #260

### Checklist


timtebeek (approved):
Thanks @mccartney ! I've applied a few small polishing commits, such as visiting method invocations directly, as opposed to via method declarations, to match cases nested in assertAll for instance. We'll also want to update the method type, as chained assertions look at the type, not the simple name.

PR #651, opened 2024-12-17
add Mockito1to4MigrationTest

sullis:
## What's changed?

add unit test for Mockito1to4Migration

## What's your motivation?

Motivation:  I could not find a unit test for Mockito1to4Migration

### Checklist


timtebeek (approved):
Thanks @sullis ! Odd how we never had an explicit recipe for this one, but good to have it now. :)

PR #655, opened 2024-12-21
Add recipe for replacing deprecated method AbstractDateAssert

kthoms:
This adds a recipe to replace `org.assertj.core.api.AbstractDateAssert.isEqualToIgnoringMillis(java.util.Date)` by the proposed method `isCloseTo()`

## What's changed?
New refaster recipe `java/org/openrewrite/java/testing/assertj/IsEqualToIgnoringMillisToIsCloseTo.java`

## What's your motivation?

closes#653

The recipe is not working yet as expected. It should add a static import for the method, but does not.

Not sure what I should test as negative case here.

@timtebeek 

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -42,6 +42,7 @@ dependencies {
 
     runtimeOnly("tech.picnic.error-prone-support:error-prone-contrib:latest.release:recipes")
     compileOnly("org.junit.jupiter:junit-jupiter-engine:latest.release")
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Great addition here @kthoms ! I've added it to the AssertJ best practice recipes just before the recipe that adopts static imports, that way there's no need for the recipe or tests to account for the import style. With that we're good to merge, thanks!

Issue #664, opened 2025-01-13
Code does not compile after AssertionsForClassTypes import is replaced by Assertions


barbulescu:
## What version of OpenRewrite are you using?
I am using 3.0.0

- OpenRewrite v6.0.0
- Maven/Gradle plugin v1.2.3
- rewrite-_module_ v3.0.0

## How are you running OpenRewrite?
not relevant

## What is the smallest, simplest way to reproduce the problem?
```java
import java.util.Map;
import static org.assertj.core.api.AssertionsForClassTypes.assertThat;

class Test {
  void test() {
    assertThat(Map.of("a", 1).entrySet()).hasNoNullFieldsOrProperties();
  }
}
```

## What did you expect to see?
```java
import java.util.Map;
import static org.assertj.core.api.AssertionsForClassTypes.assertThat;

class Test {
  void test() {
    assertThat(Map.of("a", 1).entrySet()).hasNoNullFieldsOrProperties();
  }
}
```

## What did you see instead?
```java
import java.util.Map;
import static org.assertj.core.api.Assertions.assertThat;

class Test {
  void test() {
    assertThat(Map.of("a", 1).entrySet()).hasNoNullFieldsOrProperties();
  }
}
```
Code does not compile anymore after this change.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
yes

timtebeek:
Thanks for the report! Curious case; that'l likely stem from these lines that have been in there since 2021:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/fc9d8b46f40b336d4ab0149078861f7743758a2b/src/main/resources/META-INF/rewrite/assertj.yml#L61-L63

The `hasNoNullFieldsOrProperties` seems to be defined on `org.assertj.core.api.AbstractObjectAssert`, which I'd imagine is returned from both `AssertionsForClassTypes.assertThat` and `Assertions.assertThat`.

Is there any other way to make it work with `Assertions.assertThat`?

scordio:
I think the root cause is partially how the client code uses AssertJ classes and the expectation behind that, but also how `AssertionsForClassTypes` might be redundant and misleading when placed next to `Assertions`.

[...]

PR #665, opened 2025-01-13
AssertionsForClassTypes import is replaced by Assertions

barbulescu:
## What's changed?
Add test to reproduce the issue.

## What's your motivation?
Code does not compile after execution.

### Checklist


timtebeek:
Thanks @barbulescu ! I'd missed your PR as I don't get notified about drafts; would you want to continue to work on this? Would you need any guidance as you do so?

barbulescu:
Yes, I would like to continue but I need some guidance.

timtebeek:
Had a brief look here; looks like we get back a different assertion object from the `Assertions` class then we do from the `AssertionsForClassTypes`
image

Looks like `AbstractCollectionAssert extends AbstractAssert`, and not the `AbstractObjectAssert` we get from `AssertionsForClassTypes`.

timtebeek:
@scordio I hope you don't mind me tagging you here; don't feel pressured to respond if you're busy! For context, we have some recipes that nudge folks to use the `Assertions` class for their imports instead of any more specific AssertionsForXyz classes:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/fc9d8b46f40b336d4ab0149078861f7743758a2b/src/main/resources/META-INF/rewrite/assertj.yml#L61-L69
Up to now that had mostly worked quite well for us; it's just in this case that we've encountered a difference in the returned types (`AbstractObjectAssert` vs `AbstractCollectionAssert`) between `AssertionsForClassTypes` and `Assertions` when passing in a collection, and what assertion methods we can call on those return types (`hasNoNullFieldsOrProperties` not on `AbstractCollectionAssert`, as that `extends AbstractAssert`, not `AbstractObjectAssert`).

Is my expectation just off that folks should be able to use `Assertions` wherever?
Any conscious choice to have `AbstractCollectionAssert` not offer `hasNoNullFieldsOrProperties`?

scordio:
At first glance, it looks like an oversight on our end. I'll have a deeper look and get back to you!

scordio:
Some thoughts added at https://github.com/openrewrite/rewrite-testing-frameworks/issues/664#issuecomment-2600212334.

> Is my expectation just off that folks should be able to use `Assertions` wherever?

Your expectation makes sense, although there can be corner cases that require a more specific strategy (more in my other comment).

> Any conscious choice to have `AbstractCollectionAssert` not offer `hasNoNullFieldsOrProperties`?

Yes – in this context, `hasNoNullFieldsOrProperties` makes little sense as you usually don't want to validate the internal state of a `Collection` instance.

timtebeek:
Thanks for chiming in @scordio ! With more details added to the associated issue let's track the suggested next steps there.
- #664

If it turns out we do want to make code changes to accommodate we can already reopen this pull request.

PR #667, opened 2025-01-17
Add `org.mockito:mockito-junit-jupiter:3.x` in explicit test scope

MBoegers:
## What's changed?

`org.openrewrite.java.testing.mockito.Mockito1to3Migration` added `org.mockito:mockito-junit-jupiter:3.x` in runtime scope which is pretty dangerous.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/mockito.yml
+++ b/src/main/resources/META-INF/rewrite/mockito.yml
@@ -174,6 +174,7 @@ recipeList:
       version: 3.x
       onlyIfUsing: org.mockito.junit.jupiter.*
       acceptTransitive: true
+      scope: test
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
       groupId: org.mockito
       artifactId: "*"


src/main/resources/META-INF/rewrite/mockito.yml
```diff
@@ -174,6 +174,7 @@ recipeList:
       version: 3.x
       onlyIfUsing: org.mockito.junit.jupiter.*
       acceptTransitive: true
+      scope: test
```
timtebeek, line 177: Hmm; I thought in the past we intentionally did not specify a scope, as `AddDependency` should then determine the scope based on where the `onlyIfUsing` classes are being used. It's not per-se wrong, but could for a shared testing library then for instance not add the dependency in the correct scope for downstream users of that testing lib.

How'd you arrive at this change? And did you verify it against our internal services that this indeed fixes it?

timtebeek (approved):
Approved despite the question/context; if you've seen this help then we can merge as is.

src/main/resources/META-INF/rewrite/mockito.yml
MBoegers: The behavior of determining the right scope is not documented. It says default is `runtime` and we saw this recipe add the dependency as `implementation`, updates it up to modern Spring and also updating the `testImplementation`.

MBoegers:
Will double check

PR #669, opened 2025-01-23
Feat find unit tests

justine-gehring:
## What's changed?

Adding a recipe to find examples of how a method is being used in unit tests.

## What's your motivation?

This recipe can then be used for rewrite-comprehension to feed more information to an LLM.


src/main/java/org/openrewrite/java/testing/search/FindUnitTests.java
```diff
@@ -0,0 +1,118 @@
...
+        return Preconditions.check(new IsLikelyNotTest().getVisitor(), tableRowVisitor);
+    }
+
+}
+
+@Value
+class UnitTest {
+    String clazz;
+    String unitTestName;
+    String unitTest;
```
timtebeek, line 117: Instead of String we might want to use the types of the LST elements matched here instead.

Issue #640, opened 2024-11-08
JUnit 4 to 5 migration does not add the mockito-junit-jupiter dependency


FieteO:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v5.22.0

## How are you running OpenRewrite?

I am using the command line invocation:

```sh
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-spring:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.spring.boot2.SpringBoot2JUnit4to5Migration -Drewrite.exportDatatables=true
```

## What is the smallest, simplest way to reproduce the problem?

```java
import org.junit.runner.RunWith;
import org.mockito.junit.MockitoJUnitRunner;

@RunWith(MockitoJUnitRunner.class)
public class MyClass {}
```

```xml
...
<dependencies>
  <dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter</artifactId>
    <scope>test</scope>
  </dependency>
  <dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <scope>test</scope>
  </dependency>
</dependencies>
```

## What did you expect to see?

```xml
...
<dependencies>
  <dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter</artifactId>
    <scope>test</scope>
  </dependency>
  <dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <scope>test</scope>
  </dependency>
  <dependency>
      <groupId>org.mockito</groupId>
      <artifactId>mockito-junit-jupiter</artifactId>
      <scope>test</scope>
  </dependency>
</dependencies>
```

## What did you see instead?
The `mockito-junit-jupiter` dependency is not added.
```xml
...
<dependencies>
  <dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter</artifactId>
    <scope>test</scope>
  </dependency>
  <dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <scope>test</scope>
  </dependency>
</dependencies>
```

[...]

PR #670, opened 2025-01-28
Add mockito-junit-jupiter when using runner in JUnit 4

promanenko:
## What's changed?
Add `mockito-junit-jupiter`dependency 
https://github.com/openrewrite/rewrite-testing-frameworks/blob/6aa80c958f3fc0d0c9954c7a24dd0481fe49eb83/src/main/resources/META-INF/rewrite/mockito.yml#L171-L177 expected to handle `@MockitoExtension` didn't work because `AddDependency` is a scanning recipe and scanning for `org.mockito.junit.jupiter.*` is performed before the package is added on `editSources` phase by [RunnerToExtension](https://github.com/openrewrite/rewrite-testing-frameworks/blob/6aa80c958f3fc0d0c9954c7a24dd0481fe49eb83/src/main/resources/META-INF/rewrite/junit5.yml#L201)
Now `onlyIfUsing` is looking for the initial runners that are present on the `scanSources` phase. 
**Note:** if the old version of recipe is ran in two cycles then it adds `mockito-junit-jupiter` dependency in the second cycle.

## What's your motivation?
- Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/640

Alternatives considered:
1. Modify `RunnerToExtension` to handle addition of a dependency if needed. It contradicts the best practice [If it can be declarative, it should be declarative](
https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#if-it-can-be-declarative-it-should-be-declarative)
2. Make `RunnerToExtension` recipe to cause another cycle. It contradicts the best practice [Stay single cycle](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#stay-single-cycle)

## Any additional context
There is a similar `rewrite runners -> add dependency` sequence in [vertx migration](https://github.com/openrewrite/rewrite-testing-frameworks/blob/6aa80c958f3fc0d0c9954c7a24dd0481fe49eb83/src/main/resources/META-INF/rewrite/junit5.yml#L234C29-L234C65) but it is looking for `org.vertx.testtools.VertxUnitRunner` which doesn't exist and it seems to be a separate issue.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/junit5.yml
+++ b/src/main/resources/META-INF/rewrite/junit5.yml
@@ -196,6 +196,13 @@ tags:
   - junit
   - mockito
 recipeList:
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: org.mockito
+      artifactId: mockito-junit-jupiter
+      version: 4.x
+      onlyIfUsing: org.mockito..MockitoJUnit*Runner
+      acceptTransitive: true
+      scope: test
   - org.openrewrite.java.testing.mockito.Mockito1to4Migration
   - org.openrewrite.java.testing.mockito.MockitoJUnitRunnerSilentToExtension
   - org.openrewrite.java.testing.junit5.RunnerToExtension:
--- a/src/main/resources/META-INF/rewrite/junit5.yml
+++ b/src/main/resources/META-INF/rewrite/junit5.yml
@@ -253,7 +253,7 @@ recipeList:
       groupId: io.vertx
       artifactId: vertx-junit5
       version: 4.x
-      onlyIfUsing: io.vertx.junit5.VertxExtension
+      onlyIfUsing: org.vertx.testtools.VertxUnitRunner
       acceptTransitive: true
 ---
 type: specs.openrewrite.org/v1beta/recipe
--- a/src/main/resources/META-INF/rewrite/mockito.yml
+++ b/src/main/resources/META-INF/rewrite/mockito.yml
@@ -168,13 +168,6 @@ recipeList:
   - org.openrewrite.java.testing.mockito.MockUtilsToStatic
   - org.openrewrite.java.testing.junit5.MockitoJUnitToMockitoExtension
   - org.openrewrite.java.testing.mockito.ReplacePowerMockito
-  - org.openrewrite.java.dependencies.AddDependency:
-      groupId: org.mockito
-      artifactId: mockito-junit-jupiter
-      version: 3.x
-      onlyIfUsing: org.mockito.junit.jupiter.*
-      acceptTransitive: true
-      scope: test
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
       groupId: org.mockito
       artifactId: "*"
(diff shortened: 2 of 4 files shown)


MBoegers:
Looks good so far, easy and small 👍 
Do you think we still need the `AddDependency` Recipe the in the [rewrite-testing-frameworks/src/main/resources/META-INF/rewrite/mockito.yml](https://github.com/openrewrite/rewrite-testing-frameworks/blob/6aa80c958f3fc0d0c9954c7a24dd0481fe49eb83/src/main/resources/META-INF/rewrite/mockito.yml#L171-L177)?

promanenko:
> Looks good so far, easy and small 👍 Do you think we still need the `AddDependency` Recipe the in the [rewrite-testing-frameworks/src/main/resources/META-INF/rewrite/mockito.yml](https://github.com/openrewrite/rewrite-testing-frameworks/blob/6aa80c958f3fc0d0c9954c7a24dd0481fe49eb83/src/main/resources/META-INF/rewrite/mockito.yml#L171-L177)?

Good catch, now I don't see a case when `onlyIfUsing: org.mockito.junit.jupiter.* ` will catch what is not yet handled by `onlyIfUsing: org.mockito..MockitoJUnit*Runner`. `Mockito1to3Migration` could be called separately, but if `org.mockito.junit.jupiter.*` is already present then JUnit 5 is used and `mockito-junit-jupiter` dependency should be also present already.
I left the old `AddDependency` as a safety net and not to break any subtle edge case, but I would be happy to remove it if you think it's safe.

src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -196,6 +196,13 @@ tags:
   - junit
   - mockito
 recipeList:
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: org.mockito
+      artifactId: mockito-junit-jupiter
+      version: 4.x
+      onlyIfUsing: org.mockito..MockitoJUnit*Runner
```
timtebeek, line 203: Neat to see you picked up on the pattern to use here!

timtebeek (approved):
Good to see the fix here, thanks! Indeed best to add the dependency if not present transitively, and you've correctly deduced to use the old classes through `onlyIfUsing`. Thanks!

src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -246,7 +253,7 @@ recipeList:
       groupId: io.vertx
       artifactId: vertx-junit5
       version: 4.x
-      onlyIfUsing: io.vertx.junit5.VertxExtension
+      onlyIfUsing: org.vertx.testtools.VertxUnitRunner
```
promanenko, line 256: "org.vertx.testtools.VertxUnitRunner" seems to be non-existent. Google search shows it only as a part of openrewrite docs
image

timtebeek: Thanks a lot! I suspect that might have been botched at some point. I've now looked up the corresponding classes and have arrived at: https://github.com/openrewrite/rewrite-testing-frameworks/commit/2a3d615dc1eb1e70fb9386c502e10362b2b63982

PR #673, opened 2025-02-11
Junit4 to assertj

### LINKED ISSUE #672: Enable JUnitAssertXXXToAssertThat for JUnit 4
## What problem are you trying to solve?
I want to migrate JUnit 4 assertions to AssertJ. The existing recipes migrate only JUnit 5 assertions. From the recipe's description it is also not clear that "JUnit `assertEquals` to AssertJ" refer to JU 5 only.

## Describe the solution you'd like
The existing recipes might be enhanced to cover both cases by trying to match methods for `org.junit.jupiter.api.Assertions` or `org.junit.Assert` and resolve the parameters differently. The assertions are quite similar except for the order of arguments.

There could be also dedicated JUnit**4**AssertXXXToAssertThat recipes. This would likely produce more duplicated code.

Migrating to JUnit 5 is the ultimate goal, but I would like to do it stepwise and migrate the assertions before migrating the tests themselves.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
Yes, I have a working draft for one recipe already. I could contribute this already, but it feels as if all JUnitAssertXXXToAssertThat recipses should be enhanced together. Might be surprising if only one recipe behaves different.

kthoms:
## What's changed?

This change enables JUnitAssertXXXToAssertThat recipes to also handle JUnit 4 assertions. Therefore it introduces an abstract superclass that allows to configure JUnit 4 & 5 specific method matchers and a visitor that handles both configurations.

Tests are added for the refactoring of JUnit 4 style assertions. Note that JUnit 4 has mostly matching assertions, but less capabilities in detail:
- No support for message suppliers
- No assertion for instanceOf
The tests are reduced by that for JUnit 4.

## What's your motivation?

- fixes #672 

This is for a migration to JUnit 5 where I want to migrate the assertions before the test classes themselves, because due to extensive use of JUnit 4 rules and parametrized tests the migration has to be done partially manually and changes to the assertions are distracting.

The structure of the abstract's class logic and naming.

Performance consideration: When JUnit 4 style is migrated, first the JUnit 5 configuration is tried to be matched.

### Checklist


timtebeek:
As discussed here we'll close this PR due to the conflicts and complicated maintenance going forward.
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/672#issuecomment-2778652098

Folks that wish to remain on JUnit 4, or first wish to migrate to AssertJ before moving to JUnit 5 can use the recipes as they are on this branch. Thanks again!

PR #677, opened 2025-02-21
Fix failing `warnStrictnessRuleAddMockitoSettingsWithStaticImport()`

knutwannheden:


knutwannheden:
@timtebeek Strangely enough, locally the test only passes when I apply this change, but on CI it is the other way around.

timtebeek:
Odd indeed! Explored for a brief bit: I'm at least seeing the same as CI, even when I attempt to explicitly remove that static import.

knutwannheden:
Strange...

PR #678, opened 2025-02-27
MockitoWhenOnStaticToMockStatic improvements

jevanlingen:
## What's changed?
The `MockitoWhenOnStaticToMockStatic` recipe has some improvements:
 - Use a lambda in when-statement (as it is supposed to be)
 - Uses `MockedStatic` field when mock is used in a `@Before` method
 - Recipe works with nested statements 

## What's your motivation?
Recipe did not do a proper migration.


src/test/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStaticTest.java
```diff
@@ -147,4 +177,148 @@ void test() {
...
+                      assertEquals(A.getNumber(), -1);
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void shouldNotRefactorMockito_WhenMockIsAssigned() {
```
jevanlingen, line 299: ⚠️ You could debate the recipe should still do something, because `when(A.getNumber())` will no longer work anymore (the bytecode manipulation hack of mockito-inline has been removed). I chose to leave it out of scope and at least make sure the recipe does not break anything.

If you don't agree, let me know. I can work on this recipe even more.

src/test/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStaticTest.java
```diff
@@ -91,28 +92,58 @@ void test() {
     }
 
     @Test
-    void shouldHandleMultipleStaticMocks() {
+    @Disabled("Current recipe does only work for direct `method invocation` in a `method declaration`.")
+    void shouldRefactorMockito_WhenNestedStatements() {
```
jevanlingen, line 96: 💡 The recipe does not work for nested statements. Though that can be fixed, I chose not to dive into this to deeply.

knutwannheden: I am not sure I understand why this is a problem. If it isn't hard to fix, I think we should do it, because this will be really surprising.

knutwannheden: Can we more or less just replace `getWhenArg()` with an override of `visitMethodInvocation()`?

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -15,26 +15,31 @@
...
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import static org.openrewrite.java.VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER;
+import static org.openrewrite.java.VariableNameUtils.generateVariableName;
+import static org.openrewrite.java.tree.Flag.Static;
+
 public class MockitoWhenOnStaticToMockStatic extends Recipe {
 
+    private static final AnnotationMatcher BEFORE = new AnnotationMatcher("org.junit.Before");
+    private static final AnnotationMatcher AFTER = new AnnotationMatcher("org.junit.After");
```
knutwannheden, line 42: Like in the [example of this issue](https://github.com/powermock/powermock/issues/1089), I think we should probably also support `@BeforeClass` and `@AfterClass`, right? Although note that the example contains a bug, as it has no corresponding `@AfterClass` annotated method.

src/test/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStaticTest.java
knutwannheden: It seems like maybe the statements can be moved into the try-with-resources block and then after that the super-call is made on that result to transform the statements inside that statement list. But it is indeed more complicated than I thought.

src/test/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStaticTest.java
```diff
@@ -21,13 +21,24 @@
...
 
+    //language=java
+    public static final SourceSpecs CLASS_A = java(
+      """
+        public class A {
+            public static Integer getNumber() {
+                return 42;
+            }
+        }
+        """
```
Laurens-W, line 39: We can still add `SourceSpec::skip` as we don't want it to be considered part of the assertions

Laurens-W (approved):
Other than the 1 tiny comment LGTM

PR #679, opened 2025-03-03
Fix final field `@InjectMock` recipe

nielsdebruin:
## What's changed?

The `NoInitializationForInjectMock` recipe removes unnecessary initializers for fields that have already been marked with the `@InjectMocks` annotation. However, a compile error will occur if that field is currently final since it sees an uninitialized field. The recipe has been modified to remove the final modifier to remedy this.

## Motivation
Take the following code:
```java
@InjectMocks
final MyObject myObject = new MyObject("someField");
```
Applying the current recipe yields
```java
@InjectMocks
final MyObject myObject; // Does not compile
```
Applying the modified recipe yields
```java
@InjectMocks
MyObject myObject;
```

Alternatively, you might consider skipping any final fields instead or adding an option to choose the behavior.

### Checklist


timtebeek (approved):
Thanks @nielsdebruin !

PR #684, opened 2025-03-18
ReplacePowerMockito to support `PowerMockito.whenNew` constructor mocking - the simplest case

greg-at-moderne:
## What's changed?

Adding support to `ReplacePowerMockito` for dealing with constructor mocking using `PowerMockito.whenNew` in the **simplest case** only - i.e. when using no arguments constructor.

The plan is to have another PR for constructors with arguments.

## What's your motivation?

To address one of the customers' feedback.

## Context

Related:#646

PR #686, opened 2025-03-20
Supporting arguments in PowerMockitoWhenNewToMockito

greg-at-moderne:
## What's changed?

Continuation of #684.

Enhancing the support for `PowerMockitoWhenNewToMockito` recipe for mocking constructors other than 0-arg ones (as #684 did).

## What's your motivation?

To address one of the customers' feedback.

Yes! I am glad you are asking:

As submitted the code will **ignore** the actual mocked constructor arguments, resulting in mocking all creations of objects of given class. As opposed to only some, i.e. these whose arguments matched. This is because based on my research, I couldn't find any direct counterpart of input like `.withArguments("param1", 3)`.

The reason I think ignoring the arguments is that I think typically you are interested in mocking just one call to the constructor. And even if you are interested in multiple calls. This is still **possible** with the code given. You mock one class, and then call `when()` twice with different arguments to provide different return values of the object methods.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoWhenNewToMockito.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoWhenNewToMockito.java
@@ -35,7 +35,9 @@ import static org.openrewrite.java.VariableNameUtils.generateVariableName;
 public class PowerMockitoWhenNewToMockito extends Recipe {
 
     private static final MethodMatcher PM_WHEN_NEW = new MethodMatcher("org.powermock.api.mockito.PowerMockito whenNew(..)");
-    private static final MethodMatcher WITH_NO_ARGUMENTS = new MethodMatcher("*..* withNoArguments(..)");
+    private static final MethodMatcher WITH_NO_ARGUMENTS = new MethodMatcher("*..* withNoArguments()");
+    private static final MethodMatcher WITH_ARGUMENTS = new MethodMatcher("*..* withArguments(..)");
+    private static final MethodMatcher WITH_ANY_ARGUMENTS = new MethodMatcher("*..* withAnyArguments()");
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoWhenNewToMockito.java
```diff
@@ -35,7 +35,9 @@
 public class PowerMockitoWhenNewToMockito extends Recipe {
 
     private static final MethodMatcher PM_WHEN_NEW = new MethodMatcher("org.powermock.api.mockito.PowerMockito whenNew(..)");
-    private static final MethodMatcher WITH_NO_ARGUMENTS = new MethodMatcher("*..* withNoArguments(..)");
+    private static final MethodMatcher WITH_NO_ARGUMENTS = new MethodMatcher("*..* withNoArguments()");
```
jevanlingen, line 38: Shouldn't `*..*` not actually be `org.powermock.api.mockito.PowerMockito`?

jevanlingen (approved):

src/main/java/org/openrewrite/java/testing/mockito/PowerMockitoWhenNewToMockito.java
```diff
@@ -35,7 +35,9 @@
...
+    private static final MethodMatcher WITH_ARGUMENTS = new MethodMatcher("*..* withArguments(..)");
+    private static final MethodMatcher WITH_ANY_ARGUMENTS = new MethodMatcher("*..* withAnyArguments()");
```
jevanlingen, line 40: Ow, now I look a little bit better, there is no unit test for this case. Please add one (or change an existing one)

greg-at-moderne: efe6c32

Issue #224, opened 2022-04-27
Remove "@VisibleForTesting" if code is accessed from production code


yeikel:
Similar SPEC : [RSPEC-5803](https://jira.sonarsource.com/browse/RSPEC-5803)

tkvangorder:
This could actually be quite tricky to implement when encountered. What would the remediation strategy be? Look for an accessor? This might be best as a search recipe.

yeikel:
> This could actually be quite tricky to implement when encountered. What would the remediation strategy be? Look for an accessor? This might be best as a search recipe. 

1. Look for any non test file accessing the variable
2. Remove the annotation if #1 is met

tkvangorder:
oh, gotcha. Yep, that makes sense.

PR #689, opened 2025-04-02
Added recipe to remove `@VisibleForTesting` annotation when used from production code

JohannisK:
## What's changed?
Added recipe to remove @VisibleForTesting annotation when used from production code

## What's your motivation?
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/224

@jevanlingen 
@Laurens-W 
@greg-at-moderne

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
+++ b/src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
@@ -16,23 +16,26 @@
 package org.openrewrite.java.testing.cleanup;
 
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
 import org.openrewrite.ScanningRecipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.RemoveAnnotation;
-import org.openrewrite.java.marker.JavaSourceSet;
+import org.openrewrite.java.search.IsLikelyNotTest;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
 import org.openrewrite.java.tree.TypeUtils;
 
 import java.util.*;
 
-public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends ScanningRecipe<RemoveVisibleForTestingAnnotationWhenUsedInProduction.Scanned> {
+import static org.openrewrite.Preconditions.and;
 
-    public static class Scanned {
-        List<String> methods = new ArrayList<>();
-        List<String> fields = new ArrayList<>();
-        List<String> classes = new ArrayList<>();
+public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends ScanningRecipe<RemoveVisibleForTestingAnnotationWhenUsedInProduction.VisibleForTesting> {
+
+    public static class VisibleForTesting {
+        Set<String> methodPatterns = new HashSet<>();
+        Set<String> fieldPatterns = new HashSet<>();
+        Set<String> classPatterns = new HashSet<>();
     }
 
     @Override
@@ -51,21 +54,21 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
     }
 
     @Override
-    public Scanned getInitialValue(ExecutionContext ctx) {
-        return new Scanned();
+    public VisibleForTesting getInitialValue(ExecutionContext ctx) {
+        return new VisibleForTesting();
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getScanner(Scanned acc) {
-        return new JavaIsoVisitor<ExecutionContext>() {
+    public TreeVisitor<?, ExecutionContext> getScanner(VisibleForTesting acc) {
+        JavaIsoVisitor<ExecutionContext> scanningVisitor = new JavaIsoVisitor<ExecutionContext>() {
 
             @Override
             public J.FieldAccess visitFieldAccess(J.FieldAccess fa, ExecutionContext ctx) {
                 if (fa.getTarget().getType() instanceof JavaType.Class) {
-                    checkAndRegister(acc.classes, fa.getTarget().getType());
+                    checkAndRegister(acc.classPatterns, fa.getTarget().getType());
                 }
                 if (fa.getName().getFieldType() != null) {
-                    checkAndRegister(acc.fields, fa.getName().getFieldType());
+                    checkAndRegister(acc.fieldPatterns, fa.getName().getFieldType());
                 }
                 return super.visitFieldAccess(fa, ctx);
             }
@@ -73,7 +76,7 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
             @Override
             public J.MemberReference visitMemberReference(J.MemberReference mr, ExecutionContext ctx) {
                 if (mr.getMethodType() != null) {
-                    checkAndRegister(acc.methods, mr.getMethodType());
+                    checkAndRegister(acc.methodPatterns, mr.getMethodType());
                 }
                 return super.visitMemberReference(mr, ctx);
             }
@@ -81,7 +84,7 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
             @Override
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
                 if (mi.getMethodType() != null) {
-                    checkAndRegister(acc.methods, mi.getMethodType());
+                    checkAndRegister(acc.methodPatterns, mi.getMethodType());
                 }
                 return super.visitMethodInvocation(mi, ctx);
             }
@@ -89,23 +92,19 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
             @Override
             public J.NewClass visitNewClass(J.NewClass nc, ExecutionContext ctx) {
                 if (nc.getConstructorType() != null) {
-                    checkAndRegister(acc.methods, nc.getConstructorType());
+                    checkAndRegister(acc.methodPatterns, nc.getConstructorType());
                 }
                 if (nc.getClazz() != null && nc.getClazz().getType() instanceof JavaType.Class) {
-                    checkAndRegister(acc.classes, nc.getClazz().getType());
+                    checkAndRegister(acc.classPatterns, nc.getClazz().getType());
                 }
                 return super.visitNewClass(nc, ctx);
             }
 
-            private void checkAndRegister(List<String> target, JavaType type) {
+            private void checkAndRegister(Set<String> target, JavaType type) {
                 if (!target.contains(TypeUtils.toString(type))) {
                     getAnnotations(type).forEach(annotation -> {
                         if ("VisibleForTesting".equals(annotation.getClassName())) {
-                            J.CompilationUnit cu = getCursor().firstEnclosing(J.CompilationUnit.class);
-                            if (cu != null &&
-                                    cu.getMarkers().findFirst(JavaSourceSet.class).filter(s -> "main".equals(s.getName().toLowerCase(Locale.ROOT))).isPresent()){
-                                target.add(TypeUtils.toString(type));
-                            }
+                            target.add(TypeUtils.toString(type));
                         }
                     });
                 }
@@ -122,11 +121,12 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
                 return Collections.emptyList();
             }
         };
+        return Preconditions.check(new IsLikelyNotTest().getVisitor(), scanningVisitor);
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
-        return new JavaIsoVisitor<ExecutionContext>() {
+    public TreeVisitor<?, ExecutionContext> getVisitor(VisibleForTesting acc) {
+        JavaIsoVisitor<ExecutionContext> visitor = new JavaIsoVisitor<ExecutionContext>() {
 
             @Override
             public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations vd, ExecutionContext ctx) {
@@ -134,8 +134,8 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
                 if (!variableDeclarations.getVariables().isEmpty()) {
                     // if none of the variables in the declaration are used from production code, the annotation should be kept
                     boolean keepAnnotation = variableDeclarations.getVariables().stream()
-                            .filter(elem -> elem.getVariableType() != null)
-                            .noneMatch(elem -> acc.fields.contains(TypeUtils.toString(elem.getVariableType())));
+                          .filter(elem -> elem.getVariableType() != null)
+                          .noneMatch(elem -> acc.fieldPatterns.contains(TypeUtils.toString(elem.getVariableType())));
                     if (!keepAnnotation) {
                         return (J.VariableDeclarations) getElement(ctx, variableDeclarations.getLeadingAnnotations(), variableDeclarations);
                     }
@@ -146,7 +146,7 @@ public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends Scann
             @Override
             public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration md, ExecutionContext ctx) {
                 J.MethodDeclaration methodDeclaration = super.visitMethodDeclaration(md, ctx);
-                if (methodDeclaration.getMethodType() != null && acc.methods.contains(TypeUtils.toString(methodDeclaration.getMethodType()))) {
+                if (methodDeclaration.getMethodType() != null && acc.methodPatterns.contains(TypeUtils.toString(methodDeclaration.getMethodType()))) {
                     return (J.MethodDeclaration) getElement(ctx, methodDeclaration.getLeadingAnnotations(), methodDeclaration);
                 }
                 return methodDeclaration;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                if (method.getMethodType() != null) {
+                    if (!acc.methods.contains(method.getMethodType())) {
+                        method.getMethodType().getAnnotations().forEach(annotation -> {
+                            if ("VisibleForTesting".equals(annotation.getClassName())) {
+                                J.CompilationUnit compilationUnit = getCursor().firstEnclosing(J.CompilationUnit.class);
+                                if (compilationUnit != null && compilationUnit.getSourcePath().startsWith("src/main")) {
```
jkschneider, line 98: This should check the `JavaSourceSet` marker to see if it has name `main`. Not every project will use the Maven idiom for source set pathing.

jkschneider (changes requested):

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,196 @@
...
+        return new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations multiVariable, ExecutionContext ctx) {
+                J.VariableDeclarations variableDeclarations = super.visitVariableDeclarations(multiVariable, ctx);
+                if(!variableDeclarations.getVariables().isEmpty()) {
+                    // if none of the variables in the declaration are used from production code, the annotation should be kept
+                    boolean keepAnnotation = variableDeclarations.getVariables().stream().noneMatch(elem -> acc.fields.contains(elem.getVariableType()));
+                    if (!keepAnnotation) {
+                        Optional<J.Annotation> annotation = variableDeclarations.getLeadingAnnotations().stream()
```
jevanlingen, line 153: Lot's of repetitive code here. I think you go very far with a generic function. Just out the top of my head:

```java
private <@Nullable T extends J> getElement(List<J.Annotation> leadingAnnotations) {
    Optional<J.Annotation> annotation = leadingAnnotations().stream() .filter(elem -> "VisibleForTesting".equals(elem.getSimpleName())).findFirst();
    if (annotation.isPresent() && annotation.get().getType() instanceof JavaType.Class) {
        JavaType.Class type = (JavaType.Class) annotation.get().getType();
        return (T) new RemoveAnnotation("@" + type.getFullyQualifiedName()).getVisitor().visitNonNull(variableDeclarations, ctx, getCursor().getParentOrThrow());
    }

    return null;
}

File: src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java [hunk 3]
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Remove `@VisibleForTesting` annotation when target is used in production";
+    }
+
+    @Override
+    public String getDescription() {
+        return "The `@VisibleForTesting` annotation is used when a method or a field has been made more accessible then it would normally be, solely for testing purposes. " +
```
jevanlingen, line 48:
```suggestion
        return "The `@VisibleForTesting` annotation marks a method or field that is intentionally made more accessible (e.g., changing its visibility from private or package-private to public or protected) solely for testing purposes.." +
                "The annotation serves as an indicator that the increased visibility is not part of the intended public API but exists only to support testability. " +
```

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+        return new Scanned();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Scanned acc) {
+        return new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.FieldAccess visitFieldAccess(J.FieldAccess fieldAccess, ExecutionContext ctx) {
+                // Mark classes
```
jevanlingen, line 65: I don't think these two comments are necessary. It quite easy to understand by the if-statement itself.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            @Override
+            public J.NewClass visitNewClass(J.NewClass newClass, ExecutionContext ctx) {
+                // Mark constructors
+                if (newClass.getConstructorType() != null) {
+                    checkAndRegister(acc.methods, newClass.getConstructorType());
+                }
+                // Mark classes
+                if (newClass.getClazz() != null && newClass.getClazz().getType() != null && newClass.getClazz().getType() instanceof JavaType.Class) {
```
jevanlingen, line 91:
```suggestion
                if (newClass.getClazz() != null && newClass.getClazz().getType() instanceof JavaType.Class) {
```

The extra null check is unnecessary. An `instanceof` check is _also_ a null check.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+                    checkAndRegister(acc.classes, newClass.getClazz().getType());
+                }
+                return super.visitNewClass(newClass, ctx);
+            }
+
+            private void checkAndRegister(List<String> target, JavaType type) {
+                if (!target.contains(TypeUtils.toString(type))) {
+                    getAnnotations(type).forEach(annotation -> {
+                        if ("VisibleForTesting".equals(annotation.getClassName())) {
+                            J.CompilationUnit compilationUnit = getCursor().firstEnclosing(J.CompilationUnit.class);
```
jevanlingen, line 101: Same here, the `J.CompilationUnit` is ofter abbreviated to `cu`.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+                }
+                // Mark fields
+                if (fieldAccess.getName().getFieldType() != null) {
+                    checkAndRegister(acc.fields, fieldAccess.getName().getFieldType());
+                }
+                return super.visitFieldAccess(fieldAccess, ctx);
+            }
+
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
```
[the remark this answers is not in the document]
Laurens-W: I think it's fine to leave it like this in this case, I usually use those short names when I want to reassign the initial value because it's going to potentially be modified by the visitor :)

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+                    getAnnotations(type).forEach(annotation -> {
+                        if ("VisibleForTesting".equals(annotation.getClassName())) {
+                            J.CompilationUnit compilationUnit = getCursor().firstEnclosing(J.CompilationUnit.class);
+                            if (compilationUnit != null) {
+                                compilationUnit
+                                        .getMarkers()
+                                        .findFirst(JavaSourceSet.class)
+                                        .filter(elem -> "main".equals(elem.getName()))
+                                        .ifPresent(sourceSet -> target.add(TypeUtils.toString(type)));
+                            }
```
jevanlingen, line 108: I think I would write it more like:

```suggestion
                            if (cu != null &&
                                   cu.getMarkers().findFirst(JavaSourceSet.class).filter(s -> s.getName().toLowerCase(Locale.ROOT).contains("main")).isPresent()) {
                                   target.add(TypeUtils.toString(type))
                            }
```

JohannisK: Can you elaborate on this change?
s -> s.getName().toLowerCase(Locale.ROOT).contains("main")

It seems there are several considerations when trying to assert a file is in the production code?

jevanlingen (approved):

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
jevanlingen: I don't know, but I always try to find an existing example. Something you did here is also used in:

https://github.com/openrewrite/rewrite/blob/c08ea6273755822643bdb6d95e8350b50abf5be3/rewrite-java/src/main/java/org/openrewrite/java/search/IsLikelyTest.java#L65

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+                if (nc.getClazz() != null && nc.getClazz().getType() instanceof JavaType.Class) {
+                    checkAndRegister(acc.classes, nc.getClazz().getType());
+                }
+                return super.visitNewClass(nc, ctx);
+            }
+
+            private void checkAndRegister(List<String> target, JavaType type) {
+                if (!target.contains(TypeUtils.toString(type))) {
+                    getAnnotations(type).forEach(annotation -> {
+                        if ("VisibleForTesting".equals(annotation.getClassName())) {
```
timtebeek, line 103: Try to use `service(AnnotationService.class)` and the `matches` method that as, as opposed to retrieving annotation manually and matching them by their name.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+                    checkAndRegister(acc.methods, nc.getConstructorType());
+                }
+                if (nc.getClazz() != null && nc.getClazz().getType() instanceof JavaType.Class) {
+                    checkAndRegister(acc.classes, nc.getClazz().getType());
+                }
+                return super.visitNewClass(nc, ctx);
+            }
+
+            private void checkAndRegister(List<String> target, JavaType type) {
+                if (!target.contains(TypeUtils.toString(type))) {
```
timtebeek, line 101: If we're doing these `contains` checks, would it then not make more sense to use a `Set`?

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Override
+    public Scanned getInitialValue(ExecutionContext ctx) {
+        return new Scanned();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Scanned acc) {
+        return new JavaIsoVisitor<ExecutionContext>() {
```
timtebeek, line 60: It looks like we're only interested here in `src/main` right? Best then to use a precondition here for `IsLikelyNotTest`.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+                    return ((JavaType.Method) type).getAnnotations();
+                }
+                return Collections.emptyList();
+            }
+        };
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor(Scanned acc) {
+        return new JavaIsoVisitor<ExecutionContext>() {
```
timtebeek, line 129: Similarly here for better performance we can add combine two preconditions:
1. IsLikelyNotTest
2. UsesType VisibleForTest

That way we can cheaply skip visiting LST elements for a lot of compilation units, due to the cache used types we already maintain.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+import org.openrewrite.java.marker.JavaSourceSet;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
+
+import java.util.*;
+
+public class RemoveVisibleForTestingAnnotationWhenUsedInProduction extends ScanningRecipe<RemoveVisibleForTestingAnnotationWhenUsedInProduction.Scanned> {
+
+    public static class Scanned {
```
timtebeek, line 32: Can we be more expressive with the naming here? `Scanned` is a bit generic. Are these perhaps `VisibleForTesting`?

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,176 @@
...
+        List<String> methods = new ArrayList<>();
+        List<String> fields = new ArrayList<>();
+        List<String> classes = new ArrayList<>();
```
timtebeek, line 35: Similarly here let's be explicit about what exactly is captured; are these `methodPatterns` perhaps? Or some other qualifier that gives me a hint at how to read and use these String values.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
JohannisK: Yes

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
JohannisK: Thnx! I did :)

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
JohannisK: That is not possible her, because the cursor is at the invocation site, not de declaration site.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
JohannisK: UseType<>("*VisibleForTesting") does not work.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,178 @@
...
+                }
+                if (nc.getClazz() != null && nc.getClazz().getType() instanceof JavaType.Class) {
+                    checkAndRegister(acc.classPatterns, nc.getClazz().getType());
+                }
+                return super.visitNewClass(nc, ctx);
+            }
+
+            private void checkAndRegister(Set<String> target, JavaType type) {
+                if (!target.contains(TypeUtils.toString(type))) {
+                    getAnnotations(type).forEach(annotation -> {
```
JohannisK, line 105: I couldn't use `service(AnnotationService.class).matches(getCursor(), "*.VisibleForTesting")" here.
Also FQN did not work.

src/main/java/org/openrewrite/java/testing/cleanup/RemoveVisibleForTestingAnnotationWhenUsedInProduction.java
```diff
@@ -0,0 +1,175 @@
...
+                      .filter(elem -> "VisibleForTesting".equals(elem.getSimpleName()))
+                      .findFirst();
+                if (annotation.isPresent() && annotation.get().getType() instanceof JavaType.Class) {
+                    JavaType.Class type = (JavaType.Class) annotation.get().getType();
+                    return new RemoveAnnotation("@" + type.getFullyQualifiedName()).getVisitor().visitNonNull(target, ctx, getCursor().getParentOrThrow());
+                }
+                return target;
+            }
+        };
+        return Preconditions.check(new IsLikelyNotTest().getVisitor(), visitor);
```
JohannisK, line 173: UseType<>() is not working here. It does work with the FQN, but not with "*.VisibleForTesting"

timtebeek: That might be a limitation in openrewrite/rewrite; ideally we'd detect the use of that annotation on the type, especially when we include implicit usages, but it appears we do not for now. Could require a fix upstream, but not pressing for now.

timtebeek:
As discussed we're still seeing performance issues, likely because we're having to visit & determine the type of all references to classes, methods and fields to see if their target is annotated with `VisibleForTesting`. That's because we can not currently use any narrower precondition in our scanning visitor:
https://github.com/openrewrite/rewrite-testing-frameworks/pull/689/files#diff-7ad8934e882d4b7ce7b095083167e188aa0bb1b228249f8e76afab4da7e876d2R129

Perhaps it'd be an option to expand `UsesType` such that [when`includeImplicit = True`](https://github.com/openrewrite/rewrite/blob/6144a9542b1cd0771675939475b7569d35d5b0ef/rewrite-java/src/main/java/org/openrewrite/java/search/UsesType.java#L50-L51), that it then also matches compilation units where references to targets annotated with the `typePattern`, but that would require changes to the parser to prepopulate [TypesInUse](https://github.com/openrewrite/rewrite/blob/6144a9542b1cd0771675939475b7569d35d5b0ef/rewrite-java/src/main/java/org/openrewrite/java/internal/TypesInUse.java#L36-L41), which could shift performance impact to the parser instead.

PR #690, opened 2025-04-03
Replace AssertJ's `hasSize` with `hasSameSizeAs` where applicable

arodionov:
Creating a new recipe for simplifying `hasSize` assertion.
Where the `hasSize` assertion has an argument with `*.size()/length` explicit call, replacing this with `hasSameSizeAs`

Uncommented test

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/SimplifyHasSizeAssertion.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/SimplifyHasSizeAssertion.java
@@ -30,7 +30,6 @@ import java.util.Collections;
 
 public class SimplifyHasSizeAssertion extends Recipe {
 
-    private static final MethodMatcher ASSERT_THAT_MATCHER = new MethodMatcher("org.assertj.core.api.Assertions assertThat(..)");
     private static final MethodMatcher HAS_SIZE_MATCHER = new MethodMatcher("org.assertj.core.api.* hasSize(int)");
 
     private static final MethodMatcher CHAR_SEQUENCE_LENGTH_MATCHER = new MethodMatcher("java.lang.CharSequence length()", true);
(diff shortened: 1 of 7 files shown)


src/main/java/org/openrewrite/java/testing/assertj/SimplifyHasSizeJAssertion.java
```diff
@@ -0,0 +1,98 @@
...
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+
+import java.util.Collections;
+
+public class SimplifyHasSizeJAssertion extends Recipe {
+
+    public static final String ASSERT_THAT = "org.assertj.core.api.Assertions assertThat(..)";
```
jevanlingen, line 34: The public fields here can be private as well right?

src/main/java/org/openrewrite/java/testing/assertj/SimplifyHasSizeJAssertion.java
```diff
@@ -0,0 +1,98 @@
...
+                        if (expression instanceof J.MethodInvocation) {
+                            Expression argument = ((J.MethodInvocation) expression).getSelect();
+                            JavaType type = argument.getType();
+
+                            if (CHAR_SEQUENCE_TYPE_MATCHER.matches(type) ||
+                                    ITERABLE_TYPE_MATCHER.matches(type) ||
+                                    MAP_TYPE_MATCHER.matches(type)) {
+                                mi = mi.withMethodType(mi.getMethodType().withName(HAS_SAME_SIZE_AS))
+                                        .withName(mi.getName().withSimpleName(HAS_SAME_SIZE_AS))
+                                        .withArguments(Collections.singletonList(argument));
```
jevanlingen, line 81: Same code as line 89 till 91. Use a method to remove code duplication.

jevanlingen (approved):

src/main/java/org/openrewrite/java/testing/assertj/SimplifyHasSizeAssertion.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class SimplifyHasSizeAssertion extends Recipe {
+
+    private static final MethodMatcher ASSERT_THAT_MATCHER = new MethodMatcher("org.assertj.core.api.Assertions assertThat(..)");
+    private static final MethodMatcher HAS_SIZE_MATCHER = new MethodMatcher("org.assertj.core.api.* hasSize(int)");
+    private static final String HAS_SAME_SIZE_AS = "hasSameSizeAs";
+
+    private static final TypeMatcher CHAR_SEQUENCE_TYPE_MATCHER = new TypeMatcher("java.lang.CharSequence", true);
+    private static final TypeMatcher ITERABLE_TYPE_MATCHER = new TypeMatcher("java.lang.Iterable", true);
+    private static final TypeMatcher MAP_TYPE_MATCHER = new TypeMatcher("java.util.Map", true);
```
timtebeek, line 40: Rather than matching the type of the argument, I think it would be better to match the methods we're after here:
`CharSequence#length()`
`Iterable#size()`
`Map#size()`

That ought to then fix the failing test I'd just added, and remove the need for the null check on the expression select.

timtebeek (changes requested):
Found a bit of an edge case, but figured good to be specific about what we're matching, as now we match any method called on a CharSequence, Iterable or Map.

src/main/java/org/openrewrite/java/testing/assertj/SimplifyHasSizeAssertion.java
arodionov: Good catch!
Thanks for the suggestions!

timtebeek (approved):
Thanks again here @arodionov ! I've pushed some more minor polishing commits to your branch, and with that we're good to merge. Great to continue to improve the assertions for AssertJ.

PR #691, opened 2025-04-04
Recipe to replace @Rule Timeout to JUnit 5 Timeout annotation

arodionov:
Created a new recipe for migrating `@Rule Timeout` test class variable to JUnit 5 `@Timeout` class annotation

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.testing.junit5;
 
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
@@ -55,7 +56,7 @@ public class TimeoutRuleToClassAnnotation extends Recipe {
             public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
                 J.ClassDeclaration cd = super.visitClassDeclaration(classDecl, ctx);
 
-                AtomicReference<Expression> initializer = new AtomicReference<>();
+                AtomicReference<@Nullable Expression> initializer = new AtomicReference<>();
 
                 cd = cd.withBody(cd.getBody().withStatements(ListUtils.map(cd.getBody().getStatements(), statement -> {
                     if (statement instanceof J.VariableDeclarations) {
@@ -76,30 +77,31 @@ public class TimeoutRuleToClassAnnotation extends Recipe {
                     return statement;
                 })));
 
-                if (initializer.get() != null) {
+                Expression initializerValue = initializer.get();
+                if (initializerValue != null) {
                     maybeRemoveImport("org.junit.Rule");
                     maybeRemoveImport("org.junit.rules.Timeout");
-                    return insertTimeoutAnnotation(initializer.get(), cd, ctx);
+                    return insertTimeoutAnnotation(initializerValue, cd, ctx);
                 }
                 return cd;
             }
 
             private J.ClassDeclaration insertTimeoutAnnotation(Expression ex, J.ClassDeclaration cd, ExecutionContext ctx) {
-                JavaTemplate.Builder builder;
+                String template;
                 Object[] params;
                 if (ex instanceof J.NewClass) {
                     List<Expression> arguments = ((J.NewClass) ex).getArguments();
                     if (arguments.size() == 2) {
-                        builder = JavaTemplate.builder("@Timeout(value = #{any(long)}, unit = #{any(TimeUnit)})");
+                        template = "@Timeout(value = #{any(long)}, unit = #{any(TimeUnit)})";
                         params = new Object[]{arguments.get(0), arguments.get(1)};
                     } else {
-                        builder = JavaTemplate.builder("@Timeout(value = #{any(long)}, unit = TimeUnit.MILLISECONDS)");
+                        template = "@Timeout(value = #{any(long)}, unit = TimeUnit.MILLISECONDS)";
                         params = new Object[]{arguments.get(0)};
                     }
                 } else if (ex instanceof J.MethodInvocation) {
                     String simpleName = ((J.MethodInvocation) ex).getSimpleName();
                     String units = simpleName.equals("millis") ? "MILLISECONDS" : "SECONDS";
-                    builder = JavaTemplate.builder("@Timeout(value = #{any(long)}, unit = TimeUnit." + units + ")");
+                    template = "@Timeout(value = #{any(long)}, unit = TimeUnit." + units + ")";
                     params = new Object[]{((J.MethodInvocation) ex).getArguments().get(0)};
                 } else {
                     return cd;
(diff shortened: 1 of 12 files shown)


src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+        private JavaParser.@Nullable Builder<?, ?> javaParser;
+
+        private JavaParser.Builder<?, ?> javaParser(ExecutionContext ctx) {
+            if (javaParser == null) {
+                javaParser = JavaParser.fromJavaVersion()
+                        .classpathFromResources(ctx, "junit-jupiter-api-5", "hamcrest-3");
+            }
+            return javaParser;
+
+        }
```
timtebeek, line 64: This is actually an outdated pattern; we try not to use this anymore, as we've since added caching for JavaTemplates such that there's a reduced need for caching the builder. I'll phase this out as you likely didn't know when looking at a similar recipe. Perhaps you can do the same wherever you saw this copied from?

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+    public String getDescription() {
+        return "Replace usages of JUnit 4's `@Rule Timeout` with JUnit 5 `Timeout` class annotation.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("org.junit.rules.Timeout", false), new TimeoutRuleToClassAnnotationVisitor());
+    }
+
+    public static class TimeoutRuleToClassAnnotationVisitor extends JavaIsoVisitor<ExecutionContext> {
```
timtebeek, line 53: This is another outdated pattern; when there's no need for a named public reusable visitor, then it's fine to just use an anonymous visitor. Again best to apply the same wherever you copied this pattern from.

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replace usages of JUnit 4's `@Rule Timeout` with JUnit 5 `Timeout` class annotation.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("org.junit.rules.Timeout", false), new TimeoutRuleToClassAnnotationVisitor());
```
timtebeek, line 50: Great that you're using preconditions here; those help speed up the recipe run through the cached TypesInUse.

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+        @Override
+        public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
+            J.ClassDeclaration cd = super.visitClassDeclaration(classDecl, ctx);
+
+            AtomicReference<Expression> initializer = new AtomicReference<>();
+
+            cd = cd.withBody(cd.getBody().withStatements(ListUtils.map(cd.getBody().getStatements(), statement -> {
+                if (statement instanceof J.VariableDeclarations) {
+                    //noinspection ConstantConditions
+                    if (TypeUtils.isOfClassType(((J.VariableDeclarations) statement).getTypeExpression().getType(),
```
timtebeek, line 75: I noticed `org.junit.rules.Timeout` is not `final`; would we want to support custom rules that extend `Timeout`? And is that supported by `isOfClassType`? Or would we need a sibling like `isAssignableTo`?

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+                        return null;
+                    }
+                }
+                return statement;
+            })));
+
+            if (initializer.get() != null) {
+                cd = insertTimeoutAnnotation(initializer.get(), cd, ctx);
+                maybeRemoveImport("org.junit.Rule");
+                maybeRemoveImport("org.junit.rules.Timeout");
```
timtebeek, line 90: Where possible I prefer a direct return as opposed to repeated assignments to the same variable.
```suggestion
                maybeRemoveImport("org.junit.Rule");
                maybeRemoveImport("org.junit.rules.Timeout");
                return insertTimeoutAnnotation(initializer.get(), cd, ctx);
```

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+                    .imports("org.junit.jupiter.api.Timeout", "java.util.concurrent.TimeUnit")
+                    .build()
+                    .apply(
+                            updateCursor(cd),
+                            cd.getCoordinates().addAnnotation(Comparator.comparing(J.Annotation::getSimpleName)),
+                            params
+                    );
+            maybeAddImport("org.junit.jupiter.api.Timeout");
+            maybeAddImport("java.util.concurrent.TimeUnit");
+            return cd;
```
timtebeek, line 126:
```suggestion
                maybeAddImport("org.junit.jupiter.api.Timeout");
                maybeAddImport("java.util.concurrent.TimeUnit");
                return builder.javaParser(JavaParser.fromJavaVersion()
                                .classpathFromResources(ctx, "junit-jupiter-api-5", "hamcrest-3"))
                        .imports("org.junit.jupiter.api.Timeout", "java.util.concurrent.TimeUnit")
                        .build()
                        .apply(
                                updateCursor(cd),
                                cd.getCoordinates().addAnnotation(Comparator.comparing(J.Annotation::getSimpleName)),
                                params
                        );
```

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
```diff
@@ -0,0 +1,129 @@
...
+                    builder = JavaTemplate.builder("@Timeout(value = #{any(long)}, unit = TimeUnit.MILLISECONDS)");
+                    params = new Object[]{arguments.get(0)};
+                }
+            } else if (ex instanceof J.MethodInvocation) {
+                String simpleName = ((J.MethodInvocation) ex).getName().getSimpleName();
+                String units = simpleName.equals("millis") ? "MILLISECONDS" : "SECONDS";
+                builder = JavaTemplate.builder("@Timeout(value = #{any(long)}, unit = TimeUnit." + units + ")");
+                params = new Object[]{((J.MethodInvocation) ex).getArguments().get(0)};
+            } else {
+                return cd;
```
timtebeek, line 113: In this case we've already removed the rule above right? So for any unhandled case we'd be removing the rule without a  replacement. That to me indicates we should have similar restrictions when removing the rule as we do for adding the annotation.

timtebeek (changes requested):

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
arodionov: Yes, I've seen this at https://github.com/openrewrite/rewrite-testing-frameworks/blob/main/src/main/java/org/openrewrite/java/testing/junit5/UpdateTestAnnotation.java

src/main/java/org/openrewrite/java/testing/junit5/TimeoutRuleToClassAnnotation.java
arodionov: I prefer not to consider this case since we don't know what the initialiser can look like.

timtebeek (approved):
Good to see this added, and the old patterns cleared out.

You might want to go over the last polishing commits I did such that you can apply the same to PRs still to follow, thanks!

Issue #512, opened 2024-04-25
Mockito: Recipe to replace initMocks with openMocks


lucashan:
## What problem are you trying to solve?
When upgrading Java applications and Mockito version, there is a MockitoAnnotiations API (`initMocks`) that is deprecated (see [mockito-core 3.4.0 documentation](https://www.javadoc.io/doc/org.mockito/mockito-core/3.4.0/org/mockito/MockitoAnnotations.html) for details).

## What precondition(s) should be checked before applying this recipe?
The recommendation is to replace `initMocks` with `openMocks(Object)`, which is equivalent to `openMocks(testClass).close()`. The close method should however only be called after completed usage of testClass. If using static-mocks or custom [MockMaker](https://www.javadoc.io/static/org.mockito/mockito-core/3.4.0/org/mockito/plugins/MockMaker.html)s, using this method might cause misbehavior of mocks injected into the test class. 

## Have you considered any alternatives or workarounds?
N/A

## Any additional context
I'm not entirely sure if OpenRewrite has an existing utilities/mechanisms to simplify such complex refactorings, I wanted to surface this issue to check if there are any. 

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?
N/A

timtebeek:
Thanks for the suggestion @lucashan ! We should be able to convert such cases as described with a new recipe. We already have a similar recipe test to bring folks from JUnit 4 and Mockito 1 to JUnit 5 and Mockito v3; seems only logical to automate the next migration test involving `initMocks` / `openMocks` as well.

Ideally we start with a unit test similar to this one, describing exactly what we should aim for with this recipe.
https://github.com/openrewrite/rewrite-testing-frameworks/blob/fafcb85d7c8c78ca3debfc563d91f55364067957/src/test/java/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.java#L46-L122

arodionov:
I see two cases (recipes) here:
- remove `MockitoAnnotations.initMocks(this)` completely if there specified some runners `@RunWith(MockitoJUnitRunner.class)` /  `@ExtendWith(MockitoExtension.class)`
- replace  `MockitoAnnotations.initMocks(this)` =>  `mocks = MockitoAnnotations.openMocks(this)` and add a method with cleanup `mocks.close();`

PR #695, opened 2025-04-08
Recipe to remove initMocks if specified JUnit runners

arodionov:
Add a recipe to remove `MockitoAnnotations.initMocks(this)` (and surrounding method if empty) if specified class-level JUnit runners `@RunWith(MockitoJUnitRunner.class)` or `@ExtendWith(MockitoExtension.class)`.

- https://github.com/openrewrite/rewrite-testing-frameworks/issues/512


arodionov:
I think this recipe can be used wider, not only during migration from Mockito from 1.x to 5.x. but also for test "cleaning" purposes.

This recipe can also be extended (or created a separate recipe) to remove 

```
private AutoCloseable mocks;

    @Before
    public void setUp() {
        mocks = MockitoAnnotations.openMocks(this);
    }

    @After
    public void tearDown() throws Exception {
        mocks.close(); 
    }
```

If a class-level JUnit runners are used.

src/main/java/org/openrewrite/java/testing/mockito/RemoveInitMocksIfRunnersSpecified.java
```diff
@@ -0,0 +1,89 @@
...
+                            if (md.getBody() != null && md.getBody().getStatements().isEmpty()) {
+                                return null;
+                            }
+                        }
+                        return md;
+                    }
+
+                    @Override
+                    public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration cd, ExecutionContext ctx) {
+                        for (J.Annotation annotation : cd.getLeadingAnnotations()) {
```
timtebeek, line 79: Instead of only looking at the leading annotations, try to use `service(AnnotationService.class)` to check annotations.

arodionov: updated

timtebeek (approved):
Great to see! I like how you cover both JUnit 4 and 5, and take care to clean up the setup methods only when empty.

I've pushed a very small change: cursor messaging wasn't necessary here as we can determine if a `MethodInvocation` was removed from a `MethodDeclaration` by comparing the response from `super.visitMethodDeclaration` with the original `methodDeclaration`. That's a common (and cheap!) comparision we do, and avoids unneeded complexity.

Cursor messaging we use either when exchanging objects, or when multiple paths are possible downstream of a visit, and you need to be able to differentiate which was taken. That wasn't needed here, as we only remove a single type of method invocation.

PR #699, opened 2025-04-10
Convert JUnit AssertThrows with description message over to AssertJ

lewisbirks:
## What's changed?

Added support to the `JUnitAssertThrowsToAssertExceptionType` recipe for retaining descriptions.

## What's your motivation?

The `assertThrows` has overloads with 3 parameters, the last parameter being a message to display on failure, which is not currently supported.

I couldn't get the tests to work when I included the types within the `#{any()}` as was done within the existing code, the tests would then fail with an error of the form  
`/*~~(MethodInvocation type is missing or malformed)~~>*/assertThatExceptionOfType(NullPointerException.class).as("message").isThrownBy(() -> foo())`  
that I wasn't knowledgable enough to get to the bottom of.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.testing.assertj;
 
+import java.util.List;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Great to see, thanks a lot! Good to not leave these dangling JUnit assertions just because there was a message on the assertion.

I've cleaned up a bit of the type wrangling that was there previously, which helped simplify the implementation even more.

PR #700, opened 2025-04-10
Recipe to replace `initMocks` with `openMocks`

arodionov:
If the test class contains mocks initialization 

```java
@BeforeEach
public void setUp() {
       // ...
       MockitoAnnotations.initMocks(this);
       // ...
}
```

it should be replaced and extended in the following way:

```java
private AutoCloseable mocks;

@BeforeEach
public void setUp() {
       // ...
       mocks = MockitoAnnotations.openMocks(this);
       // ...
}

@AfterEach
public void tearDown() throws Exception {
       // ...
       mocks.close();
       // ...
}

```

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMock.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMock.java
@@ -64,11 +64,8 @@ public class ReplaceInitMockToOpenMock extends Recipe {
                         J.ClassDeclaration cd = super.visitClassDeclaration(classDecl, ctx);
                         if (getCursor().getMessage("initMocksFound", false)) {
                             variableName = generateVariableName("mocks", getCursor(), INCREMENT_NUMBER);
-                            J.ClassDeclaration after = JavaTemplate.builder("private AutoCloseable " + variableName + ";")
-                                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx))
-                                    .build()
-                                    .apply(getCursor(), cd.getBody().getCoordinates().firstStatement());
-
+                            J.ClassDeclaration after = JavaTemplate.apply("private AutoCloseable " + variableName + ";",
+                                    getCursor(), cd.getBody().getCoordinates().firstStatement());
                             return maybeAutoFormat(cd, after, ctx);
                         }
                         return cd;
--- a/src/main/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMock.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMock.java
@@ -16,7 +16,6 @@
 package org.openrewrite.java.testing.mockito;
 
 import org.openrewrite.*;
-import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.*;
 import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.search.UsesType;
@@ -24,8 +23,8 @@ import org.openrewrite.java.service.AnnotationService;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.Statement;
 
-import java.util.List;
-import java.util.Objects;
+import static org.openrewrite.java.VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER;
+import static org.openrewrite.java.VariableNameUtils.generateVariableName;
 
 public class ReplaceInitMockToOpenMock extends Recipe {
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.java
```diff
@@ -51,7 +51,7 @@ public void defaults(RecipeSpec spec) {
     void replaceMockAnnotation() {
         //language=java
         rewriteRun(
-          spec -> spec.typeValidationOptions(TypeValidation.none()),
+          spec -> spec.typeValidationOptions(TypeValidation.none()).expectedCyclesThatMakeChanges(2),
```
Laurens-W, line 54: Could we look at reducing this back to a single cycle? In the future we will disallow multiple cycles so best to not introduce these expectations.
Usually this behavior occurs when the preconditions are not strict enough, so those are worth looking into first!

src/test/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMockTest.java
```diff
@@ -0,0 +1,409 @@
...
+                      mocks.close();
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void mocksVariableIsAlreadyPresent() {
```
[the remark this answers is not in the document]
arodionov: I'd like to test that we consider the case if the `mocks` variable is already present in the method scope.

src/test/java/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.java
arodionov: Thanks for your comments!

If it is a migration recipe consisting of many recipes, should it also take one cycle?

Laurens-W: Preferably yes :)

timtebeek: So what's likely happening here is that we run the Mockito recipes before we update the before/after test annotations:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/55495a87cd1638ffb7de38147c4ce84f4557844e/src/main/resources/META-INF/rewrite/junit5.yml#L77-L89

And then the recipe you've written has a precondition that expects JUnit 5 annotations:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/f61e3c5f0e4e4f6fbc5a6815717849765417213e/src/main/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMock.java#L52-L59

So the first time the `ReplaceInitMockToOpenMock` recipe runs, the precondition does not match, but after the later switch of JUnit before/after annotations to beforeEach/afterEach mean that the second iteration _does_ have it's precondition matched.

We can likely fix this issue by either adjusting the preconditions, or by adjusting the order in which the migration recipes run.

src/main/java/org/openrewrite/java/testing/mockito/ReplaceInitMockToOpenMock.java
```diff
@@ -0,0 +1,149 @@
...
+                                maybeRemoveImport("org.mockito.MockitoAnnotations.initMocks");
+                                for (Statement st : md.getBody().getStatements()) {
+                                    if (st instanceof J.MethodInvocation && INIT_MOCKS_MATCHER.matches((J.MethodInvocation) st)) {
+                                        return JavaTemplate.builder(variableName + " = MockitoAnnotations.openMocks(this);")
+                                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "mockito-core"))
+                                                .imports("org.mockito.MockitoAnnotations")
+                                                .contextSensitive()
+                                                .build()
+                                                .apply(getCursor(), st.getCoordinates().replace());
+                                    }
```
timtebeek, line 125: When stepping through with the debugger I saw the second edit cycle had the same printed output, but a different `initMocks` instance (_not_ `openMocks`) on these lines: https://github.com/openrewrite/rewrite/blob/9569cfb6b7c88d015ba1369fa5639bcf23db4dc0/rewrite-test/src/main/java/org/openrewrite/test/LargeSourceSetCheckingExpectedCycles.java#L91-L98

I suspect this JavaTemplate has something to do with that, as the `st.getCoordinates().replace()` when visiting the method declaration is a bit suspect.

src/test/java/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.java
arodionov: It seems that the issue happens because different recipes use different Mockito versions 
`JunitMockitoUpgradeIntegrationTest` uses `mockito-all-1.10` 
If I change it to `mockito-core`, the recipe takes 1 cycle.

timtebeek (approved):
Great to have this resolved! Nice find on that version mismatch for the final failing test.

PR #702, opened 2025-04-11
Convert JUnit AssertThrows with assigment over to AssertJ

lewisbirks:
## What's changed?

Allow for JUnit AssertThrows assertions that use the returned value to be converted to the AssertJ equivalent with AssertThatExceptionOfType.

## What's your motivation?

When this conversion was causing an issue, as resolved by #331, the decision was made to ignore conversions of assertions that used the return value due to no functional equivalent being available in AssertJ.  
Since then AssertJ has released [v3.27.0](https://github.com/assertj/assertj/releases/tag/assertj-build-3.27.0) which has introduced the [`actual`](https://www.javadoc.io/doc/org.assertj/assertj-core/3.27.0/org/assertj/core/api/AbstractAssert.html#actual()) method which allows the retrieval of the actual value. This can be used to ensure that functionality can be replicated, although ideally any subsequent assertions of the exception would be using chained assertions from the initial `assertThatExceptionOfType(X.class).isThrownBy(() -> Y())`.

As far as I can tell the [latest version of AssertJ](https://github.com/openrewrite/rewrite-testing-frameworks/blob/b36584a2d6f47d2cdbd3bfc71c314dd14c541d39/src/main/resources/META-INF/rewrite/assertj.yml#L483) is pulled so this method would be available but I'm happy to be corrected on this.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
@@ -28,6 +28,7 @@ import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 
 import java.util.List;
+import java.util.function.UnaryOperator;
 
 public class JUnitAssertThrowsToAssertExceptionType extends Recipe {
 
@@ -48,6 +49,7 @@ public class JUnitAssertThrowsToAssertExceptionType extends Recipe {
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
         return Preconditions.check(new UsesMethod<>(ASSERT_THROWS_MATCHER), new JavaIsoVisitor<ExecutionContext>() {
+
             @Override
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionTypeTest.java
```diff
@@ -124,6 +122,17 @@ public void throwsExceptionWithSpecificType() {
...
               }
+              """,
+            """
+              import static org.assertj.core.api.AssertionsForClassTypes.assertThatExceptionOfType;
+
+              public class SimpleExpectedExceptionTest {
+                  public void throwsExceptionWithSpecificType() {
+                      NullPointerException npe = assertThatExceptionOfType(NullPointerException.class).isThrownBy(() -> {
+                          throw new NullPointerException();
+                      }).actual();
```
timtebeek, line 133: Neat! Didn't know `actual()` also worked for asserted exceptions, but it's definitely nice to see & have.

lewisbirks: me neither until it popped up in the IDE autocomplete 😅

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
```diff
@@ -48,31 +49,33 @@ public String getDescription() {
...
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                 J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
                 if (!ASSERT_THROWS_MATCHER.matches(mi)) {
                     return mi;
                 }
 
-                if (!(getCursor().getParentTreeCursor().getValue() instanceof J.Block)) {
-                    return mi;
-                }
+                boolean returnActual = !(getCursor().getParentTreeCursor().getValue() instanceof J.Block);
```
timtebeek, line 60: I wonder if we might over eagerly add the `actual()` here in cases such as `() -> assertThrows(...)...`. There the parent tree cursor would also not be a block, but we might not use the returned value. Would it make sense here to limit ourselves initially to J.Block and `J.Assignment`?

We can always separately look to also cover cases where we might be uses as a Consumer, or even Supplier.

src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
```diff
@@ -48,31 +49,33 @@ public String getDescription() {
...
 
                 maybeAddImport(ASSERTIONS_FOR_CLASS_TYPES, "assertThatExceptionOfType");
                 maybeRemoveImport(JUNIT_ASSERTIONS + ".assertThrows");
                 maybeRemoveImport(JUNIT_ASSERTIONS);
 
                 List<Expression> args = mi.getArguments();
+
+                UnaryOperator<String> decorator = returnActual ? code -> code + ".actual()" : UnaryOperator.identity();
```
timtebeek, line 68: Would it not be easier to use the following here?
```java
if (returnActual) {
  code += ".actual()";
}

**Review by timtebeek — COMMENTED** [review]
Great to see how you're expanding support here, thanks a lot! Couple quick questions before we do a more thorough review and merge.

File: src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java (hunk 2 above)
Comment (lewisbirks) [reply src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java:60]: I can have a play around with this, I was just keeping the same condition that we were using to break before

File: src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java (hunk 3 above)
Comment (lewisbirks) [reply src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java:68]: Probably yeah 😅

File: src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java (hunk 2 above)
Comment (timtebeek) [reply src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java:60]: Yes I figured as much; my reason to call it out now is that we didn't do anything previously, whereas now we might convert a lot more (potentially unexpected) cases by suffixing `.actual()`. Thanks for having a closer look!

Comment (lewisbirks) [reply src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java:60]: @timtebeek do you mean something like this or am I over thinking things?

```java
/**
 * Check if there is a return type which would indicate the need for using
 * {@code .actual()} in the AssertJ call.
 * <p>
 * If the presence of a return type could not be determined then {@code Optional.empty()} is returned 
 * and the current {@code J.MethodInvocation} should be used without further changes.
 *
 * @return {@code Optional.of(true)} if there is a return type otherwise {@code Optional.of(false)}.
 * If it could not be determined then {@code Optional.empty()}.
 */
private Optional<Boolean> hasReturnType() {
    Object parent = getCursor().getParentTreeCursor().getValue();

    if (parent instanceof J.Lambda) {
        // Not currently supported, need to figure out if the lambda has a return type
        return Optional.empty();
    }

    // These all have the method invocation return something
    if (parent instanceof J.Assignment
            || parent instanceof J.VariableDeclarations
            || parent instanceof J.VariableDeclarations.NamedVariable
            || parent instanceof J.Return
            || parent instanceof J.Ternary) {
        return Optional.of(true);
    }

    if (parent instanceof J.Block) {
        return Optional.of(false);
    }

    // Unknown parent type so not supported
    return Optional.empty();
}
```

timtebeek: That would help, yes! Looks like we could drop the first `if` as `J.Lambda` does not extend or implement any of those others.

timtebeek (approved):
Great to see! Thanks a lot for extending support here, and iterating on the solution after feedback.

PR #705, opened 2025-04-19
Add support to migrate EnvironmentVariables TestRule

amishra-u:
## What's changed?
Added support to migrate EnvironmentVariables TestRule to Junit5 Extension

### Anything in particular you'd like reviewers to focus on?
Not sure, why test passes when I execute in Isolation but fails with full test. Probably something do to with type table.
@timtebeek can you help here.

## Any additional context
This recipe migrates the use of EnvironmentVariables from the JUnit 4-specific System Rules library to the System Stubs library

### Checklist


build.gradle.kts
```diff
@@ -32,6 +33,8 @@ recipeDependencies {
     parserClasspath("org.testcontainers:testcontainers:1.20.6")
     parserClasspath("pl.pragmatists:JUnitParams:1.+")
     parserClasspath("org.easytesting:fest-assert-core:2.+")
+    parserClasspath("uk.org.webcompere:system-stubs-core:2.1.8")
+    parserClasspath("uk.org.webcompere:system-stubs-jupiter:2.1.8")
```
timtebeek, line 37: Did you consider this library instead? That is a bit of an incubator for new features in JUnit proper, and perhaps preferred over introducing a "foreign" library dependency.
https://junit-pioneer.org/docs/environment-variables/

src/main/resources/META-INF/rewrite/classpath.tsv.zip
timtebeek, line 1: This file was updated through `./gradlew createTypeTable`, and is preferred over packaged jars where possible.
See this page for more details: https://docs.openrewrite.org/authoring-recipes/multiple-versions#type-tables

timtebeek (commented):
Great start, thanks! One comment to potentially switch the direction of this PR.

build.gradle.kts
amishra-u: Yes, I checked other libraries. junit-pioneer is also a third party extension to [Junit5](https://github.com/junit-team/junit5/wiki/Third-party-Extensions)
1. JUnit Pioneer modifies system variables using reflection. However, starting with JDK 17, reflective access to JDK internals is restricted, and enabling it requires explicitly passing --add-opens JVM arguments.
2. System Stubs is kind of successor to System Rules and the api structure is very similar which simplifies the recipe greatly. System Stubs doesn't uses reflection instead use bytecode manipulation, so doesn't have the same problem
ref:
https://junit-pioneer.org/docs/environment-variables/#warnings-for-reflective-access
https://www.baeldung.com/java-system-stubs#7-jdk-compatibility

amishra-u:
@timtebeek fixed test failure.

src/main/java/org/openrewrite/java/testing/junit5/Junit4Utils.java
```diff
@@ -0,0 +1,89 @@
...
+            stmtAnnotations = ((J.ClassDeclaration) stmt).getLeadingAnnotations();
+        } else if (stmt instanceof J.MethodDeclaration) {
+            stmtAnnotations = ((J.MethodDeclaration) stmt).getLeadingAnnotations();
+        } else if (stmt instanceof J.VariableDeclarations) {
+            stmtAnnotations = ((J.VariableDeclarations) stmt).getLeadingAnnotations();
+        }
+        return stmtAnnotations.stream()
+                .filter(a -> a.getType() instanceof JavaType.Class)
+                .map(a -> (JavaType.Class) a.getType())
+                .anyMatch(t -> annotations.contains(t.getFullyQualifiedName()));
```
timtebeek, line 87: I wonder if using the `service(AnnotationService.class)` in the original caller might be the better option here, or even `Traits.annotated`. Right now this quite a bit further removed, and I think this convenience method might have missed the availability of those other two.

amishra-u: Liked trait concept. Update this one to use traits, removed lots of redundant code. Also used in my next pr https://github.com/openrewrite/rewrite-testing-frameworks/pull/709

Thanks for pointing out.

sambsnyd:
Thanks @amishra-u !

PR #707, opened 2025-04-19
Fix: Don't modify regular test to parameterized test incorrectly

amishra-u:
## What's changed?
Fixed a bug in the JUnitParamsRunnerToParameterized recipe that incorrectly converted regular tests to @ParameterizedTest if the test class contained any @ParameterizedTest annotations.

### Checklist


timtebeek (approved):
Looks great; thanks for your continued & expanded support here @amishra-u !

PR #708, opened 2025-04-19
Fix ParameterizedRunnerToParameterized recipe: Preserve field-dependent setup logic

amishra-u:
## What's changed?
In JUnit4, parameterized tests populate class fields before executing the @Before method, allowing setup logic to rely on those values. However, in JUnit5, @ParameterizedTest methods receive arguments directly, and the test instance is created before those values are available. As a result, there is no lifecycle hook where the arguments can be injected before @BeforeEach runs.

This mismatch breaks any setup logic that depends on parameter-initialized fields.

To preserve behavior consistent with JUnit4, this change manually invokes the setup method after setting the parameter values to the class fields inside the test method. This ensures any logic in @BeforeEach continues to function correctly with the parameter values in place.

### Possible improvement for future
Invoke the setup method if it references or depends on the parameter values. This avoids unnecessary calls when the setup logic is independent of the test parameters.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/ParameterizedRunnerToParameterizedTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/ParameterizedRunnerToParameterizedTest.java
@@ -235,6 +235,7 @@ class ParameterizedRunnerToParameterizedTest implements RewriteTest {
           spec -> spec.typeValidationOptions(TypeValidation.none()),
           java(
             """
+              import org.junit.Before;
               import org.junit.Test;
               import org.junit.runner.RunWith;
               import org.junit.runners.Parameterized;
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/junit5/ParameterizedRunnerToParameterizedTest.java
```diff
@@ -261,12 +262,16 @@ public static List<Object[]> parameters() {
                       return Arrays.asList(new Object[]{124, "Otis", "TheDog", Map.of("toys", "ball", "treats", "bacon")}, new Object[]{126, "Garfield", "TheBoss", Map.of("toys", "yarn", "treats", "fish")});
                   }
 
+                  @Before
+                  public void setUp() {}
```
timtebeek, line 266: I think indeed it would be good to use the variables in this method, and then only call `setUp` when indeed used. Otherwise we might get some surprising unnecessary calls to `@Before` methods. Do you think that's doable?

amishra-u: I'm not expecting this to cause any issues—it was just a nice-to-have, like only calling the method when it's truly needed. If you're okay with it, I'd prefer to handle that case in a separate PR.

timtebeek (approved):

PR #709, opened 2025-04-21
Implement MockitoJUnitRunner to Extension Recipe

amishra-u:
## What's changed?
Created a Recipe to migrate MockitoJUnitRunner to MockitoExtension. 

## What's your motivation?
The previous implementation had few issues:
1. It only handled lenient and warn modes, and incorrectly migrated warn mode to strict mode (while the default for MockitoJUnitRunner is warn, the default for MockitoExtension is strict).
2. Mode-specific handling was split between imperative (for lenient) and declarative (for warn) recipes.

### Changes
1. Consolidated all mode-handling logic (strict, warn, and lenient) into a single imperative recipe.
2. New recipe correctly maps warn mode to maintain expected behavior and also handled strict mode for MockitoJUnitRunner.
3. Marked the previous imperative recipe as deprecated for backward compatibility instead of removing it.

### Followup
I will create a followup PR to update the UseMockitoExtension declarative recipe to leverage this new recipe.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitRunnerSilentToExtension.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/MockitoJUnitRunnerSilentToExtension.java
@@ -30,6 +30,10 @@ import org.openrewrite.java.tree.J;
 import java.util.Collections;
 import java.util.Comparator;
 
+/**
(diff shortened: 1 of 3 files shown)


amishra-u:
@timtebeek please take a look. Hoping this to be part of next release.

sambsnyd:
Thanks @amishra-u

PR #711, opened 2025-04-23
Utilize MockitoJUnitRunnerToExtension for UseMockitoExtension

amishra-u:
## What's changed?
Updated UseMockitoExtension declarative recipe to use MockitoJUnitRunnerToExtension recipe and fixed test.

### Checklist


Jenson3210 (approved):
I validated that we did not lose test coverage of Silent -> Lenient by removing the `org.openrewrite.java.testing.mockito.MockitoJUnitRunnerSilentToExtension` recipe from the list as there is a test for this in `MockitoJUnitRunnerToExtension` also.

Other than that, happy to see a more generic recipe be used for the replacement of a deprecated more focused one.

Issue #703, opened 2025-04-12
`SimplifyChainedAssertJAssertion` should skip `Optional` values with wildcards


uhafner:
## What version of OpenRewrite are you using?

I am using 

    
    <rewrite-maven-plugin.version>6.4.0</rewrite-maven-plugin.version>
    <rewrite-testing-frameworks.version>3.6.1</rewrite-testing-frameworks.version>
    <rewrite-static-analysis.version>2.6.0</rewrite-static-analysis.version>
    <rewrite-migrate-java.version>3.7.0</rewrite-migrate-java.version>
    <rewrite-recommendations.version>1.15.0</rewrite-recommendations.version>

## How are you running OpenRewrite?

Maven, see   https://github.com/jenkinsci/analysis-pom-plugin/blob/307b0a997b11b90498cb744bffb191aa8b9e8ec2/pom.xml#L269-L315

## What is the smallest, simplest way to reproduce the problem?

```java
import org.junit.jupiter.api.Test;

import java.util.Optional;

import static org.assertj.core.api.Assertions.*;

class BugTest {
    @Test
    void shouldShowRewriteBug() {
        Integer integer = 42;
        Optional<? extends Integer> optional = Optional.of(integer);

        assertThat(optional.get()).isEqualTo(integer);
    }
}
```

## What did you expect to see?

In this case the expansion cannot be applied without compile errors, so it should be skipped. 

## What did you see instead?

This code with compile errors is generated:

```java
class BugTest {
    @Test
    void shouldShowRewriteBug() {
        Integer integer = 42;
        Optional<? extends Integer> optional = Optional.of(integer);

        assertThat(optional).contains(integer);
    }
}
```

## What is the full stack trace of any errors you encountered?

No Stacktrace, the generated code does not compile with Java 21.0.6

## Are you interested in [contributing a fix to OpenRewrite]

Not right now.

timtebeek:
[...]

PR #713, opened 2025-04-26
Include additional AssertJ rules/recipes from Error Prone Support

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/703


timtebeek:
This PR should help some of the Map cases that now fail to match again:
- https://github.com/PicnicSupermarket/error-prone-support/pull/1664

timtebeek (commented):
Not the complete replacement for a lot of declarative recipes that I had hoped for, but still some nice improvements that we can already apply by merging this increment.

PR #714, opened 2025-04-29
Fix Incorrect Migration of TemporaryFolder.newFile in TemporaryFolderToTmpDir Recipe

amishra-u:
## What's changed?
The TemporaryFolderToTmpDir recipe incorrectly migrates `temporaryFolder.newFile(fileName)` to `Files.createTempFile(fileName, null, temporaryFolder)`. This approach changes the behavior: instead of creating a file with the exact name fileName, it uses fileName as a prefix and appends random characters as a suffix.

To address this, introduced a new newFile method that correctly replicates the original behavior. This method creates a File instance with `new File(temporaryFolder, fileName)` which is identical to `temporaryFolder.newFile` method.

**Note**: I didn't touch the `temporaryFolder.newFile` without arguments as internally translates to `Files.createTempFile`

Additionally, the migration now updates all occurrences of temporaryFolder.newFile(fileName) to use the new newFile(fileName) method, preserving the original semantics accurately.
### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
@@ -15,7 +15,7 @@
  */
 package org.openrewrite.java.testing.junit5;
 
-import lombok.AllArgsConstructor;
+import lombok.RequiredArgsConstructor;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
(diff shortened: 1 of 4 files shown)


amishra-u:
@timtebeek can you review this one please.

Jenson3210 (approved):
Did some reading into the junit-5-temporary-directory documentation and cannot see why this change would be wrong. (Have never used this feature myself in junit4 nor 5 yet) 

I already like that 1 method gets reused for multiple usages in a single `ClassDeclaration`.  If this means that we have these new static methods popping up suddenly in all classes, there will be multiple opinions on that. My personal opinion: I like to keep tests clean / not impacting each other, so can bare a little repetition if it keeps the test looser coupled with other test / systems.

amishra-u:
@timtebeek Can you please merge, I need to create another change on top of this #720

PR #715, opened 2025-04-30
Fix Import of Mockito classes in MockitoRuleToExtension recipe

amishra-u:
## What's changed?
The MockitoJUnitToMockitoExtension recipe was not importing mockito-core as a resource, which caused missing type attribution for Mockito-related expressions generated by the Java template parser. Previously, tests passed because the maybeAddImport method was used with onlyReferenced = false, masking the issue.

To fix this, imported mockito-core for corresponding JavaTempalteParser.
To verify the fix, updated the maybeAddImport method to use onlyReferenced = true and confirmed that mockito related imports being added. 

### Checklist


timtebeek (approved):
Perfect, thanks!

PR #718, opened 2025-05-02
Fix ParameterizedRunnerToParameterized: Remove redundant super call

amishra-u:
## What's changed?
ParameterizedRunnerToParameterized recipe converts constructor to initMethod but it may have redundant super call, this fix removes that redundant super call.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/ParameterizedRunnerToParameterized.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/ParameterizedRunnerToParameterized.java
@@ -76,7 +76,7 @@ public class ParameterizedRunnerToParameterized extends Recipe {
 
                 // Constructor Injected Test
                 if (parametersMethodName != null && constructorParams != null && constructorParams.stream().anyMatch(org.openrewrite.java.tree.J.VariableDeclarations.class::isInstance)) {
-                    doAfterVisit(new ParameterizedRunnerToParameterizedTestsVisitor(classDecl, parametersMethodName, initMethodName, parametersAnnotationArguments, constructorParams, true, beforeMethodName,ctx));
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/junit5/ParameterizedRunnerToParameterized.java
```diff
@@ -386,5 +386,22 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
+
+        @Override
+        public J.@Nullable MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+            J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
+            J.MethodDeclaration enclosingMethod = getCursor().firstEnclosing(J.MethodDeclaration.class);
+            // remove redundant super call, otherwise it will create compilation error when this constructor is converted to regular method.
+            // TODO: Handle super call with arguments.
+            if (enclosingMethod != null && enclosingMethod.isConstructor() && mi.getName().getSimpleName().equals("super") && isEmptyArgs(mi)) {
+                return null;
+            }
```
timtebeek, line 398: [Elsewhere](https://github.com/openrewrite/rewrite-spring/blob/70153755a1077235c1064e01b0703898ce0a650d/src/main/java/org/openrewrite/java/spring/batch/ReplaceSupportClassWithItsInterface.java#L97-L110) we've solved this slightly differently. Not saying either is better, but figured call it out here since you had a todo item attached, and the MethodMatcher approach could match those too. But perhaps you meant to handle any side effects of a `super(args..)`, rather than just removing those.

timtebeek (approved):
Made a slight change that removed an additional `super` call; hope you agree!

PR #719, opened 2025-05-03
Fix exception handling in ReplaceInitMockito recipe

amishra-u:
## What's changed?
When ReplaceInitMockToOpenMock recipe add `mocks.close()` to a method it doesn't add throws declaration if it was missing due to which compilation fails after migration.

Updated recipe to handle this case. 

### Checklist


timtebeek (commented):
Perfect addition, thanks a lot! 

/cc @arodionov

PR #721, opened 2025-05-08
ExpectedExceptionToAssertThrows: Handle branching and var usage in except methods

amishra-u:
## What's Changed?

Previously, the `ExpectedExceptionToAssertThrows` recipe failed to correctly handle:

- **Conditional exception registration** (commonly seen in parameterized tests)
- **Variable usage in `expectMessage` arguments**

When variables declared before the `thrown.expect(...)` calls were referenced in assertions, results in **uncompilable code** after transformation.

### Problem Example (Before)
Input:
```java
void test() {
    int[] a = new int[] { 1 };
    thrown.expect(IndexOutOfBoundsException.class);
    thrown.expectMessage("Index 1 out of bounds for length 1");
    thrown.expectMessage("Index 1 out of bounds for length " + a.length);
    int b = a[1];
}

```
### Output (broken):
```java
void test() {
    Throwable exception = assertThrows(IndexOutOfBoundsException.class, () -> {
        int[] a = new int[]{1};
        int b = a[1];
    });
    assertTrue(exception.getMessage().contains("Index 1 out of bounds for length 1"));
    assertThat(exception.getMessage(), containsString("Index 1 out of bounds for length " + a.length)); //  'a' is undefined
}
````
The updated recipe now wraps only the statements that appear after the thrown.expect call. Such that all the variables declared will be available in assert call scope.

### Conditional Exception Handling

The updated recipe now also supports conditionally triggered exceptions, which were previously unhandled. These often appear in parameterized tests.

**1. If-Else Conditional:**
```java
if (shouldThrow) {
    this.thrown.expect(IllegalArgumentException.class);
    this.thrown.expectMessage("input must be greater than or equal to zero");
    foo(-1);
} else {
    assertThat(foo(2), equalTo(4));
}
```
**2. Single If Block:**
```java
public void testConditionalThrow() {
    int x = 2;
    if (x < 0) {
        this.thrown.expect(IllegalArgumentException.class);
        this.thrown.expectMessage("input must be greater than or equal to zero");
    }
    int y = foo(x);
    assertThat(y, equalTo(4));
}
```
The updated recipe correctly handles both patterns, including nested conditional blocks. (refer unit test to understand the transformation)

### Additional Improvements

- Enhanced transformation of `thrown.expectMessage(String)` to use Hamcrest matchers, aligning with the original behavior.

**Before:**
```java
assertTrue(exception.getMessage().contains("Index 1 out of bounds for length 1"));
```

**After:**
```java
assertThat(exception.getMessage(), containsString("Invalid location: gs://"));
```

@timtebeek 

### Checklist


amishra-u:
@timtebeek updated the PR summary to clearly justify the changes introduced in this update?

timtebeek:
Hi thanks! Didn't get to this one; there's been a lot to get to with codemix.ai next week, but we hope to get this in soon still.

amishra-u:
@timtebeek can you please take a look

MBoegers:
Thanks @amishra-u for your recurring efforts =)

It is a good practice to chain multiple visitors if you are doing multiple steps instead of one big visitor. In this something like add `assertThrows`, add `assertThat`s and remove `thrown.expect*` calls and fields. It contributes to maintainability and understandability a lot. From your understanding, as you rewrite a good potion of the recipe, Is this applicable here?

I like the idea of aligning a recipe with its siblings, but here we are (and previous were) on an opinionated path. Usually, we try not to introduce a second framework (in this case Harmcrest), as it is opinionated and there are recipes for that. The better approach is to migrate to pure JUnit Jupiter assertions and from there go the desired supportive library. But again, not your PR and I have logged an issue to address this repo-wide.

amishra-u:
> Thanks @amishra-u for your recurring efforts =)
> 
> It is a good practice to chain multiple visitors if you are doing multiple steps instead of one big visitor. In this something like add `assertThrows`, add `assertThat`s and remove `thrown.expect*` calls and fields. It contributes to maintainability and understandability a lot. From your understanding, as you rewrite a good potion of the recipe, Is this applicable here?
> 
> I like the idea of aligning a recipe with its siblings, but here we are (and previous were) on an opinionated path. Usually, we try not to introduce a second framework (in this case Harmcrest), as it is opinionated and there are recipes for that. The better approach is to migrate to pure JUnit Jupiter assertions and from there go the desired supportive library. But again, not your PR and I have logged an issue to address this repo-wide.

Thanks for the suggestion!
In this case, I rewrote the entire recipe to handle all relevant cases. I aimed to break the code into logical methods and eliminate redundancy, so I believe it's in good shape. That said, I'm happy to make adjustments if there's anything specific you'd like to see changed.

PR #722, opened 2025-05-08
Fix TemporaryFolderToTempDir: Don't add methods to unrelated classes

amishra-u:
## What's changed?
The TemporaryFolderToTempDir recipe was previously adding newFolder and newFile methods to all classes in the compilation unit and this may produce uncompilable code. The updated code now adds these methods only to the relevant class.

Before
```
class MyTest {
      @Rule
      TemporaryFolder tempDir1 = new TemporaryFolder();
      void foo() throws IOException {
          File file1 = tempDir1.newFolder();
      }
      class InnerClass {
         private static File newFolder(File root, String... subDirs) throws IOException {}
      }
      
      private static File newFolder(File root, String... subDirs) throws IOException { }
  }
```
After
```
class MyTest {
      @Rule
      TemporaryFolder tempDir1 = new TemporaryFolder();
      void foo() throws IOException {
          File file1 = tempDir1.newFolder();
      }
      class InnerClass { }
      
      private static File newFolder(File root, String... subDirs) throws IOException { }
  }
```

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/TemporaryFolderToTempDir.java
@@ -125,7 +125,7 @@ class TemporaryFolderToTempDirVisitor extends JavaVisitor<ExecutionContext> {
                 case "newFile":
                     return convertToNewFile(mi, ctx);
                 case "newFolder":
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Yes makes sense, thanks!

PR #724, opened 2025-05-09
Added implementation to extract any expression arguments from last statement in assertThrows

JohannisK:
## What's changed?
This PR changes the recipe to extract all `expression` arguments into variables outside the `assertThrows()` call.

## What's your motivation?
`assertThrows(..)` should only contain the expression that is expected to throw. 
```
@Test
void test() {
    assertThrows(Exception.class, () -> {
        doA();
        doB();
        testThing(getC());
    });
}
```
breaks that convention because `getC()` could throw.


src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -113,29 +109,80 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                            type = e.getType().toString();
+                        } else if (e.getType() != null && TypeUtils.asClass(e.getType()) != null) {
+                            type = TypeUtils.asClass(e.getType()).getClassName();
+                            builder.contextSensitive().imports(TypeUtils.asFullyQualified(e.getType()).getFullyQualifiedName());
+                            maybeAddImport(TypeUtils.asFullyQualified(e.getType()).getFullyQualifiedName(), false);
+                        }
+
+                        Statement varDecl = builder
+                                .build()
+                                .apply(new Cursor(getCursor(), lambdaStatement), lambdaStatement.getCoordinates().replace(), type, e);
```
Laurens-W, line 167: You're passing in `String type` here which doesn't resolve nicely, if you pass a reference to the actual type (`e.getType()`) it works, though it is fully qualified, so you'll have to register the `ShortenFullyQualifiedTypeReferences` to shorten it

JohannisK: I tried this, but it still fails?
if you comment out line 172 you'll notice it does work for String, but it does not work for the SomeObject reference...
This might be related to the classPath, or the Shortner not picking up the newly added import?

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -113,29 +112,103 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                    }
+                    lambdaStatement = mi.withArguments(lambdaArguments);
+                }
+                return lambdaStatement;
+            }
+
+            private String getVariableName(Expression e) {
+                if (e instanceof J.MethodInvocation) {
+                    String name = ((J.MethodInvocation) e).getSimpleName();
+                    name = name.replaceAll("^get", "");
```
knutwannheden, line 207: Should we also check for the `is` prefix of a boolean or will we in this case never have a boolean anyway?

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -113,29 +112,103 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                                        new JavaType.Variable(null, 0, getVariableName(e), null, e.getType(), null)))));
+
+                        J.Identifier name = varDecl.getVariables().get(0).getName();
+                        statements.add(varDecl.withPrefix(methodStatement.getPrefix().withComments(emptyList())));
+                        lambdaArguments.add(name);
+                    }
+                    lambdaStatement = mi.withArguments(lambdaArguments);
+                }
+                return lambdaStatement;
+            }
```
timtebeek, line 202: as [discussed on Slack](https://moderneinc.slack.com/archives/C088P8067NH/p1746806201767849?thread_ts=1746715345.295159&cid=C088P8067NH) this was based in part on an example that could be simplified with an anonymous visitor; putting that reminder here to use when reviewing & polishing, as I don't expect you to know what adjustments to make based on that comment alone.

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -113,29 +112,104 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                    J.MethodInvocation mi = (J.MethodInvocation) lambdaStatement;
+                    List<Expression> lambdaArguments = new ArrayList<>(mi.getArguments().size());
+                    for (Expression e : mi.getArguments()) {
+                        if (e instanceof J.Identifier || e instanceof J.Literal || e instanceof J.Empty) {
+                            lambdaArguments.add(e);
+                            continue;
+                        }
+
+                        JavaTemplate.Builder builder = JavaTemplate.builder("#{} " + getVariableName(e) + " = #{any()};\n")
+                                .javaParser(JavaParser.fromJavaVersion().classpath(JavaParser.runtimeClasspath()).logCompilationWarningsAndErrors(true));
```
jevanlingen:
```suggestion
```

(used code)

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
jevanlingen: Good call, I'm going to review this one after the code has been simplified.

src/test/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatementTest.java
```diff
@@ -280,4 +281,445 @@ interface InnerInterface {
...
+                  String getB() { return "B"; }
+                  String getC() { return "C"; }
+                  void testThing(String one, String two) {}
+              }
+              """
+          )
+        );
+    }
+
+    @Disabled("Creates duplicate variables `b`")
```
timtebeek: Not yet clear how to resolve this presently, as we can not update the cursor while generating variables inside ListUtils I think.

src/test/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatementTest.java
```diff
@@ -280,4 +281,445 @@ interface InnerInterface {
...
+                  String getA() { return "A"; }
+                  String getB() { return "B"; }
+                  void testThing(String one, String two) {}
+              }
+              """
+          )
+        );
+    }
+
+    @Disabled("Not implemented yet")
```
timtebeek: We could restructure the logic to first and separately pull out variable from method invocaton arguments, and only then look to retain the last argument. Right now those two are mixed, which is why this case is not handled.

timtebeek (commented):
Already addressed quite a few of the earlier criticisms by now adopting JavaTemplate, and making better use of ListUtils. Also discovered two limitations that I've called out with tests for now. Without fixes there we could potentially do harm, so perhaps best addressed. Fixing the last one might help the second to last one as well.

src/test/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatementTest.java
Jenson3210: I stored a map<basename, suffixcounter> so I can use `VariableNameUtils` for the generation of the name and then just change the suffix counter if applicable. No updates of the cursor etc needed in that case. Bit more custom code in this visitor here.

src/test/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatementTest.java
Jenson3210: I just removed a size check that limited this behaviour to the ones where we have more than 1 line inside. 
As the rest of the code already properly used the `with`ers and `ListUtils`, no new lst element was generated if size was 1 and no variables needed to be extracted.

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -176,15 +173,34 @@ private Statement extractExpressionArguments(Statement lambdaStatement, List<Sta
...
                     String name = ((J.MethodInvocation) e).getSimpleName();
                     name = name.replaceAll("^get", "");
                     name = name.replaceAll("^is", "");
                     name = StringUtils.uncapitalize(name);
-                    return VariableNameUtils.generateVariableName(!name.isEmpty() ? name : "x", new Cursor(getCursor(), e), VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER);
+                    variableName = VariableNameUtils.generateVariableName(!name.isEmpty() ? name : "x", new Cursor(getCursor(), e), VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER);
+                } else {
+                    variableName = VariableNameUtils.generateVariableName("x", new Cursor(getCursor(), e), VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER);
+                }
+                if (variableName.matches(".*\\d+$")) {
```
JohannisK: I think extracting prefix and suffix can be a lot shorter, something like:
```
Pattern pattern = Pattern.compile("^(.+?)(\\d+)$");
Matcher matcher = pattern.matcher(input);

if (matcher.matches()) {
        String varname = matcher.group(1);
        int suffix = Integer.parseInt(matcher.group(2));
} else {
        //no trailing ints
}
```

Jenson3210: changed to use Matcher. Thanks!

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
                         if (idx < lambdaStatements.size() - 1) {
-                            return lambdaStatement.withPrefix(methodStatement.getPrefix().withComments(Collections.emptyList()));
+                            return lambdaStatement.withPrefix(methodStatement.getPrefix().withComments(emptyList()));
                         }
 
+                        List<Statement> variableAssignments = new ArrayList<>();
+                        Space prefix = methodStatement.getPrefix().withComments(emptyList());
+                        final Statement newLambdaStatement = extractExpressionArguments(lambdaStatement, variableAssignments, prefix);
                         J.MethodInvocation newAssertThrows = methodInvocation.withArguments(
                                 ListUtils.map(arguments, (argIdx, argument) -> {
```
jevanlingen, line 124: It took me a while to figure out what this `argsIdx == 1` did. I think it's clearer if you just use:

```java
// Only retain the last statement in the lambda block
ListUtils.map(arguments, it ->
            it instanceof J.Lambda ? lambda.withBody(body.withStatements(singletonList(newLambdaStatement))) : it)
```

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
 
                         J.VariableDeclarations.NamedVariable newAssertThrowsVar = assertThrowsVar.withInitializer(newAssertThrows);
-                        return assertThrowsWithVarDec.withVariables(singletonList(newAssertThrowsVar));
+                        variableAssignments.add(assertThrowsWithVarDec.withVariables(singletonList(newAssertThrowsVar)));
+                        return variableAssignments;
                     });
                 })));
             }
+
+            private Statement extractExpressionArguments(Statement lambdaStatement, List<Statement> precedingVars, Space varPrefix) {
```
jevanlingen, line 148: Is there another way to fill the `precedingVars`? This method is now doing two things:
- getting the last call as statement
- filling the [optional] statements in the `precedingVars` variable

By doing two things at once, this method is harder to understand.

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                            variableTypeFqn = e.getType();
+                        } else if (e.getType() instanceof JavaType.FullyQualified) {
+                            JavaType.FullyQualified aClass = (JavaType.FullyQualified) e.getType();
+                            variableTypeShort = aClass.getClassName();
+                            variableTypeFqn = aClass;
+                            maybeAddImport(aClass.getFullyQualifiedName(), false);
+                        }
+
+                        JavaTemplate.Builder builder = JavaTemplate.builder("#{} #{} = #{any()};");
+                        J.VariableDeclarations varDecl = builder.build()
```
jevanlingen, line 170:
```suggestion
                        J.VariableDeclarations varDecl = JavaTemplate.builder("#{} #{} = #{any()};")
                                .build()
```

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                                .apply(new Cursor(getCursor(), lambdaStatement),
+                                        lambdaStatement.getCoordinates().replace(),
+                                        variableTypeShort, getVariableName(e, generatedVariableSuffixes), e);
```
jevanlingen:
```suggestion
                        Cursor c = new Cursor(getCursor(), lambdaStatement);
                        J.VariableDeclarations varDecl = JavaTemplate.builder("#{} #{} = #{any()};")
                                .build()
                                .apply(c, lambdaStatement.getCoordinates().replace(), variableTypeShort, getVariableName(e, generatedVariableSuffixes), e);
```

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                        precedingVars.add(varDecl
+                                .withPrefix(varPrefix).withType(variableTypeFqn));
```
jevanlingen:
```suggestion
                       
                        precedingVars.add(varDecl.withPrefix(varPrefix).withType(variableTypeFqn));
                        
```

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                String variableName;
+                if (e instanceof J.MethodInvocation) {
+                    String name = ((J.MethodInvocation) e).getSimpleName();
+                    name = name.replaceAll("^get", "");
+                    name = name.replaceAll("^is", "");
+                    name = StringUtils.uncapitalize(name);
+                    variableName = VariableNameUtils.generateVariableName(!name.isEmpty() ? name : "x", new Cursor(getCursor(), e), VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER);
+                } else {
+                    variableName = VariableNameUtils.generateVariableName("x", new Cursor(getCursor(), e), VariableNameUtils.GenerationStrategy.INCREMENT_NUMBER);
+                }
```
jevanlingen: This `getVariableName` function is sadly enough a weird mix of using the `VariableNameUtils.generateVariableName` with a custom increment-number-strategy, because at this point you are one level deeper than the cursor (normally you can use the `updateCursor` to prevent your newly created variables being unknown in a possible next step).

I can't come up with a much nicer solution, but at least you can do it like:


```suggestion
                String baseName = "x";
                if (e instanceof J.MethodInvocation) {
                    String name = ((J.MethodInvocation) e).getSimpleName();
                    name = name.replaceAll("^get", "");
                    name = name.replaceAll("^is", "");
                    name = StringUtils.uncapitalize(name);
                    baseName = !name.isEmpty() ? name : "x";
                }
                String variableName = VariableNameUtils.generateVariableName(baseName, getCursor(), INCREMENT_NUMBER)
```

_Notice you don't need `new Cursor(getCursor(), e)`, because you just want to select the current method invocation block to know all the variables._

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -106,37 +111,102 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration methodDecl
...
+                if (matcher.matches()) {
+                    String prefix = matcher.group(1);
+                    int suffix = Integer.parseInt(matcher.group(2));
+                    generatedVariableSuffixes.putIfAbsent(prefix, suffix);
+                    variableName = prefix;
+                }
+                if (generatedVariableSuffixes.containsKey(variableName)) {
+                    int suffix = generatedVariableSuffixes.get(variableName);
+                    generatedVariableSuffixes.put(variableName, suffix + 1);
+                    variableName += suffix;
```
jevanlingen: This code is not good enough, as it can still produce a variable that already exists. I showed it with: https://github.com/openrewrite/rewrite-testing-frameworks/pull/724/commits/8edf33154c09258d149d3b26fd0290a20de2bd0f

Jenson3210: Pushed something to address this one

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
Jenson3210: removed `new Cursor()`

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
Jenson3210: That would also match 
```java
public static <T extends Throwable> T assertThrows(Class<T> expectedType, Executable executable, Supplier<String> messageSupplier) {
```
as the supplier can also be a lambda

jevanlingen (approved):

timtebeek (approved):
Thanks all! Appreciate the extensive test coverage here.

PR #731, opened 2025-05-19
Fix import ambiguity regression by reordering AddImport/RemoveImport

amishra-u:
## What's changed?
PR [#5439](https://github.com/openrewrite/rewrite/pull/5439) introduced logic to detect and handle import ambiguity when adding imports. However, this caused some tests to fail due to the order in which imports were being added and removed in some recipes. When new imports were added before the old ones were removed, this led to false positives for ambiguity, resulting in the use of fully qualified names.

This PR addresses the issue by reordering the maybeAddImport and maybeRemoveImport call. Now maybeRemoveImport precedes the maybeAddImport, to avoid premature ambiguity detection. 
Additionally, some unit tests contained genuinely ambiguous imports that went undetected earlier. Fixed those test to align with the improved import resolution logic.

### Checklist


amishra-u:
CI will pass when this change lands.
https://github.com/openrewrite/rewrite/pull/5452

PR #735, opened 2025-05-20
[draft] Fix for Assertions.assertDoesNotThrow generator

arodionov:
Find the reproducer


arodionov:
Verified locally on https://github.com/apache/shiro

timtebeek (approved):
Great to see; wherever possible indeed better not to have anything context specific, and since we already use placeholders here and pass in the original LST elements we should be good to remove this.

PR #736, opened 2025-05-21
Add Rewrite Recipe to replace AssertJ hasCauseReference with cause

knutwannheden:
This commit introduces a new OpenRewrite recipe that replaces the AssertJ method `hasCauseReference(Throwable)` with `cause().isEqualTo(Throwable)`.

The `hasCauseReference()` method was deprecated in AssertJ 3.27.1 in favor of using `cause()` for better assertion chaining, as discussed in assertj/assertj#3715.

The recipe includes:
- `ReplaceHasCauseReferenceWithCause.java`: The core recipe logic.
- `ReplaceHasCauseReferenceWithCauseTest.java`: Unit tests to verify the transformation, including a documented example and a case where the expected cause is a new instance.

The recipe correctly handles the transformation and updates imports as necessary.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/ReplaceHasCauseReferenceWithCause.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/ReplaceHasCauseReferenceWithCause.java
@@ -1,0 +1,52 @@
+package org.openrewrite.java.testing.assertj;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaTemplate;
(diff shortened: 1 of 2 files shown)


knutwannheden:
A Refaster recipe might more sense. The AI just refused.

timtebeek:
- For now added here https://github.com/PicnicSupermarket/error-prone-support/pull/1710

Perhaps that's a better target for automated PRs given the large volume of examples over there,

Stephan202:
> Perhaps that's a better target for automated PRs given the large volume of examples over there,

Based on local testing every few months, AIs get increasingly better at this indeed. One additional improvement I have on my TODO list (time permitting...), is to move the test cases inside the Refaster rules themselves, using `@TestInput` and `@TestOutput` (or similar) annotated methods. Then the LLM doesn't need to reason across files. (And it would simplify manual coding and review.)

PR #739, opened 2025-05-24
Recipe for closing unclosed static mocks

amishra-u:
## What's changed?
Recipe to close unclosed static mocks in tests. 

## What's your motivation?
In JUnit4, when a timeout is specified, the test executes in a separate thread. Mockito stores static mocks in thread-local storage, which works fine in this context. However, in JUnit5, tests typically run on the main thread, even with timeout, leading to issues when static mocks are not explicitly closed.

This discrepancy causes tests to fail during migration from JUnit4 to JUnit5. Additionally, unclosed static mocks can leak into subsequent tests, leading to unpredictable failures and making debugging significantly more difficult. 

@timtebeek 

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/mockito.yml
+++ b/src/main/resources/META-INF/rewrite/mockito.yml
@@ -82,6 +82,7 @@ recipeList:
       artifactId: byte-buddy*
       newVersion: 1.12.19
   - org.openrewrite.java.testing.mockito.ReplaceInitMockToOpenMock
+  - org.openrewrite.java.testing.mockito.CloseUnclosedStaticMocks
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.testing.mockito.Mockito1to3Migration


greg-at-moderne (approved):
I am hardly an SME on this topic. But the description you have provided and the test case contributed make sense to me. I have reviewed the code from a general perspective and have no comments. LGTM.

I am gonna give my colleagues a chance to submit some feedback they might have. Otherwise let's merge.

greg-at-moderne:
Thanks for the contribution!

src/test/java/org/openrewrite/java/testing/mockito/CloseUnclosedStaticMocksTest.java
```diff
@@ -0,0 +1,429 @@
...
+            )
+          );
+    }
+
+    @DocumentExample
+    @Test
+    void shouldWrapMockStaticInTryWithResources() {
+        //language=java
+        rewriteRun(
+          spec -> spec.afterTypeValidationOptions(TypeValidation.builder().identifiers(false).build()), // TODO Remove escape hatch
```
timtebeek, line 60: Tried for a bit if we could remove this ignore, but it seemed a bit tricky. May try again when fresh, although I'm not opposed to merging as is.

timtebeek:
I've pushed one last change to include this by default, given the challenge outlined above, such that folks do not need to seek this out explicitly for a recipe run.

src/test/java/org/openrewrite/java/testing/mockito/CloseUnclosedStaticMocksTest.java
amishra-u: We can achieve this by splitting the logic into two passes. The first pass will scan the file to determine what changes are needed, and the second pass will apply those changes. This approach is similar to scanningRecipe, but we don’t need to use scanningRecipe here since all the required information is available within the same file.

However, this might not be worth the effort, given that the functionality is already implemented.

Issue #477, opened 2024-02-07
Spring Boot 3.2 Migration breaks Testcontainers due to exclusion of JUnit 4 as part of JUnit4to5Migration


tedyoung:
## What version of OpenRewrite are you using?

Ran command via

`mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-spring:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2`

Which used Rewrite 5.22.0, relevant output:

```
[INFO] <<< rewrite:5.22.0:run (default-cli) < process-test-classes @ testcontainer-migrate <<<
[INFO] 
[INFO] --- rewrite:5.22.0:run (default-cli) @ testcontainer-migrate ---
[INFO] Using active recipe(s) [org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2]
[INFO] Using active styles(s) []
Downloading from central: https://repo.maven.apache.org/maven2/org/openrewrite/recipe/rewrite-spring/maven-metadata.xml
Downloaded from central: https://repo.maven.apache.org/maven2/org/openrewrite/recipe/rewrite-spring/maven-metadata.xml (2.8 kB at 132 kB/s)
[INFO] Validating active recipes...
[INFO] Project [testcontainer-migrate] Resolving Poms...
[INFO] Project [testcontainer-migrate] Parsing source files
[INFO] Running recipe(s)...
[WARNING] Changes have been made to pom.xml by:
[WARNING]     org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2
[WARNING]         org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_1
[WARNING]             org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_0
[WARNING]                 org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_7
[WARNING]                     org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_6
[WARNING]                         org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_5
[WARNING]                             org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_4
[WARNING]                                 org.openrewrite.java.spring.boot2.SpringBoot2JUnit4to5Migration
[...]

Issue #429, opened 2023-11-22
JUnit4to5Migration excludes required dependency of testcontainers


simonzn:
org.testcontainers:testcontainers has a hard dependency on Junit4, see testcontainers/testcontainers-java#970

The JUnit4to5Migration adds an exclusion to the dependency, which breaks our builds. The issue above suggests crude workarounds like creating dummy classes in your project, but it seems cleaner to not add the exclusion in the first place.

As a workaround I added another recipe to our list which reverts the change; this has the (IMO cosmetic) issue that it creates an empty diff. Pull request is coming up for discussion.

I am aware this is really an issue with testcontainers and not so much with the JUnit4to5Migration recipe, but I think many projects could benefit from a workaround. This should be safe as well - once the issue in testcontainers is resolved it's a NOOP.

## What version of OpenRewrite are you using?
I am using
- OpenRewrite 8.9.0
- rewrite-maven-plugin 5.14.0-SNAPSHOT
- rewrite-testing-frameworks 2.1.0

## How are you running OpenRewrite?
I am executing the JUnit4to5Migration recipe with the Maven command line.

## What is the smallest, simplest way to reproduce the problem?
```xml
<project>
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.example.jackson</groupId>
    <artifactId>test-plugins</artifactId>
    <version>1.0.0</version>
    <dependencies>
        <dependency>
            <groupId>org.testcontainers</groupId>
            <artifactId>testcontainers</artifactId>
            <version>1.18.3</version>
            <scope>test</scope>
        </dependency>
    </dependencies>
</project>
```

## What did you expect to see?
```xml
<project>
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.example.jackson</groupId>
    <artifactId>test-plugins</artifactId>
    <version>1.0.0</version>
    <dependencies>
        <dependency>
            <groupId>org.testcontainers</groupId>
            <artifactId>testcontainers</artifactId>
            <version>1.18.3</version>
            <scope>test</scope>
        </dependency>
    </dependencies>
</project>
```

## What did you see instead?

```xml
<project>
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.example.jackson</groupId>
    <artifactId>test-plugins</artifactId>
    <version>1.0.0</version>
    <dependencies>
        <dependency>
            <groupId>org.testcontainers</groupId>
            <artifactId>testcontainers</artifactId>
[...]

PR #746, opened 2025-06-07
Do not ever exclude junit if Testcontainers is used

DidierLoiseau:
## What's changed?
Instead of adding the JUnit 4 exclusion to all dependencies and then removing it from specific ones that depend on Testcontainers, just don’t add the exclusion when Testcontainers is used.

## What's your motivation?
We have implemented some test libraries that (transitively) depend on Testcontainers. When (indirectly) running the JUnit 5 migration recipe, it always adds the junit exclusion on those, which breaks the build. This forces us to un-exclude JUnit 4 as a followup recipe.

Current workaround is to remove the exclusion explicitly for all dependencies after JUnit 5 migration:
```yml
  - org.openrewrite.maven.RemoveExclusion:
      groupId: '*'
      artifactId: '*'
      exclusionGroupId: junit
      exclusionArtifactId: junit
```

## Any additional context
This makes the recipe more conservative. If a project transitively depends on Testcontainers, it won’t add the exclusion to dependencies that do not depend on Testcontainers (but still depend on JUnit 4) but that’s ok since exclusions have to be added everywhere to work properly.

This provides a more generic solution to #429 and #477.

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
@@ -162,157 +162,165 @@ class JUnit5MigrationTest implements RewriteTest {
         // Just using play-test_2.13 as an example because it appears to still depend on junit.
         // In practice, this would probably just break it, I assume.
         //language=xml
-        String before = """
-          <project>
-              <modelVersion>4.0.0</modelVersion>
-              <parent>
-                  <groupId>org.springframework.boot</groupId>
-                  <artifactId>spring-boot-starter-parent</artifactId>
-                  <version>3.2.1</version>
-                  <relativePath/> <!-- lookup parent from repository -->
-              </parent>
-              <groupId>dev.ted</groupId>
-              <artifactId>needs-exclusion</artifactId>
-              <version>0.0.1</version>
-              <dependencies>
-                  <dependency>
+        rewriteRun(
+          pomXml(
+            """
+              <project>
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
```diff
@@ -157,9 +157,76 @@ void upgradeMavenPluginVersions() {
         );
     }
 
+    @Test
+    void excludeJunit4Dependency() {
+        // Just using play-test_2.13 as an example because it appears to still depend on junit.
```
DidierLoiseau, line 162: I actually wanted to provide a pom with the junit dependency in the test, and then run it as:
```java
        rewriteRun(
          mavenProject("depends-on-junit", pomXml(remoteDepWithJunit, SourceSpec::skip)),
          mavenProject("needs-exclusion", pomXml(before, after)));
```
and somehow it works – the exclusion is added as intended – but it also gets a `MavenDownloadException` added in the output.

I was trying to do the same as other Maven tests that use `SourceSpec::skip`, but they all seem to do that for a parent, not a dependency.

timtebeek (approved):
Welcome back! Great use of a precondition indeed for this longer standing wart in our recipes.

Still will keep fingers crossed for a Junit4-less-version-of-testcontainers
 - https://github.com/testcontainers/testcontainers-java/issues/970
 - https://github.com/testcontainers/testcontainers-java/pull/10351

PR #760, opened 2025-07-07
Adding dependency guards to the Junit 4 -> 5 migration for when the POM or build.gradle have a dependency on `org.testng:testng`. 

steve-aom-elliott:
- Relies on https://github.com/openrewrite/rewrite/pull/5725

## What's changed?
There was already a code guard for usage of `org.testng..*`, but not a guard for if there was a dependency on it in pom.xml or build.gradle. This meant it was able to make changes to some of those files regarding exclusions that left it in a half-way state.

### Checklist


src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -57,6 +57,12 @@ preconditions:
   - org.openrewrite.java.search.DoesNotUseType:
       fullyQualifiedTypeName: org.testng..*
       includeImplicit: true
+  - org.openrewrite.gradle.search.DoesNotIncludeDependency:
+      groupId: org.testng
+      artifactId: testng
+  - org.openrewrite.maven.search.DoesNotIncludeDependency:
+      groupId: org.testng
+      artifactId: testng
```
sambsnyd, line 65: Repeated twice?

sambsnyd (approved):
When there are both gradle and maven variations of a recipe we often create one which bundles the two together in rewrite-java-dependencies

steve-aom-elliott:
I'll see if I can round out a PR for that as well and then switch this over to that.

src/main/resources/META-INF/rewrite/junit5.yml
Laurens-W: Could we introduce an overarching recipe?
`org.openrewrite.java.dependencies.DoesNotIncludeDependency`?

steve-aom-elliott:
Added recipe in `rewrite-java-dependencies` as `org.openrewrite.java.dependencies.search.DoesNotIncludeDependency`
- Relies on https://github.com/openrewrite/rewrite-java-dependencies/pull/144 which in turn relies on https://github.com/openrewrite/rewrite/pull/5758

src/main/java/org/openrewrite/java/testing/junit5/TestNgGuard.java
```diff
@@ -0,0 +1,171 @@
...
+
+public class TestNgGuard extends ScanningRecipe<TestNgGuard.Accumulator> {
+    @Override
+    public String getDisplayName() {
+        return "JUnit Jupiter migration from JUnit 4.x";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Migrates JUnit 4.x tests to JUnit Jupiter.";
```
timtebeek: These look copied from elsewhere.

steve-aom-elliott: Updated in latest push

steve-aom-elliott:
Working on fixes for higher level tests. I imagine they're situations where I need to wrap the test in a `mavenProject`, though maybe this is an indicator that I need to make the guard more flexible for situations where for some reason there's no surrounding project for a Java file.

timtebeek:
Discussed a potential big simplification: it's unlikely the classes are present without a dependency, so focus on only that.
```diff
diff --git i/src/main/java/org/openrewrite/java/testing/junit5/TestNgGuard.java w/src/main/java/org/openrewrite/java/testing/junit5/TestNgGuard.java
index 80198ae0..5461b63d 100644
--- i/src/main/java/org/openrewrite/java/testing/junit5/TestNgGuard.java
+++ w/src/main/java/org/openrewrite/java/testing/junit5/TestNgGuard.java
@@ -15,10 +15,10 @@
  */
 package org.openrewrite.java.testing.junit5;
 
-import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.java.dependencies.search.DoesNotIncludeDependency;
+import org.openrewrite.java.dependencies.search.ModuleHasDependency;
 import org.openrewrite.java.marker.JavaProject;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.marker.SearchResult;
@@ -27,7 +27,7 @@ import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
 
-public class TestNgGuard extends ScanningRecipe<TestNgGuard.Accumulator> {
+public class TestNgGuard extends ScanningRecipe<Set<JavaProject>> {
     @Override
     public String getDisplayName() {
         return "Find `TestNG`-free Maven / Gradle and Java files";
@@ -39,123 +39,23 @@ public class TestNgGuard extends ScanningRecipe<TestNgGuard.Accumulator> {
                 "that are part of a project that does not have TestNG dependencies nor usage of TestNG classes.";
     }
 
-    @Value
-    public static class Accumulator {
-        Set<JavaProject> projectsWithoutTestNgDependency;
-        Set<JavaProject> projectsWithoutTestNgTypeUsage;
-        Set<JavaProject> projectsWithTestNgDependency;
-        Set<JavaProject> projectsWithTestNgTypeUsage;
-        Set<SourceFile> looseUndesirables;
-        Set<SourceFile> looseDesirables;
+    @Override
+    public Set<JavaProject> getInitialValue(ExecutionContext ctx) {
+        return new HashSet<>();
     }
 
     @Override
-    public Accumulator getInitialValue(ExecutionContext ctx) {
-        return new Accumulator(
-            new HashSet<>(),
-            new HashSet<>(),
-            new HashSet<>(),
-            new HashSet<>(),
-            new HashSet<>(),
-            new HashSet<>()
-        );
+    public TreeVisitor<?, ExecutionContext> getScanner(Set<JavaProject> acc) {
+        return new ModuleHasDependency("org.testng", "testng*", null, null).getScanner(acc);
     }
 
     @Override
-    public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
+    public TreeVisitor<?, ExecutionContext> getVisitor(Set<JavaProject> acc) {
         return new TreeVisitor<Tree, ExecutionContext>() {
-            private final TreeVisitor<?, ExecutionContext> ut = new UsesType<>("org.testng..*", true);
-            private final TreeVisitor<?, ExecutionContext> dnid = new DoesNotIncludeDependency("org.testng", "testng*", null, null, null).getVisitor();
-
             @Override
             public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
-                assert tree != null;
-                if (!(tree instanceof SourceFile)) {
-                    return tree;
-                }
-                SourceFile s = (SourceFile) tree;
-                if (dnid.isAcceptable(s, ctx)) {
-                    Tree after = dnid.visit(tree, ctx);
-                    Optional<JavaProject> maybeProject = tree.getMarkers().findFirst(JavaProject.class);
-                    if (after != tree) {
-                        if (maybeProject.isPresent()) {
-                            acc.projectsWithoutTestNgDependency.add(maybeProject.get());
-                        } else {
-                            acc.looseDesirables.add(s);
-                        }
-                    } else {
-                        if (maybeProject.isPresent()) {
-                            acc.projectsWithTestNgDependency.add(maybeProject.get());
-                        } else {
-                            acc.looseUndesirables.add(s);
-                        }
-                    }
-                } else if (ut.isAcceptable(s, ctx)) {
-                    Tree after = ut.visit(tree, ctx);
-                    Optional<JavaProject> maybeProject = tree.getMarkers().findFirst(JavaProject.class);
-                    // Note: Uses `UsesType` for inverted checking because `DoesNotUseType` will mark non Java-code files as well
-                    if (after == tree) {
-                        if (maybeProject.isPresent()) {
-                            acc.projectsWithoutTestNgTypeUsage.add(maybeProject.get());
-                        } else {
-                            acc.looseDesirables.add(s);
-                        }
-                    } else {
-                        if (maybeProject.isPresent()) {
-                            acc.projectsWithTestNgTypeUsage.add(maybeProject.get());
-                        } else {
-                            acc.looseUndesirables.add(s);
-                        }
-                    }
-                }
-                return tree;
-            }
-        };
-    }
-
-    @Override
-    public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
-        return new TreeVisitor<Tree, ExecutionContext>() {
-            private final TreeVisitor<?, ExecutionContext> ut = new UsesType<>("org.testng..*", true);
-            private final TreeVisitor<?, ExecutionContext> dnid = new DoesNotIncludeDependency("org.testng", "testng*", null, null, null).getVisitor();
-
-            @Override
-            public boolean isAcceptable(SourceFile sourceFile, ExecutionContext ctx) {
-                return ut.isAcceptable(sourceFile, ctx) || dnid.isAcceptable(sourceFile, ctx);
-            }
-
-            @Override
-            public Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
-                assert tree != null;
-                boolean isLooseDesirable = acc.looseDesirables.contains((SourceFile) tree);
-                boolean isLooseUndesirable = acc.looseUndesirables.contains((SourceFile) tree);
                 Optional<JavaProject> maybeJp = tree.getMarkers().findFirst(JavaProject.class);
-                if (!maybeJp.isPresent()) {
-                    // if has TestNG or didn't scan
-                    if (isLooseUndesirable || !isLooseDesirable) {
-                        return tree;
-                    }
-                    return SearchResult.found(tree);
-                }
-                JavaProject jp = maybeJp.get();
-                boolean noTestNgDep = acc.projectsWithoutTestNgDependency.contains(jp);
-                boolean noTestNgUsage = acc.projectsWithoutTestNgTypeUsage.contains(jp);
-                boolean testNgDep = acc.projectsWithTestNgDependency.contains(jp);
-                boolean testNgUsage = acc.projectsWithTestNgTypeUsage.contains(jp);
-                // if any TestNG, break
-                if (testNgUsage || testNgDep) {
-                    return tree;
-                }
-                // if not a scanned project
-                if (!noTestNgDep && !noTestNgUsage) {
-                    return tree;
-                }
-                boolean depFreeOrVacuous = noTestNgDep || acc.projectsWithoutTestNgDependency.isEmpty();
-                boolean usageFreeOrVacuous = noTestNgUsage || acc.projectsWithoutTestNgTypeUsage.isEmpty();
-                if (depFreeOrVacuous && usageFreeOrVacuous) {
-                    return SearchResult.found(tree);
-                }
-                return tree;
+                return maybeJp.isPresent() && acc.contains(maybeJp.get()) ? tree : SearchResult.found(tree);
             }
         };
     }
diff --git i/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java w/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
index f41fbf5d..dfa6bb4f 100644
--- i/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
+++ w/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
@@ -700,29 +700,5 @@ class JUnit5MigrationTest implements RewriteTest {
               )
             );
         }
-
-        @Test
-        void noChangesIfTestNgCassIncluded() {
-            rewriteRun(
-              spec -> spec
-                .parser(JavaParser.fromJavaVersion()
-                  .classpathFromResources(new InMemoryExecutionContext(), "junit-4", "testng")),
-              mavenProject("project",
-                //language=java
-                java(
-                  """
-                    import org.junit.Ignore;
-                    import org.testng.annotations.Test;
-
-                    class ExampleClass {
-                        @Ignore
-                        @Test
-                        public void testMethod() {}
-                    }
-                    """
-                )
-              )
-            );
-        }
     }
 }
```

src/main/resources/META-INF/rewrite/junit5.yml
```diff
@@ -58,9 +58,7 @@ tags:
   - testing
   - junit
 preconditions:
-  - org.openrewrite.java.search.DoesNotUseType:
-      fullyQualifiedTypeName: org.testng..*
-      includeImplicit: true
+  - org.openrewrite.java.testing.junit5.TestNgGuard
```
timtebeek: The likeliness of this precondition now matching in the wild is quite a bit larger. I'm wondering if we should ask whoever reported this issue to take on this precondition in their recipes, as opposed to including it here.

In practical terms I'd still hope it to be rare that TestNG is present on the classpath in anyway, so we could still adopt this PR as is, but figured point this out before a merge.

timtebeek (approved):
Approved save for the small caveat pointed out in
 - https://github.com/openrewrite/rewrite-testing-frameworks/pull/760/files#r2241141235

if @sambsnyd agrees then we can merge, to have some shared responsibility.

src/main/resources/META-INF/rewrite/junit5.yml
steve-aom-elliott: It looks like entry point for them is `org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_1` which is fairly high up the chain (about 10 levels above the eventual `org.openrewrite.java.testing.junit5.JUnit4to5Migration` composite recipe), hmm. The composite recipe that we have the condition on here itself calls _47_ other recipes, so it's quite the expansion to alter that uniquely. I know that the composite recipe also leads back up to `org.openrewrite.java.spring.boot3.SpringBoot33BestPractices`, `org.openrewrite.java.testing.assertj.Assertj` and `org.openrewrite.java.testing.junit5.JUnit5BestPractices`.

src/main/java/org/openrewrite/java/testing/junit5/TestNgGuard.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public Set<JavaProject> getInitialValue(ExecutionContext ctx) {
+        return new HashSet<>();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Set<JavaProject> acc) {
+        return new ModuleHasDependency("org.testng", "testng*", null, null).getScanner(acc);
```
sambsnyd: Seems like we could add a flag to `ModuleHasDependency` to invert its own marking behavior, then we wouldn't need specialized "module doesn't have dependency" recipes like this one

steve-aom-elliott: - @sambsnyd https://github.com/openrewrite/rewrite-java-dependencies/pull/146

steve-aom-elliott: Switched to using the new option on `ModuleHasDependency` instead of a separate class in the latest push.

steve-aom-elliott:
Failing at the moment because it seems `openrewrite/rewrite-java-dependencies` isn't seeing the new version of Gradle `ChangeDependency` when it's trying to make a build: https://github.com/openrewrite/rewrite-java-dependencies/actions/runs/16660315340/job/47155660481#step:6:350

steve-aom-elliott:
- Added PR https://github.com/openrewrite/rewrite-java-dependencies/pull/147 for allowing to mark non-project-wrapped loose Java files so that it doesn't break tests upstream that aren't caring about having TestNG in the first place.

steve-aom-elliott:
@sambsnyd Looks like this is back to green if you'd be able to give a final look over with it using `ModuleHasDependency` instead of a separate recipe for the guard.

timtebeek (approved):

PR #763, opened 2025-07-09
Inconvertible Mockito `when` situation involving lambda and static method.

steve-aom-elliott:
Showing situation where we wouldn't be able to safely convert `when(..).thenReturn(..)` to `staticMocked.when(..).thenReturn(..)`

## What's changed?
- There's a particular setup that would require a wider scoped change than the current recipe is trying to target to get correct, and would require context about how a try-with-resource variable is used to get correct. Current thought is that rather than have it result immediately in something that can't be compiled, to not make a change instead.

### Checklist


src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -120,7 +120,8 @@ private List<Statement> maybeWrapStatementsInTryWithResourcesMockedStatic(J.Bloc
 
             private @Nullable String getClassName(J.MethodInvocation whenArg) {
                 J.Identifier clazz = null;
-                if (whenArg.getSelect() instanceof J.Identifier) {
+                // Having a fieldType implies that something is an instance of a class rather than a class itself
```
sambsnyd:
```suggestion
                // Having a fieldType implies that something is a field rather than a class itself
```

PR #765, opened 2025-07-11
Removing `NoInitializationForInjectMock` recipe due to side effects when using recipe

steve-aom-elliott:
## What's changed?
- Removed `NoInitializationForInjectMock` recipe, corresponding tests and updated another test that expected its behaviour to be present.

## What's your motivation?
Original motivation here was there were cases where
```java
@InjectMocks
private AInterface = new AImplementation()
```
was getting changed to
```java
@InjectMocks
private AInterface;
```
which then wouldn't compile because you can't inject into an interface.

I did some more small scale tests, and there seem to be a number of cases where using this recipe at all might result in varying levels of behaviour difference.

Assuming you have
```java
class A {
  private Object object;
  public A() {}
  public A(Object object) {
    this.object = object;
  }
}
```

If you start with
```java
//...
@Mock
Object objectMock;

@InjectMocks
A subject = new A();

@BeforeEach
void setup() {
  when(objectMock.toString()).thenReturn("blah");
}
//...
```
By default, what's happening is that **only** `A()` is called, but `Object object` inside `subject` is configured as the mocked version still.

### Approach 1 - Drop initializer
If you change
```java
@InjectMocks
A subject = new A();
```
to
```java
@InjectMocks
A subject;
```
`A(Object object)` is called instead, and `Object object` inside `subject` is configured as the mocked version.

### Approach 2 - Drop annotation
If you change
```java
@InjectMocks
A subject = new A();
```
to
```java
A subject = new A();
```
`A()` is called still but `Object object` inside `subject` is `null`.

Unfortunately this means that either way you try to make a change, it has the potential to be a behavioural difference, so the safest path here likely to not make use of this recipe.

### Checklist


timtebeek (approved):
Thanks for the detailed analysis!

Issue #762, opened 2025-07-08
Avoid `assertTrue()` or `assertFalse()` with constant boolean arguments


greg-at-moderne:
## What problem are you trying to solve?
Test code readability.

In a test case:
- `assertTrue(true)` is a no-op
- `assertTrue(false)` can be replaced with `fail()`

Similar for `assertFalse`, etc. The same logic could probably be done in several variants for various test frameworks.

## Describe the situation before applying the recipe

```java
    @Test
    void testExceptionThrown() {
        try {
            someOperationWith("Invalid argument");
            fail("Expected exception was not thrown");
        } catch (IllegalArgumentException e) {
            assertTrue(true);             
        }
    }
```

## Describe the situation after applying the recipe

```java
    @Test
    void testExceptionThrown() {
        try {
            someOperationWith("Invalid argument");
            fail("Expected exception was not thrown");
        } catch (IllegalArgumentException e) {
        }
    }
```

## Any additional context
I've stumbled upon this pattern when providing OSS examples for 
- #761.

aet2505:
@timtebeek  I've just seen that the linked issue https://github.com/openrewrite/rewrite-testing-frameworks/issues/761 is using testNg not junit, should we clone these same recipes for testNg assertions

timtebeek:
Thanks for the sharp eye there; I'd say no need for now: we support TestNG mostly when migrating away to JUnit and AssertJ. We could in theory add TestNG best practices recipes, but I wonder if there's enough of an audience to warrant the effort. Personally I'd nudge folks to switch to AssertJ.

aet2505:
Cool! I might have a go at the linked one over the next day or so. Will put a PR up if I get anywhere interesting

PR #767, opened 2025-07-13
Avoid assertTrue() or assertFalse() with constant boolean arguments 

aet2505:
## What's changed?
- Enhance existing recipe to convert `assertTrue(false, <message>)` to `fail(<message>)` to handle the no message overload
- Add new recipe to remove cases such as `assertTrue(true)`
- Add new recipe to Junit5BestPractices recipe list

## What's your motivation?
Resolve Issue https://github.com/openrewrite/rewrite-testing-frameworks/issues/762

In the new recipe any message passed to the assert will be lost.
While this message will never be fired, it potentially is used as a pseudo comment? Not sure what the best pattern here is.

- I considered keeping these as refaster recipes but that would lead to a change in recipe name for the enhancement. Maintaining the recipe name felt beneficial.
- Again due to maintaining the recipe name I chose to add the second case as a new recipe. While this could be merged it did not feel like behaviour that would be expected under the existing name

### Checklist


src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToNoOpRecipe.java
```diff
@@ -0,0 +1,98 @@
...
+            JavaParser.Builder<?, ?> javaParser = null;
+
+            private JavaParser.Builder<?, ?> javaParser(ExecutionContext ctx) {
+                if (javaParser == null) {
+                    javaParser = JavaParser.fromJavaVersion()
+                            .classpathFromResources(ctx, "junit-jupiter-api-5");
+                }
+                return javaParser;
+            }
+
```
timtebeek: Why this inclusion?
```suggestion
```

src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToNoOpRecipe.java
```diff
@@ -0,0 +1,98 @@
...
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+
+
+public class AssertLiteralBooleanToNoOpRecipe extends Recipe {
```
timtebeek: Instead of a new imperative recipe, could this not have been a Refaster recipe as well?

src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToFailRecipe.java
```diff
@@ -0,0 +1,114 @@
...
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaType;
+
+
+public class AssertLiteralBooleanToFailRecipe extends Recipe {
```
timtebeek: What made it that this could no longer be a Refaster recipe? As that's not clear immediately, and Refaster is preferred where possible.

src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToNoOpRecipe.java
aet2505: This probably could have been a Refaster recipe. 
Assuming an empty method is valid in Refaster, equivalent to returning null from the visitor.
Will rework this

src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToFailRecipe.java
aet2505: I'd have needed to make two Refaster recipes to handle this, making the top level recipe a 'set' (not sure if right term)
This would have changed the recipe id which I thought might have unintended impact.
I chose to make this an imperative recipe for this reason.
If changing IDs is not an impact I can refactor to be a refaster recipe, please let me know your preference

Thank you for the review comments so far!

timtebeek: I'd view it as merely adding a new Refaster recipe alongside the original one we already had, and then including both in a declarative recipe.

src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToNoOpRecipe.java
aet2505: I've just had a look at changing this and it doesn't appear that an empty body is supported by Refaster

I get the following stack trace when compiling with an empty body

Caused by: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
at jdk.compiler/com.sun.tools.javac.util.List.get(List.java:476)
at org.openrewrite.java.template.processor.RefasterTemplateProcessor$TemplateDescriptor.validate(RefasterTemplateProcessor.java:938)

src/main/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanToFailRecipe.java
aet2505: Having now seen that converting the other recipe to refaster is not possible.
And knowing that maintaining the recipe ID is not important.
I'd suggest merging these two recipes into a single recipe which handles all assertTrue/False(<literal>) cases.

Any thoughts @timtebeek?

timtebeek: Seeing this use case I think we can make an adjustment to the annotation processor to allow empty after blocks:
- https://github.com/openrewrite/rewrite-templating/pull/153

src/test/java/org/openrewrite/java/testing/cleanup/AssertLiteralBooleanRemovedTest.java
```diff
@@ -0,0 +1,113 @@
...
+                      Assertions.assertFalse(false, "message");
+                      Assertions.assertTrue(true, "message");
+                      Assertions.assertFalse(false);
+                      Assertions.assertTrue(true);
+                  }
+              }
+              """,
+            """
+              import org.junit.jupiter.api.Assertions;
+
```
timtebeek: This can be removed after a new release containing
- https://github.com/openrewrite/rewrite-templating/pull/155
```suggestion
```

timtebeek (approved):
Thanks for the help here @aet2505 ! I've made some changes to the Refaster annotation processor such that we can still use Refaster here, as it's that much more convenient that I didn't want to have to maintain imperative recipes. We'll merge this once there's a new release of rewrite-templating tomorrow to apply the last code suggestion here.

PR #770, opened 2025-07-18
Remove all PowerMock imports after migration

JohannisK:
## What's changed?
Added a Cleanup recipe to remove all powermock imports after migration

## What's your motivation?
In some cases (for instance when types are imported but not used) the migration recipe does not hit `maybeRemoveImport(..)`
Since we do remove the dependencies this results in compile errors.

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/mockito/ReplacePowerMockitoIntegrationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/mockito/ReplacePowerMockitoIntegrationTest.java
@@ -886,101 +886,23 @@ class ReplacePowerMockitoIntegrationTest implements RewriteTest {
         );
     }
 
-    @Test
-    void powerMockitoUnused() {
+    @ParameterizedTest
+    @ValueSource(strings = {
+      "import org.powermock.api.mockito.PowerMockito;",
+      "import static org.powermock.api.mockito.PowerMockito.whenNew;",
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/testing/mockito/CleanupPowerMockImports.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static class CleanupPowerMockImportsVisitor extends JavaIsoVisitor<ExecutionContext> {
+        @Override
+        public @Nullable J preVisit(J tree, ExecutionContext ctx) {
+            stopAfterPreVisit();
+            if (tree instanceof JavaSourceFile) {
+                JavaSourceFile sf = (JavaSourceFile) tree;
+                for (J.Import _import : sf.getImports()) {
+                    if (_import.getPackageName().startsWith("org.powermock")) {
+                        maybeRemoveImport(_import.getPackageName() + "." + _import.getClassName());
```
timtebeek: Would we want to force remove these imports instead?

JohannisK: We could, but I think it would fail one way or another.
If the type is still used we don't remove the import, compilation will fail because the library is missing.
If the type is still used but we force to remove the import, compilation will fail because we use a type which is not imported.

I can't seem to find a variant of `maybeRemoveImport` where I can force removal. Do you have a preference (and example)?

PR #788, opened 2025-08-10
MockitoJUnitRunnerToExtension: Handle existing MockitoExtension

amishra-u:
## What's changed?
This PR updates the MockitoJUnitRunnerToExtension recipe to better handle the migration from MockitoJUnitRunner to MockitoExtension. Specifically:

- It now checks for the presence of an existing @ExtendWith(MockitoExtension.class) annotation to avoid adding duplicates.
- It verifies the configured strictness level, and only updates it if the existing level is more restrictive than what JUnit 4's runner.

## What's your motivation?
In some cases a developer may have bothe MockitoRule and MockitoJunitRunner both in same test class. In that case the MockitoJUnitToMockitoExtension and MockitoJUnitRunnerToExtension both tries to add `MockitoExtension` or `MockitoSettings` which is not repeatable and test fails with compilation error. Updated MockitoJUnitRunnerToExtension to not add these extension if one already exists, Since MockitoJUnitRunnerToExtension recipes runs after MockitoJUnitToMockitoExtension Junit4toJunit5 recipe.

### Checklist


timtebeek (approved):
Thanks for the changes here! Wasn't expected indeed that classes would already have both, but good to handle that gracefully.

PR #795, opened 2025-08-20
Adapt `@EnabledOnJre`, `@DisabledOnJre`, `@EnabledForJreRange` and `@DisabledForJreRange` for JUnit 6's Java 17 minimum

Jenson3210:
# Add recipe to simplify JRE version conditions for JUnit 6 migration

## Summary

This PR introduces a new recipe `MinimumJreConditions` that simplifies JUnit Jupiter test conditions based on a minimum JRE version requirement. As part of the JUnit 6 migration path, this recipe helps clean up obsolete JRE version conditions when migrating to a minimum Java version.

### New Recipe: `MinimumJreConditions`

The recipe processes JUnit Jupiter's JRE conditional annotations (`@EnabledOnJre`, `@DisabledOnJre`, `@EnabledForJreRange`, `@DisabledForJreRange`) and:

- **Removes obsolete test methods** that are conditionally enabled only for JRE versions older than the minimum
- **Simplifies annotations** by removing references to older JRE versions from multi-version conditions
- **Cleans up annotations** that become redundant when only the minimum JRE version remains
- **Preserves necessary conditions** for JRE versions at or above the minimum requirement

### Key Features

1. **Supports multiple annotation formats:**
   - JRE enum values: `JRE.JAVA_8`, `JRE.JAVA_17`, etc. (also the statically imported ones)
   - Value attribute: `value = JRE.JAVA_17`
   - Version integers: `versions = 17` or `versions = { 17, 21 }`
   - Range attributes: `minVersion = 17, maxVersion = 21`

2. **Handles range conditions:**
   - Removes tests enabled only for ranges ending before the minimum JRE
   - Preserves tests for ranges that include or exceed the minimum JRE
   - Removes annotations for disabled ranges that end before the minimum

3. **Smart annotation cleanup:**
   - Removes annotations when they become redundant (e.g., `@EnabledOnJre(JRE.JAVA_17)` when 17 is the minimum)
   - Updates multi-version conditions by filtering out obsolete versions
   - Preserves test structure when conditions are still relevant

### Integration with JUnit 6 Migration

The recipe has been added to the main JUnit 6 migration recipe with Java 17 as the default minimum version, aligning with JUnit 6's requirements.

### Files Changed

- `src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java` - Recipe implementation
- `src/main/resources/META-INF/rewrite/junit6.yml` - Added recipe to JUnit 6 migration
- `src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java` - Comprehensive test coverage

### Before
```java
class MyTest {
    @Test
    @EnabledOnJre(JRE.JAVA_8)
    void testOnlyOnJava8() {
        // This test only runs on Java 8
    }
    
    @Test
    @EnabledOnJre(versions = { 11, 17, 21 })
    void testOnMultipleVersions() {
        // This test runs on Java 11, 17, and 21
    }
    
    @Test
    @EnabledForJreRange(min = JRE.JAVA_8, max = JRE.JAVA_11)
    void testOnOldVersions() {
        // This test runs on Java 8-11
    }
}
```

### After (with minimum JRE 17)
```java
class MyTest {
    // testOnlyOnJava8 removed - obsolete for Java 17+
    
    @Test
    @EnabledOnJre(versions = {17, 21})
    void testOnMultipleVersions() {
        // Updated to only include relevant versions
    }
    
    // testOnOldVersions removed - range is entirely below minimum
}
```

## Testing

Comprehensive test coverage includes:
- Parameterized tests for various annotation formats
- Edge cases for range conditions
- Multi-version annotation handling
- Preservation of non-test methods
- Handling of special JRE values (e.g., `JRE.OTHER`)

## Related Issues

This recipe supports the JUnit 6 migration effort by ensuring test suites are properly cleaned up when moving to newer minimum JRE requirements.

### DIFF
--- a/src/main/resources/META-INF/rewrite/junit6.yml
+++ b/src/main/resources/META-INF/rewrite/junit6.yml
@@ -56,3 +56,8 @@ recipeList:
       attributeName: lineSeparator
   - org.openrewrite.java.testing.junit6.MinimumJreConditions:
       javaVersion: 17
+  - org.openrewrite.java.ChangeMethodName:
+      methodPattern: org.junit.jupiter.api.extension.ExtensionContext.Store getOrComputeIfAbsent((..))
+      newMethodName: computeIfAbsent
+      matchOverrides: true
+      ignoreDefinition: true


src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,264 @@
...
+                }
+
+                if (enabledOnJre.isPresent()) {
+                    if (enabledOnJre.get().stream().allMatch(v -> compareVersions(v, javaVersion) < 0)) {
+                        // Remove the test method if it is enabled on a JRE version lower than the specified version
+                        return null;
+                    }
+                    if (enabledOnJre.get().stream().filter(v -> compareVersions(v, javaVersion) >= 0).allMatch(v -> compareVersions(v, javaVersion) == 0)) {
+                        // Remove the annotation if it is enabled on the same JRE version
+                        RemoveAnnotation removeAnnotation = new RemoveAnnotation("@" + ENABLED_ON_JRE);
```
Jenson3210: Is this really something we want to do? 
If our test suite can run on java 17 and 21, this impacts behavior?

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,264 @@
...
+                        m = (J.MethodDeclaration) updatedVersions.getVisitor().visit(m, ctx, getCursor().getParent());
+                    }
+                }
+                if (enabledOnJreRange.isPresent()) {
+                    if (compareVersions(enabledOnJreRange.map(Range::getMax).get(), javaVersion) < 0) {
+                        // Remove the test method if it is enabled on a JRE range that ends before the specified version
+                        return null;
+                    }
+                    if (compareVersions(enabledOnJreRange.map(Range::getMin).get(), javaVersion) < 0) {
+                        //TODO update the range to start at the specified version and convert to a @EnabledOnJre annotation if the range is now a single version
```
Jenson3210: Note for reviewers: I want to do these range updates in a separate iteration as this will be quite a complex piece (Detecting the layout / property, using the template...

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,265 @@
...
+
+@Value
+@EqualsAndHashCode(callSuper = false)
+public class MinimumJreConditions extends Recipe {
+
+    private static final String ENABLED_ON_JRE = "org.junit.jupiter.api.condition.EnabledOnJre";
+    private static final String DISABLED_ON_JRE = "org.junit.jupiter.api.condition.DisabledOnJre";
+    private static final String ENABLED_FOR_JRE_RANGE = "org.junit.jupiter.api.condition.EnabledForJreRange";
+    private static final String DISABLED_FOR_JRE_RANGE = "org.junit.jupiter.api.condition.DisabledForJreRange";
+    private static final Annotated.Matcher TEST_ANNOTATION_MATCHER = new Annotated.Matcher("@org.junit.jupiter.api.Test");
```
timtebeek: There's quite a few more test annotations we'd want to support, like `@ParameterizedTest`, `@RepeatedTest` and others similarly covered for other recipes. Can we add support here?

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                          System.out.println("Any version");
+                      }
+                  }
+                  """.formatted(jre),
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.EnabledOnJre;
+                  import org.junit.jupiter.api.condition.JRE;
+
+                  import static org.junit.jupiter.api.condition.JRE.*;
```
timtebeek: I'd have hoped to see these now unused imports removed, as opposed to having those linger after a removal.

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+
+                  class MyTest {
+                      @Test
+                      void testOnJava17() {
+                          System.out.println("Java 17");
+                      }
+                  }
+                  """
```
timtebeek: Seeing this now we should probably retain these, as the tests might just as well run on Java 21 or 25; Java 17 is the new minimum version supported, but not guaranteed to be the version used.
```suggestion
                  """.formatted(jre)
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      }
+                  }
+                  """,
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.EnabledOnJre;
+                  import org.junit.jupiter.api.condition.JRE;
+
+                  class MyTest {
+                      @Test
```
timtebeek, line 287: Similarly, this likely needs to retain only 17, to avoid running on 21+ 
```suggestion
                      @Test
                      @EnabledOnJre(versions = 17)
```

Jenson3210: I can agree there, but... test suites are often written once and then not touched. With the emerging of new java versions, people will have written a test to only run on Java 17 and not on 11. By moving to 21 later on, losing/dropping/removing the test is perhaps losing coverage on some functionality tests?

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                  }
+                  """,
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.EnabledOnJre;
+                  import org.junit.jupiter.api.condition.JRE;
+
+                  class MyTest {
+
+                      @Test
```
timtebeek, line 329: By extension, then here too:
```suggestion
                      @Test
                      @EnabledOnJre(JRE.JAVA_21)
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                          System.out.println("Not Java 8");
+                      }
+                  }
+                  """.formatted(jre),
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.DisabledOnJre;
+                  import org.junit.jupiter.api.condition.JRE;
+
+                  import static org.junit.jupiter.api.condition.JRE.*;
```
timtebeek:
```suggestion
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      void testOnJava8To11() {
+                          System.out.println("Java 8-11");
+                      }
+                  }
+                  """.formatted(range),
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.EnabledForJreRange;
+                  import org.junit.jupiter.api.condition.JRE;
+
```
timtebeek, line 525:
```suggestion
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      @Test
+                      @EnabledForJreRange(%s)
+                      void testOnJava11AndLater() {
+                          System.out.println("Java 11+");
+                      }
+                  }
+                  """.formatted(range)
+              )
+            );
+        }
```
timtebeek: I think ideally here we'd want to set a new minimum of 17 going forward right, for open ranges?

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      void testUpToJava11() {
+                          System.out.println("Up to Java 11");
+                      }
+                  }
+                  """.formatted(range),
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.EnabledForJreRange;
+                  import org.junit.jupiter.api.condition.JRE;
+
```
timtebeek:
```suggestion
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      void testOnJava8To16() {
+                          System.out.println("Java 8-16");
+                      }
+                  }
+                  """.formatted(range),
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.EnabledForJreRange;
+                  import org.junit.jupiter.api.condition.JRE;
+
```
timtebeek:
```suggestion
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      @Test
+                      @EnabledForJreRange(%s)
+                      void testOnJava11To21() {
+                          System.out.println("Java 11-21");
+                      }
+                  }
+                  """.formatted(range)
+              )
+            );
+        }
```
timtebeek: Ideally as well, set a new minimum of 17.

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      @DisabledForJreRange(%s)
+                      void testNotOnJava8To11() {
+                          System.out.println("Not Java 8-11");
+                      }
+                  }
+                  """.formatted(range),
+                """
+                  import org.junit.jupiter.api.Test;
+                  import org.junit.jupiter.api.condition.DisabledForJreRange;
+                  import org.junit.jupiter.api.condition.JRE;
```
timtebeek:
```suggestion
```

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,843 @@
...
+                      @Test
+                      @DisabledForJreRange(%s)
+                      void testNotOnJava11To21() {
+                          System.out.println("Not Java 11-21");
+                      }
+                  }
+                  """.formatted(range)
+              )
+            );
+        }
```
timtebeek: Set a new minimum to 17 if possible

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
Stephan202: Those other test annotations can be identified generically by [looking for the `@TestTemplate` meta-annotation](https://github.com/PicnicSupermarket/error-prone-support/blob/3e1db2bdd4c2fe40669e660f7118e09aecd4ad57/error-prone-utils/src/main/java/tech/picnic/errorprone/utils/MoreJUnitMatchers.java#L30-L36).

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
Jenson3210: Ranges and adapting them I had mentioned to do in another iteration of this recipe, but will get on this now that I have addressed the other review comments

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,299 @@
...
+                }
+
+                Space finalPrefix = prefix;
+                return m.withLeadingAnnotations(ListUtils.map(m.getLeadingAnnotations(), ann -> {
+                    if (ENABLED_JRE_MATCHER.get(ann, getCursor().getParent()).isPresent() || DISABLED_JRE_MATCHER.get(ann, getCursor().getParent()).isPresent() || ENABLED_JRE_RANGE_MATCHER.get(ann, getCursor().getParent()).isPresent() || DISABLED_JRE_RANGE_MATCHER.get(ann, getCursor().getParent()).isPresent()) {
+                        ann = ann
+                                .withArguments(ListUtils.map(ann.getArguments(), arg -> {
+                                    if (arg instanceof J.Assignment) {
+                                        J.Assignment ass = (J.Assignment) arg;
+                                        if (ass.getAssignment() instanceof J.NewArray) {
```
Jenson3210, line 246: Could not use the mentioned `SimplifySingleElementAnnotation` as that only acts upon the value assignment and in this annotation the array could be the value (which is the JRE) or the versions.

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
Jenson3210: I can't see the meta-annotations in the type hierarchy on our lst's (nor do I know exactly where to look), but added it like other recipes are doing it.

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
timtebeek: Sure; let me know when this is ready for another round of review and possible merge , or whether anything is still planned or in progress.

Jenson3210: @timtebeek 
Ranges have been implemented causing a bunch of repeated code lines to be extracted into separate methods with a clear name on what they do. 
No more pending items on this one

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -515,7 +515,6 @@ void testOnJava11AndLater() {
                 """
                   import org.junit.jupiter.api.Test;
                   import org.junit.jupiter.api.condition.EnabledForJreRange;
-                  import org.junit.jupiter.api.condition.JRE;
```
Jenson3210: We still need this import in the after scenario -> That's why I only call this maybeRemoveVisitor in the occasion that I removed tests or annotations. As we keep the notation that is used, the import remains valid for the cases where we move from one enum to another.

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,1072 @@
...
+                      void testNotOnJava17() {
+                          System.out.println("Not Java 17");
+                      }
+                  }
+                  """.formatted(jre)
+              )
+            );
+        }
+
+        @CsvSource({
```
timtebeek: I'm somewhat surprised at seeing changes even when we're not adding/removing values. Would it make sense to delegate to that single element annotation visitor only from visiting a changed compilation unit as well?

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
Jenson3210: I suspect the `AddOrUpdateAnnotationAttribute` does not provide sufficient type information on the newValue for the maybeRemoveImport statement to know that the type is still being used. 

By seeing this, I now realize that we might hence miss to do the correc timport for statically imported enum values. :sigh:

timtebeek: hmm; looks like there's a bug in the way we remove imports then, as it should not remove that if there's still a lingering reference. We can tackle that separately, but ideally it's possible from visitCompilationUnit.

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
```diff
@@ -0,0 +1,1073 @@
...
+                  }
+                  """
+              )
+            );
+        }
+
+        @ParameterizedTest
+        @ValueSource(strings = {
+          "min = JRE.JAVA_17, max = JRE.JAVA_21",
+          "minVersion = 17, maxVersion = 21"
```
timtebeek: Would it make sense here to remove the minimum version? Before JUnit 6 that had a signal, now that's implied from running on JUnit 6.

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
Jenson3210: Wouldn't we then have the behavior still on compilation units that are changes in a completely different location? I can check though if the `m != incominMethod` anymore -> there was a change -> unwrap

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
Jenson3210: so, if min is less than or equal to `javaVersion` just emit the min property?

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
Jenson3210: implemented

src/test/java/org/openrewrite/java/testing/junit6/MinimumJreConditionsTest.java
Jenson3210: This is now also implemented. Tests that will now always be disabled will be removed. Tests that will now always be enabled will have their annotation removed

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+                return method;
+            }
+
+            private J.MethodDeclaration removeAnnotation(J.MethodDeclaration m, String annotationType, ExecutionContext ctx) {
+                RemoveAnnotation removeAnnotation = new RemoveAnnotation("@" + annotationType);
+                maybeRemoveImport(JRE_IMPORT);
+                return removeAnnotation.getVisitor().visitMethodDeclaration(m, ctx);
+            }
+
+            private J.MethodDeclaration updateAnnotationVersions(J.MethodDeclaration m, Optional<List<String>> versions, String annotationType, ExecutionContext ctx) {
```
jevanlingen:
```suggestion
            private J.MethodDeclaration updateAnnotationVersions(J.MethodDeclaration m, List<String> versions, String annotationType, ExecutionContext ctx) {
```

Don't use Optional as input argument of a function. In this case it's unnecessary, as you are sure the optional is always true (thus handle such case on the calling side).

On a more general level, it is seen as bad practice to use the Optional as method parameter, as it was never intended as a Maybe type. See for more info https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555 (also note here that Brian Goetz is not 'just' another programmer 😉).

--

In addition to this: I think, now we have JSpecify, we can use the Optional type to an even lesser extent. From a functional side, `@Nullable Object` or `Optional<Object>` concern themselves with the same idea, namely marking a type as `T | null`. Another interesting read about this topic; https://github.com/projectlombok/lombok/wiki/Language-Design:-Null-vs.-Optional.

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+                    annotated = ENABLED_JRE_RANGE_MATCHER.get(annotationCursor);
+                    if (annotated.isPresent()) {
+                        enabledOnJreRange = ENABLED_JRE_RANGE_MATCHER.get(annotationCursor).map(Range::new);
+                        prefix = ann.getPrefix();
+                    }
+                    annotated = DISABLED_JRE_RANGE_MATCHER.get(annotationCursor);
+                    if (annotated.isPresent()) {
+                        disabledOnJreRange = DISABLED_JRE_RANGE_MATCHER.get(annotationCursor).map(Range::new);
+                        prefix = ann.getPrefix();
+                    }
```
jevanlingen:
```suggestion
                    Function<Expression, List<String>> toInitializerName = e -> {
                        if (e instanceof J.NewArray) {
                            List<Expression> initializer = ((J.NewArray) e).getInitializer();
                            return initializer == null ? emptyList() : initializer.stream().map(Objects::toString).collect(toList());
                        }
                        return singletonList(e.toString());
                    };
                    enabledOnJre = ENABLED_JRE_MATCHER.get(annotationCursor)
                            .map(a -> getDefaultAttribute(a).orElse(getAttribute(a, "versions").orElse(null)))
                            .map(toInitializerName);
                    disabledOnJre = DISABLED_JRE_MATCHER.get(annotationCursor)
                            .map(a -> getDefaultAttribute(a).orElse(getAttribute(a, "versions").orElse(null)))
                            .map(toInitializerName);
                    enabledOnJreRange = ENABLED_JRE_RANGE_MATCHER.get(annotationCursor).map(Range::new);
                    disabledOnJreRange = DISABLED_JRE_RANGE_MATCHER.get(annotationCursor).map(Range::new);
                    if (enabledOnJre.isPresent() || disabledOnJre.isPresent() ||  enabledOnJreRange.isPresent() || disabledOnJreRange.isPresent()) {
                        prefix = ann.getPrefix();
                    }
```

Doing a `isPresent` check on an Optional you are gonna map is unnecessary, as a `map` action on an empty Optional will just return the singleton EMPTY optional instance.

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+                    maybeRemoveImport(DISABLED_FOR_JRE_RANGE);
+                }
+                return c;
+            }
+
+            @Override
+            public J.@Nullable MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+                J.MethodDeclaration m = super.visitMethodDeclaration(method, ctx);
+                boolean isUnitTest = false;
+                Optional<List<String>> enabledOnJre = Optional.empty();
```
jevanlingen: I think I would drop the Optionals here, as a null check (instead of `.isPresent()`) for this properties be fine as well.

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+                RemoveAnnotation removeAnnotation = new RemoveAnnotation("@" + annotationType);
+                maybeRemoveImport(JRE_IMPORT);
+                return removeAnnotation.getVisitor().visitMethodDeclaration(m, ctx);
+            }
+
+            private J.MethodDeclaration updateAnnotationVersions(J.MethodDeclaration m, Optional<List<String>> versions, String annotationType, ExecutionContext ctx) {
+                if (versions.isPresent() && versions.get().stream().anyMatch(v -> compareVersions(v, javaVersion) < 0)) {
+                    // Update the annotation to only include JRE versions that are greater than or equal to the specified version
+                    AddOrUpdateAnnotationAttribute updatedVersions = new AddOrUpdateAnnotationAttribute(annotationType, "versions", versions.get().stream().filter(v -> compareVersions(v, javaVersion) >= 0).collect(joining(", ")), null, false, false);
+                    m = (J.MethodDeclaration) updatedVersions.getVisitor().visit(m, ctx, getCursor().getParent());
```
jevanlingen:
```suggestion
                    m = (J.MethodDeclaration) updatedVersions.getVisitor().visitNonNull(m, ctx, getCursor().getParent());
```

Check also all the other `.visit`  methods!

--

If the intention here is that null is ok, then change:
```diff
- private J.MethodDeclaration updateAnnotationVersions(J.MethodDeclaration m, Optional<List<String>> versions, String annotationType, ExecutionContext ctx) {
+ private J.@Nullable MethodDeclaration updateAnnotationVersions(J.MethodDeclaration m, Optional<List<String>> versions, String annotationType, ExecutionContext ctx) {
```

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+                }
+                return m;
+            }
+
+            // Return a method declaration with a new range only if the start is below requested java version.
+            // If the start version would be equal to the end version, then no updated method is returned as this is handled in a later phase
+            private J.MethodDeclaration updateRangeStart(J.MethodDeclaration m, String annotationtype, Range range, ExecutionContext ctx) {
+                if (compareVersions(range.getMin(), javaVersion) <= 0) {
+                    if (compareVersions(range.getMax(), javaVersion) != 0) {
```
jevanlingen:
```suggestion
                if (compareVersions(range.getMin(), javaVersion) <= 0 && compareVersions(range.getMax(), javaVersion) != 0) {
```

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+        return version;
+    }
+
+    private static String formatJreValue(String version, @Nullable RangeNotation notation) {
+        if (notation == null) {
+            return version;
+        }
+        switch (notation) {
+            case VERSION:
+                return version;
```
jevanlingen:
```suggestion
```

Image

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
Jenson3210: I can take the method extraction from this code, but the rest of the code will actually is not correct afaics. 
By assigning the value always to the optional ones, only the last annotationis taken into account. If that one does not match the condition annotation, all optionals will be marked as empty. We can only assign the relevant optional when we detected that the current annotation is the one that we have checked.

Jenson3210: As we are looping over annotations, not doing an ifPresentCheck will always assign all 4 to the latest value. If that is eg a `CSVSource`, all four will be empty optionals where there could perfectly be an annotation earlier on in the loop.

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
```diff
@@ -0,0 +1,412 @@
...
+                if (maxAttribute.get().startsWith("JRE.")) {
+                    this.maxNotation = RangeNotation.JRE;
+                } else {
+                    this.maxNotation = RangeNotation.STATIC_JRE;
+                }
+            } else if (maxVersionAttribute.isPresent()) {
+                this.maxNotation = RangeNotation.VERSION;
+            } else {
+                this.maxNotation = null;
+            }
```
jevanlingen:
```suggestion
            this.minNotation = minAttribute
                    .map(it -> it.startsWith("JRE.") ? RangeNotation.JRE : RangeNotation.STATIC_JRE)
                    .orElse(minVersionAttribute.map(__ -> RangeNotation.VERSION).orElse(null));
            this.maxNotation = maxAttribute
                    .map(it -> it.startsWith("JRE.") ? RangeNotation.JRE : RangeNotation.STATIC_JRE)
                    .orElse(maxVersionAttribute.map(__ -> RangeNotation.VERSION).orElse(null));
```

src/main/java/org/openrewrite/java/testing/junit6/MinimumJreConditions.java
Jenson3210: For the prefix, the same is true. Even if we find a way that only assigns the correct optional while in the loop, i am interested in the correct annotations prefix. Not the last one if any of the previous ones was matched.

jevanlingen: You are right of course! What you optionally could do:

```java
AtomicReference<List<String>> enabledOnJre = new AtomicReference<>();
AtomicReference<Space> prefix = new AtomicReference<>(Space.EMPTY);

for (J.Annotation ann : method.getLeadingAnnotations()) {
    // ..
    ENABLED_JRE_MATCHER.get(annotationCursor)
        .map(a -> getDefaultAttribute(a).orElse(getAttribute(a, "versions").orElse(null)))
        .map(toInitializerName)
        .ifPresent(it -> {
            enabledOnJre.set(it);
            prefix.set(ann.getPrefix());
        });
```

If that's really better I don't know for sure. So I'll leave that up to you.

---

Edit: I doubt it's better, because now you dropped all the optionals, it looks already a lot cleaner.

Jenson3210: I consider the Atomics the last resort. 😄 Even "laster" vs Optionals. 
Because then we end up with the get calls again later in the code, I did keep the extracted method (named differently and not as an inline function) though!

timtebeek (approved):
Great to see! Looking forward to seeing how this will pan out in practice; did you already run against a larger suite of projects?

timtebeek:
> This last commit won't work I guess as versions can also be an J.Array of versions.

Ah then best reverted with an appropriate test added to guard against regressions.

Jenson3210:
No,not yet!

Do you mean this one only? 
As we will need junit 6 for the composite recipe.

Jenson3210:
I was wrong... 
It was the case where @SomeAnnotation({JRE.JAVA12, JRE.JAVA_11}) would be tried to be parsed. 
The Annotated Trait has a built in support for arrays of Literals, but not for the defaulted version where the array is not of literals but of FieldAccesses.

I knew there was something to it.

timtebeek:
> No,not yet! Do you mean this one only?

We can likely eye-ball the diffs this one only would produce on an org like Apache, even locally with the CLI, and from there possibly see possible improvements, or have enough confidence to merge this PR already.

Jenson3210:
What would we want to do in this situation @timtebeek? Can we safely assume the comment on same line will be 
Screenshot 2025-08-22 at 22 24 24

Rest is looking good though (ran against Apache ~ 1200 repos, no errors and 20'ish validated results)! 
Screenshot 2025-08-22 at 22 27 17
Screenshot 2025-08-22 at 22 27 56

I must say... It keeps feeling weird to remove test cases that verify "reflection-java-8/11/15-only-usecases" in certain java versions for a library just because the test framework runs tests in a higher junit version.

timtebeek:
> What would we want to do in this situation @timtebeek? Can we safely assume the comment on same line will be 

I think it's fine not to handle that for now;if we did we should handle it in the visitor that removes the annotation instead.

> Rest is looking good though (ran against Apache ~ 1200 repos, no errors and 20'ish validated results)!

Great! Thanks for looking through those.

> I must say... It keeps feeling weird to remove test cases that verify "reflection-java-8/11/15-only-usecases" in certain java versions for a library just because the test framework runs tests in a higher junit version.

I agree it looks odd, but perhaps it helps to look at the deprecated enum values?
https://docs.junit.org/6.0.0-RC1/api/org.junit.jupiter.api/org/junit/jupiter/api/condition/JRE.html
image

So then the way I see it they are there only to avoid compilation issues, but not actually used with the recommendation to use `JAVA_17` going forward. Effectively there would be no way to run these tests, so then there would be a false sense of security to have them still; better then I think to make folks aware by removing those tests, as opposed to keeping them around.

PR #798, opened 2025-08-27
No try staticMocking when field is already statically mocked. 

Jenson3210:
When a MockedStatic is already present, we should not add a new try block as that breaks the mockito test cases


src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -270,4 +294,157 @@ private static boolean isMethodDeclarationWithAnnotation(@Nullable Statement sta
         }
         return false;
     }
+
+
+    @Value
+    private static class VariableUtil {
```
Jenson3210: Copied this from VariableNameUtils scraping the super/inherited fields for now. 
I strongly believe this should not be part of this recipe, but not sure how else to fix this issue.

Jenson3210: @timtebeek do you have any suggestions here?

Jenson3210:
Screenshot 2025-08-28 at 09 00 33

We must know the variables to know if we either do a migration or not somehow. Otherwise we have no way of knowing the migration I am afraid.

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
Jenson3210: moved to a utils package for now

timtebeek: Jacob had a very similar case yesterday for rewrite-kafka, in `MigrateWindowStorePutMethod.java`; I had half a mind to do what you've done here. The main difference I see with the approach taken there is that it only looks at anything in scope _before_ the point of use, whereas `VariableNameUtils` avoids conflicts anywhere in the same scope. That ordering might matter here if there's more than one MockedStatic, although I haven't yet looked at your implementation in detail.

Jenson3210: hmm... good point. This implementation will indeed also look at later declarations also
We have `LivenessAnalysis` as MPL but cannot use that here.

Jenson3210: The current implementation also does not check if the variables are live / assigned at the point in time. 
But I do not see any other way to fix that.

Jenson3210: would you feel it is better to just look at the variables in the class declaration? 
Typically they will be prepared with @Before and closed with @After.

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -245,6 +257,21 @@ private static List<J.Try.Resource> getMatchingFilteredResources(@Nullable List<
...
     }
 
+    private static Optional<J.Identifier> tryGetMatchedWrappingVariableName(Cursor cursor, String className) {
+        return VariableUtil.findVariablesInScope(cursor).stream()
+                .filter(identifier -> {
+                    if (MOCKED_STATIC.matches(identifier) && identifier.getType() instanceof JavaType.Parameterized) {
+                        JavaType.Parameterized parameterizedType = (JavaType.Parameterized) identifier.getType();
+                        if (parameterizedType.getTypeParameters().size() == 1) {
+                            JavaType mockedClass = parameterizedType.getTypeParameters().get(0);
+                            return TypeUtils.isAssignableTo(className, mockedClass);
```
jevanlingen:
```suggestion
                            return TypeUtils.isAssignableTo(className, parameterizedType.getTypeParameters().get(0));
```

Optional change, just shorter (but arguably less readable).

Jenson3210: Agreed, I think this one was was still from checking out what was available / working during debugging session.

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
Jenson3210: Altered it to be a non-public thing. Still variables after the current line would be visited, but i think that is better compared to the previous coverage / breaking.

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -270,4 +273,43 @@ private static boolean isMethodDeclarationWithAnnotation(@Nullable Statement sta
         }
         return false;
     }
+
+    private static J.@Nullable Identifier findMockedStaticVariable(Cursor scope, String className) {
+        JavaSourceFile compilationUnit = scope.firstEnclosing(JavaSourceFile.class);
+        if (compilationUnit == null) {
+            throw new IllegalStateException("A JavaSourceFile is required in the cursor path.");
```
timtebeek: Can we safely return `null` here instead? I don't think it'll ever happen, but it's still not a pattern we want to have folks think is ok.

timtebeek (approved):
Some further simplifications applied just now; one left over comment, but otherwise ok to merge once addressed, thanks!

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
Jenson3210: yes we can. This is remainder code from the VariableNameUtils. `null` is perfectly fine also! ✅

timtebeek (approved):
Nice! Even clearer with that AtomicReference indeed. As a small perfomance improvement you could `return block` in `visitBlock` when the reference is already set. That stops traversal early.

PR #800, opened 2025-08-28
Removed redundant checks and added support for JUnit 5 annotations

steve-aom-elliott:
## What's changed?
- First commit is for adding support for recognizing JUnit 5's annotations (as the names have changed vs JUnit 4).
- Second commit is adding tests that fail that showcase that the logic for writing the `close()` method invocation need to be refined to ensure that we're not adding the call into a method where it would cause failures.

## What's your motivation?
We were only handling JUnit 4 and very recently TestNG, but not JUnit 5 annotations.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
@@ -37,13 +37,15 @@ import static org.openrewrite.java.tree.Flag.Static;
 
 public class MockitoWhenOnStaticToMockStatic extends Recipe {
 
-    private static final AnnotationMatcher JUNIT_ANNOTATION = new AnnotationMatcher("org.junit.*");
+    private static final AnnotationMatcher JUNIT_4_ANNOTATION = new AnnotationMatcher("org.junit.*");
+    private static final AnnotationMatcher JUNIT_5_ANNOTATION = new AnnotationMatcher("org.junit.jupiter.api.*");
     private static final AnnotationMatcher TESTNG_ANNOTATION = new AnnotationMatcher("org.testng.annotations.*");
 
     private static final AnnotationMatcher BEFORE = new AnnotationMatcher("org..Before*");
     private static final AnnotationMatcher BEFORE_CLASS = new AnnotationMatcher("org..BeforeClass");
+    private static final AnnotationMatcher BEFORE_ALL = new AnnotationMatcher("org..BeforeAll");
+    private static final AnnotationMatcher BEFORE_PARAM_CLASS_INV = new AnnotationMatcher("org..BeforeParameterizedClassInvocation");
     private static final AnnotationMatcher AFTER = new AnnotationMatcher("org..After*");
-    private static final AnnotationMatcher AFTER_CLASS = new AnnotationMatcher("org..AfterClass");
 
     private static final MethodMatcher MOCKITO_WHEN = new MethodMatcher("org.mockito.Mockito when(..)");
     private static final TypeMatcher MOCKED_STATIC = new TypeMatcher("org.mockito.MockedStatic");
@@ -58,7 +60,9 @@ public class MockitoWhenOnStaticToMockStatic extends Recipe {
     @Override
     public String getDescription() {
         return "Replace `Mockito.when` on static (non mock) with try-with-resource with MockedStatic as Mockito4 no longer allows this. " +
-                "For JUnit: When `@Before` or `@BeforeClass` is used, a `close` call is added to either the `@After` or `@AfterClass` method. " +
+                "For JUnit 4: When `@Before` or `@BeforeClass` is used, a `close` call is added to either the `@After` or `@AfterClass` method. " +
+                "For JUnit 5: When `@BeforeEach`, `@BeforeAll` or `@BeforeParameterizedClassInvocation` is used, " +
+                "a `close` call is added to a corresponding `@AfterEach`, `@AfterAll` or `@AfterParameterizedClassInvocation` method. " +
                 "For TestNG: When `@BeforeMethod` or `@BeforeClass` is used, a `close` call is added to either the `@AfterMethod` or `@AfterClass` method. " +
                 "This change moves away from implicit bytecode manipulation for static method stubbing, making mocking behavior more explicit and scoped to avoid unintended side effects.";
     }
@@ -68,7 +72,7 @@ public class MockitoWhenOnStaticToMockStatic extends Recipe {
         return Preconditions.check(new UsesMethod<>(MOCKITO_WHEN), new JavaIsoVisitor<ExecutionContext>() {
             @Override
             public J.Block visitBlock(J.Block block, ExecutionContext ctx) {
-                List<Statement> newStatements = isMethodDeclarationWithAnnotation(getCursor().firstEnclosing(J.MethodDeclaration.class), BEFORE, BEFORE_CLASS) ?
+                List<Statement> newStatements = isMethodDeclarationWithAnnotation(getCursor().firstEnclosing(J.MethodDeclaration.class), BEFORE) ?
                         maybeStatementsToMockedStatic(block, block.getStatements(), ctx) :
                         maybeWrapStatementsInTryWithResourcesMockedStatic(block, block.getStatements(), ctx);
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -195,14 +199,17 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, Ex
                                 .build()
                                 .apply(updateCursor(classDecl), classDecl.getBody().getCoordinates().firstStatement());
 
-                        if (classDecl.getBody().getStatements().stream().noneMatch(it -> isMethodDeclarationWithAnnotation(it, AFTER, AFTER_CLASS))) {
-                            Optional<Statement> beforeMethodJunit = after.getBody().getStatements().stream()
-                                    .filter(it -> isMethodDeclarationWithAnnotation(it, JUNIT_ANNOTATION))
+                        if (classDecl.getBody().getStatements().stream().noneMatch(it -> isMethodDeclarationWithAnnotation(it, AFTER))) {
```
jevanlingen: Code looks fine to me! But maybe you can use some functions to remove the duplication (only if it really cleans things up).

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -37,13 +37,15 @@
...
 public class MockitoWhenOnStaticToMockStatic extends Recipe {
 
-    private static final AnnotationMatcher JUNIT_ANNOTATION = new AnnotationMatcher("org.junit.*");
+    private static final AnnotationMatcher JUNIT_4_ANNOTATION = new AnnotationMatcher("org.junit.*");
+    private static final AnnotationMatcher JUNIT_5_ANNOTATION = new AnnotationMatcher("org.junit.jupiter.api.*");
     private static final AnnotationMatcher TESTNG_ANNOTATION = new AnnotationMatcher("org.testng.annotations.*");
 
     private static final AnnotationMatcher BEFORE = new AnnotationMatcher("org..Before*");
     private static final AnnotationMatcher BEFORE_CLASS = new AnnotationMatcher("org..BeforeClass");
+    private static final AnnotationMatcher BEFORE_ALL = new AnnotationMatcher("org..BeforeAll");
```
timtebeek, line 46: Would `BeforeAll` not already be picked up by the wildcard in `"org..Before*"` ?

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -296,6 +302,86 @@ private static boolean isMethodDeclarationWithAnnotation(@Nullable Statement sta
...
+                .collect(toList());
+        if (existingSimilarMethods.isEmpty()) {
+            return baseName;
+        }
+        String newName = baseName;
+        int count = 0;
+        while (existingSimilarMethods.contains(newName)) {
+            newName = baseName + (count += 1);
+        }
+        return newName;
```
jevanlingen:
```suggestion
        return existingStatements.stream()
                .filter(it -> it instanceof J.MethodDeclaration)
                .map(it -> ((J.MethodDeclaration) it).getSimpleName())
                .filter(s -> s.matches("^" + baseName + "(\\d+)?$"))
                .max(Comparator.comparingInt(s -> s.equals(baseName) ? 0 : Integer.parseInt(s.substring(baseName.length()))))
                .map(last -> {
                    int suffix = last.equals(baseName) ? 0 : Integer.parseInt(last.substring(baseName.length()));
                    return baseName + (suffix + 1);
                })
                .orElse(baseName);
```

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -189,57 +195,57 @@ private List<Statement> mockedStatic(J.Block block, J.MethodInvocation statement
...
                                 .build()
                                 .apply(updateCursor(classDecl), classDecl.getBody().getCoordinates().firstStatement());
 
-                        if (classDecl.getBody().getStatements().stream().noneMatch(it -> isMethodDeclarationWithAnnotation(it, AFTER, AFTER_CLASS))) {
-                            Optional<Statement> beforeMethodJunit = after.getBody().getStatements().stream()
-                                    .filter(it -> isMethodDeclarationWithAnnotation(it, JUNIT_ANNOTATION))
+                        List<Statement> afterStatements = after.getBody().getStatements();
+                        AnnotationMatcher specificBeforeMatcher = new AnnotationMatcher(matchedAnnotation);
+                        if (classDecl.getBody().getStatements().stream().noneMatch(it -> isMethodDeclarationWithAnnotation(it, new AnnotationMatcher(correspondingAfterFqn)))) {
+                            String safeAfterMethodName = getSafeAfterMethodName("tearDown", afterStatements);
```
jevanlingen: Hardcoded "tearDown", is that right?

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -296,6 +302,86 @@ private static boolean isMethodDeclarationWithAnnotation(@Nullable Statement sta
...
+        }
+        String newName = baseName;
+        int count = 0;
+        while (existingSimilarMethods.contains(newName)) {
+            newName = baseName + (count += 1);
+        }
+        return newName;
+    }
+
+    private static @Nullable String getSimpleName(@Nullable String fqn) {
```
jevanlingen:
```suggestion
    private static @Nullable String getClassName(@Nullable String fqn) {
```

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -296,6 +302,86 @@ private static boolean isMethodDeclarationWithAnnotation(@Nullable Statement sta
...
+    }
+
+    private static Optional<J.Annotation> tryGetMatchedAnnotationOnMethodDeclaration(@Nullable Statement statement, AnnotationMatcher... matchers) {
+        if (statement instanceof J.MethodDeclaration) {
+            return ((J.MethodDeclaration) statement).getLeadingAnnotations().stream()
+                    .filter(it -> Arrays.stream(matchers).anyMatch(m -> m.matches(it)))
+                    .findFirst();
+        }
+        return Optional.empty();
+    }
```
jevanlingen: I would merge this method together with `getAnnotationFgn`:


```suggestion
    private static @Nullable String tryGetMatchedAnnotationOnMethodDeclaration(J.@Nullable MethodDeclaration statement, AnnotationMatcher... matchers) {
        if (statement != null) {
            return statement.getLeadingAnnotations().stream()
                    .filter(it -> Arrays.stream(matchers).anyMatch(m -> m.matches(it)))
                    .findFirst()
                    .map(J.Annotation::getType)
                    .map(Object::toString)
                    .orElse(null);
        }
        return null;
    }
```

This is not only shorter,  but by doing so, the calling side looks much better as well:

```diff
- String matchedAnnotation = requireNonNull(getAnnotationFqn(tryGetMatchedAnnotationOnMethodDeclaration(containingMethod, BEFORE).get()));
+ String matchedAnnotation = requireNonNull(tryGetMatchedAnnotationOnMethodDeclaration(containingMethod, BEFORE));
```

---

N.B.: This code works, because the `map` of an Optional has one weird/handy thing:
> If the mapping function returns a null result, then this method returns an empty Optional.

This is not true monadic behaviour, but well... Who cares 😆.

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
```diff
@@ -296,6 +302,86 @@ private static boolean isMethodDeclarationWithAnnotation(@Nullable Statement sta
...
+                    return "org.testng.annotations.AfterMethod";
+                case "org.testng.annotations.BeforeClass":
+                    return "org.testng.annotations.AfterClass";
+                case "org.testng.annotations.BeforeSuite":
+                    return "org.testng.annotations.AfterSuite";
+                case "org.testng.annotations.BeforeGroups":
+                    return "org.testng.annotations.AfterGroups";
+                case "org.testng.annotations.BeforeTest":
+                    return "org.testng.annotations.AfterTest";
+            }
```
jevanlingen:
```suggestion
            return annotationFqn.replaceFirst(".Before", ".After");
```

⚠️ Don't know if this is a good idea, so just think about the consequences before applying this suggestion! _Because you worked on this way longer than me reviewing it now, you know beter what's right and what isn't._

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
steve-aom-elliott: Well, it was previously hardcoded to `tearDown` without any checks for if there was already a method called that, but I suppose I could at least move it to a field instead.

src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
steve-aom-elliott: Renamed, but also renamed the other `getClassName` to `getClassNameFromInvocation`

timtebeek (approved):

PR #803, opened 2025-09-04
enhance TestsShouldNotBePublicTest

sullis:
## What's changed?

improve test cases in TestsShouldNotBePublicTest

## What's your motivation?

more test coverage

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/cleanup/TestsShouldNotBePublicTest.java
+++ b/src/test/java/org/openrewrite/java/testing/cleanup/TestsShouldNotBePublicTest.java
@@ -188,15 +188,25 @@ class TestsShouldNotBePublicTest implements RewriteTest {
     }
 
     @Test
-    void ignoreOverriddenMethod() {
+    void ignoreOverriddenMethod_parentClassIsAbstract() {
         //language=java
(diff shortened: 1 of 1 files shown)


timtebeek (commented):
Hi @sullis ; thanks for the offer to help. In this case I'm not sure it's a clear improvement. We try to keep our unit tests as minimal as possible for the scenario they are testing, and in this case it doesn't make a difference whether the parent class is abstract or not, as it's the methods that we visit either way. The original test method already covers both sides of this conditional:

https://github.com/openrewrite/rewrite-testing-frameworks/blob/130079ac3acf8a1c2a4a0f2588a1ce8d90b7e36e/src/main/java/org/openrewrite/java/testing/cleanup/TestsShouldNotBePublic.java#L161-L163

I'm afraid the added elements here distract from what we're testing, and might complicate stepping through with a debugger since there's more "skips" and mental parsing needed to figure out exactly what's asserted here. Because of that I'm leaning towards closing this PR for now.

Did you see or expect any issues that prompted you to add these additional elements?

timtebeek:
Closing this one for now; we can reopen if it turns out there was an underlying issue with the above referenced code.

PR #810, opened 2025-09-16
Add debug information and error handling for intermittent ParseError in GradleUseJunitJupiter

Jenson3210:
# Add debug information and error handling for intermittent ParseError in GradleUseJunitJupiter

## Summary

This PR addresses the intermittent `ClassCastException: Cannot cast org.openrewrite.tree.ParseError to org.openrewrite.groovy.tree.G$CompilationUnit` that occurs in Spring Boot 3.5 projects. The changes add comprehensive error handling and debugging capabilities to help identify the root cause of parsing failures.

## Problem Context

The issue manifests as an intermittent `ClassCastException` when the code attempts to cast a `ParseError` to `G.CompilationUnit` at line 252. This occurs rarely but consistently across different projects.

## Solution Approach
Rather than silently filtering out ParseErrors, this implementation:

1. **Provides fallback parsing**: Uses GroovyParser when GradleParser fails, ensuring files are still processed
2. **Preserves diagnostic information**: Attaches original parse errors as markers for debugging when this issue happens
3. **Enables comprehensive error analysis**: When both parsers fail, provides detailed error information from both attempts

## Testing

The changes maintain existing functionality while providing better error handling. Files that previously caused crashes will now either:
- Be successfully processed using the GroovyParser fallback, or
- Fail with detailed diagnostic information instead of generic ClassCastException

This approach ensures recipes continue working **while gathering the debugging information needed to identify and fix the root cause of the intermittent parsing failures.** 

# ! This is not a fix proposal, but a information retrieval proposal, after which the clean solution still needs to be added !


src/main/java/org/openrewrite/java/testing/junit5/GradleUseJunitJupiter.java
```diff
@@ -240,17 +240,50 @@ public J.Lambda visitLambda(J.Lambda l, ExecutionContext ctx) {
...
     private static Optional<J.MethodInvocation> createTaskUseJUnitPlatform(ExecutionContext ctx, boolean forEachInvocation) {
-        return GradleParser.builder()
+        String snippet = "plugins {\n" +
+                "    id 'java'\n" +
+                "}\n" +
+                "tasks.withType(Test)" + (forEachInvocation ? ".configureEach" : "") + " {\n" +
+                "    useJUnitPlatform()\n" +
+                "}";
+        Optional<SourceFile> parsed = GradleParser.builder()
+                .buildscriptClasspath(new ArrayList<>())
```
Jenson3210:
```suggestion
```

timtebeek (approved):

Jenson3210:
@knutwannheden You mentioned you had other idea to test this?

timtebeek:
We're in parallel making some other changes to the Gradle parser upstream:
- https://github.com/openrewrite/rewrite/pull/6227
- https://github.com/openrewrite/rewrite/pull/6228

PR #818, opened 2025-09-24
Create byteman.yml

marcosgopen:
I noticed byteman dep was missing to the updated dependency in my project, so adding here the recipe


timtebeek:
We need an upstream fix to remove the old class imports here:
- https://github.com/openrewrite/rewrite/pull/6077

PR #821, opened 2025-09-25
remove dependency: google truth

sullis:
## What's changed?

Google Truth migration recipe should remove Google Truth dependency


timtebeek:
Capturing the same context that was shared on Slack here as well: There's still (at least) two cases we don't yet convert:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/3662e547ad2165268c7542e59e2dc0b5f78b43e2/src/main/java/org/openrewrite/java/testing/truth/TruthAssertToAssertThat.java#L49-L53
https://github.com/openrewrite/rewrite-testing-frameworks/blob/3662e547ad2165268c7542e59e2dc0b5f78b43e2/src/main/java/org/openrewrite/java/testing/truth/TruthCustomSubjectsToAssertJ.java#L49-L54

That makes me somewhat hesitant to remove the dependency, as it's hard to say whether we might end up with those stray remnants causing compilation failures.

timtebeek:
Since we're doing a ChangeType just before we'd end up with compliation issues anyway. Might then as well remove the dependency. Thanks!

PR #826, opened 2025-10-01
Fixing bug in `MockitoWhenOnStaticToMockStatic` related to originally not using the FQN of the class when detecting it from a `when(X.y())` style usage.

steve-aom-elliott:
## What's changed?
Making the `isAssignableTo` check make use of the fully qualified name rather than just the simple name and updating tests to involve a mocked static class that has a package to illustrate.

## What's your motivation?
When the static mocked type belonged to a package, the matching was failing during detecting of an existing `MockedStatic<X>` field declaration, meaning it would try to produce a new try-with-resources instead of reusing an already defined field.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/MockitoWhenOnStaticToMockStatic.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.testing.mockito;
 
+import lombok.Value;
 import org.jspecify.annotations.Nullable;
 import org.openrewrite.*;
 import org.openrewrite.internal.ListUtils;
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for figuring out what needed to be done here; hope you agree with my edits to phase out that `InvokedTypeWrapper` which wasn't needed. In general try to avoid String-typing where you can; I know that was in the original here, but the underlying type works better. :)

steve-aom-elliott:
@timtebeek It makes sense. I did try to avoid the wrapper at one point, but I was attempting to just pass around a `JavaType` rather than `JavaType.Class` and was having challenges getting both pieces of information that I needed out of it at the time. Wasn't certain if I could always cast it to that or if there were scenarios that I wasn't thinking of where doing so might break it.

Thanks for the fixes!

PR #828, opened 2025-10-06
Separate best practices for JUnit 5, JUnit 6 and Jupiter

timtebeek:
This way both JUnit 5 and JUnit 6 each can have best practices, without running the JUnit 5 upgrade recipes twice.

### DIFF
--- a/src/main/resources/META-INF/rewrite/junit6.yml
+++ b/src/main/resources/META-INF/rewrite/junit6.yml
@@ -16,34 +16,6 @@
 
 ---
 type: specs.openrewrite.org/v1beta/recipe
-name: org.openrewrite.java.testing.junit6.JUnitJupiterBestPractices
-displayName: JUnit Jupiter best practices
-description: Applies best practices to tests.
-tags:
-  - testing
-  - junit
-recipeList:
-  # Included by both `JUnit6BestPractices` and `JUnit5BestPractices`, so avoid adding upgrade recipes here to run twice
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/rewrite/junit6.yml
```diff
@@ -14,6 +14,46 @@
 # limitations under the License.
 #
 
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.testing.junit6.JUnitJupiterBestPractices
```
Jenson3210: Will there not be junit6 best practices that are incompatible with junit5? Should this one not live in `org.openrewrite.java.testing.junit5.JUnitJupiterBestPractices` as that's where this list originates from?

timtebeek: So my thought was to keep this one agnostic for 5/6, and then include any JUnit 6 specific best practices under `org.openrewrite.java.testing.junit.JUnit6BestPractices` just below. Would you agree with that set up?

Jenson3210: If we want to keep it agnostic, perhaps we should have `org.openrewrite.java.testing.junit.JUnitJupiterBestPractices`?

Jenson3210: Though it will take some time for version 7 to come out 😄

Jenson3210: Approved, leaving this one up for your opinion, but I am fine either way

PR #848, opened 2025-11-21
Add `@ExtendWith(MockitoExtension.class)` Test using Mockito Annotations

MBoegers:
## What's changed?

A recipe that adds `@ExtendWith(MockitoExtension.class)` if `@Mock` or `@Captor` is used.

## What's your motivation?

Spring Boot 4.0 requires adding these https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#mockito-captor-and-mock-annotations 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
@@ -23,6 +23,7 @@ import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.search.FindAnnotations;
+import org.openrewrite.java.search.FindTypes;
 import org.openrewrite.java.search.IsLikelyTest;
 import org.openrewrite.java.tree.J;
 
@@ -45,7 +46,7 @@ public class AddMockitoExtensionIfAnnotationsUsed extends Recipe {
 
         TreeVisitor<?, ExecutionContext> hasExtendedWithAnnotation = new FindAnnotations("org.junit.jupiter.api.extension.ExtendWith(org.mockito.junit.jupiter.MockitoExtension.class)", false).getVisitor();
         @SuppressWarnings("unchecked")
-        TreeVisitor<?, ExecutionContext>[] hasManyMockitoAnnotation = new TreeVisitor[] {
+        TreeVisitor<?, ExecutionContext>[] hasAnyMockitoAnnotation = new TreeVisitor[]{
                 // see https://www.baeldung.com/mockito-annotations for examples
                 new FindAnnotations("org.mockito.Captor", false).getVisitor(),
                 new FindAnnotations("org.mockito.Mock", false).getVisitor(),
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
```diff
@@ -0,0 +1,68 @@
...
+
+                maybeAddImport("org.mockito.junit.jupiter.MockitoExtension");
+                maybeAddImport("org.junit.jupiter.api.extension.ExtendWith");
+
+                J.Annotation extendsWith = ((J.ClassDeclaration) JavaTemplate.builder("@ExtendWith(MockitoExtension.class)")
+                        .imports("org.mockito.junit.jupiter.MockitoExtension")
+                        .imports("org.junit.jupiter.api.extension.ExtendWith")
+                        .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "junit-jupiter-api", "mockito-junit-jupiter"))
+                        .build()
+                        .apply(getCursor(), classDecl.getCoordinates().replaceAnnotations()))
```
Jenson3210: Any reason of replacing all annotations? 
People using other annotations (eg. `@DBeaver`) will loose these annotations. Is that intentional?

MBoegers: I dont replace the annotation, I just use the coordinate to generate an LST element from which I take the to add later at return. IFAC I avoided the add annotation coordinate ;)

Jenson3210: missed that last line :)

MBoegers: I've added a test to prove it is working as intended see https://github.com/openrewrite/rewrite-testing-frameworks/pull/848/commits/7f01a765ed4b357c4ace48a055cd3429b55ca9f0

src/test/java/org/openrewrite/java/testing/mockito/JunitMockitoUpgradeIntegrationTest.java
```diff
@@ -89,26 +89,20 @@ public void usingAnnotationBasedMock() {
...
 
               import static org.mockito.Mockito.verify;
 
+              @ExtendWith(MockitoExtension.class)
               public class MockitoTests {
-                  private AutoCloseable mocks;
                   @Mock
                   List<String> mockedList;
 
-                  @BeforeEach
```
Jenson3210, line 107: Is it the location of your new recipe that caused this migration to now also take place as the class now has an `ExtendsWith` usage iso a `MockitoAnnotations` usage?

MBoegers: My recipe adds `@ExtendWith(MockitoExtension.class)` if `org.junit.jupiter` classes are used and here because `@Mock` is used. The extension makes the setup, so `MockitoAnnotations.openMocks(this);` is obsolete. Its the subsequent `org.openrewrite.java.testing.mockito.RemoveInitMocksIfRunnersSpecified` that removes the method. 
foor reference its in mockito.yml one line below

PR #849, opened 2025-11-21
Correcting MockWebServer migration for OkHttp

steve-aom-elliott:
## What's changed?

- New artifactId corrected from `mockwebserver3-junit5` to `mockwebserver3`
- In-progress work for changing how some classes are built / manipulated. Previously there were many methods existing on `MockResponse`. Now everything has to be switched to using `MockResponse.Builder` to change all these values.

## What's your motivation?

Compilation failures after previous setup's migration ran.

This PR covers most of the MockResponse setter and other methods where there are analogous methods on MockResponse.Builder, with the exception of `setSocketPolicy`. The `setSocketPolicy` would map onto one of the `onResponse` methods in the builder but this requires mapping `SocketPolicy` to `SocketEffect` and also choosing the right `onResponse` method. 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/UpdateMockWebServerMockResponse.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/UpdateMockWebServerMockResponse.java
@@ -100,8 +100,8 @@ public class UpdateMockWebServerMockResponse extends Recipe {
         List<Recipe> recipes = new ArrayList<>();
         for (Map.Entry<String, String> rep : REPLACEMENTS.entrySet()) {
             String methodPattern = OLD_MOCKRESPONSE_FQN.replace("$", ".") + "#" + rep.getKey();
+            recipes.add(new ChangeMethodInvocationReturnType(methodPattern, NEW_MOCKRESPONSE_FQN_BUILDER.replace("$", ".")));
             recipes.add(new ChangeMethodName(methodPattern, rep.getValue(), true, false));
-            recipes.add(new ChangeMethodInvocationReturnType(methodPattern, NEW_MOCKRESPONSE_FQN_BUILDER));
         }
         recipes.add(new ChangeType(OLD_MOCKRESPONSE_FQN, NEW_MOCKRESPONSE_FQN_BUILDER, true));
         recipes.add(new ChangePackage("okhttp3.mockwebserver", "mockwebserver3", false));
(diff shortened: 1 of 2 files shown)


steve-aom-elliott:
I'm going to try to look into whether I can set up a single visitor to perform all the type changes needed prior to using the `JavaTemplate` to write the `.build()` chaining in, as the problem I'm hitting here is that I'm changing `MockResponse` fields to `MockResponse.Builder` fields, return types from some of the methods of `MockResponse` to be on `MockResponse.Builder` instead, followed by introducing a `.build()` call that belongs to `MockResponse.Builder` but returns `MockResponse`, meaning `ChangeType` recipe effectively breaks the type attribution, but covers a lot of the scenarios otherwise that I'd want to be changing. Going to probably roll my own visitor that combines functionality from some of the other ones, or create a visitor that is capable of altering a method invocation's signature's parameter types, which I don't believe we have easily accessible a the moment.

src/main/java/org/openrewrite/java/testing/junit5/UpdateMockWebServerMockResponse.java
```diff
@@ -0,0 +1,226 @@
...
+                            .withDeclaringType(UpdateMockWebServerMockResponse.newMockResponseBuilderType)
+                            .withReturnType(UpdateMockWebServerMockResponse.newMockResponseBuilderType)
+                            .withName(newName)
+            );
+            return updated.withName(
+                    updated.getName()
+                            .withSimpleName(newName)
+                            .withType(updated.getMethodType())
+            );
+        }
```
timtebeek: Why the near duplicate method here and below?

dsgrieve: Ah! Good catch. I copied this to below for reuse but neglected to refactor this one. Fixing now.

src/test/java/org/openrewrite/java/testing/junit5/UpdateMockWebServerMockResponseTest.java
```diff
@@ -0,0 +1,240 @@
...
+                      Builder mrL = mockResponse.headersDelay(30L, TimeUnit.SECONDS);
+                      Builder mrM = mockResponse.code(500);
+                      Builder mrN = mockResponse.code(200);
+                      Builder mrO = mockResponse.status("OK");
+                      Builder mrP = mockResponse.trailers(new Headers.Builder().add("x-trailer:value").build());
+                      MockResponse.Builder mrQ = mockResponse.throttleBody(1024, 1, TimeUnit.SECONDS);
+                      Builder mrR = mockResponse.addPush(pushPromise);
+                      Builder mrS = mockResponse.settings(settings);
+                      Builder mrT = mockResponse.webSocketUpgrade(webSocketListener);
+                  }
```
timtebeek, line 234: The changes in imports and types here is because some of these are changed by map of renamed methods, whereas others are kept as is. In practice I'd expect these differences in types and imports to be rare, as the chained way of calling these methods is expected to be most prevalent.

PR #870, opened 2025-12-08
Add unit test showing null Methodinvocation type

greg-dennis:
## What's your motivation?

Add unit test showing null Methodinvocation type. This test fails in this repo but the very same test passes when added to the `rewrite` repo.

### Checklist


src/test/java/org/openrewrite/java/testing/cleanup/KotlinTestMethodShouldBeUnitTest.java
```diff
@@ -0,0 +1,80 @@
...
+            return method;
+        }
+    }
+
+    private static <T> T getOnly(Iterable<T> iterable) {
+        Iterator<T> iter = iterable.iterator();
+        T only = iter.next();
+        assertFalse(iter.hasNext());
+        return only;
+    }
```
Laurens-W, line 78:
```suggestion
    @Override
    public void defaults(RecipeSpec spec) {
        spec
          .parser(KotlinParser.builder());
    }

    @Test
    void showNullMethodInvocationType() {
        rewriteRun(
          //language=kotlin
          kotlin(
            """
              class ExampleTest {
                  fun myTest() = run {
                      return 1
                  }
              }
              """
          ));
    }
```
I've simplified it to this and am indeed seeing it fail in rewrite-testing-frameworks but succeed in rewrite

greg-dennis: Right, that simple example will show the symptom, but I found the failure to be more cryptic. The intention of the PR was to reveal the underlying issue: that the type of the `run` method invocation is null.

greg-dennis: @Laurens-W, there appears to be other code that "expects" the types of Kotlin method invocations to sometimes be null ([example](https://github.com/openrewrite/rewrite/blob/e118585d5a7081be8315c669127177e8b100002f/rewrite-kotlin/src/main/java/org/openrewrite/kotlin/internal/KotlinTreeParserVisitor.java#L2018)). If that is sometimes "expected" behavior, the following workaround might make sense: https://github.com/openrewrite/rewrite/pull/6409

Thoughts?

timtebeek:
Wondering if we'll see improvement here following
- https://github.com/openrewrite/rewrite/pull/6766

If so we can close this PR.

PR #883, opened 2025-12-23
Convert TestContainers raw types as part of upgrade

timtebeek:


timtebeek:
@ThomasVitale I noticed you had covered something similar in [io/arconia/rewrite/test/ConvertToRawType](https://github.com/arconia-io/arconia-migrations/blob/main/rewrite-test/src/main/java/io/arconia/rewrite/test/ConvertToRawType.java) ; perhaps it helps you to see this reduced implementation, even if I'm envious of your ability to leverage Java 21 there. ;)

PR #889, opened 2026-01-13
Transform `Assume` methods to without a direct counterpart to equivalent `Assumptions`

cortlepp:

@timtebeek 

### Checklist


cortlepp:
@timtebeek this is my current wip. but I have some questions before I can move forward:

1. I need junit on the classpath to write the template, but I'm not sure if that is intended
2. The normal test currently fails with `org.openrewrite.internal.RecipeRunException: java.lang.IllegalArgumentException: Unable to find classpath resource dependencies beginning with: 'opentest4j-1'`, maybe my added junit dependency messed something up there?
3. I'm not sure if I got the `Object...` case right. It's not important to me in particular, but I thought since we're already at it...
4. I'm not sure if the way this recipe is called from `org.openrewrite.java.testing.junit5.MigrateAssumptions` now actually works, since technically that recipe makes illegal changes (in the sense that they wouldn't compile). A solution would be to reimplement all of this in refaster, but I don't know if that is necessary

src/main/java/org/openrewrite/java/testing/junit5/AssumeToAssumptions.java
```diff
@@ -0,0 +1,40 @@
...
+            Assume.assumeNotNull(object);
+        }
+
+        @AfterTemplate
+        void after(Object... object) {
+            Arrays.stream(object).forEach(o -> Assumptions.assumeTrue(null != o));
+        }
+    }
+}
+
```
timtebeek:
```suggestion
}
```

PR #898, opened 2026-01-28
Skip TemporaryFolder conversion when used in RuleChain.

motlin:
## What's changed?

Added a precondition to `TemporaryFolderToTempDir` that skips conversion when `TemporaryFolder` is used within a `RuleChain`. The recipe now detects calls to `RuleChain.outerRule()` and `RuleChain.around()` that receive a `TemporaryFolder` argument, and skips the entire file in those cases.

## What's your motivation?

When `TemporaryFolder` is converted to `@TempDir File`, the resulting `File` type does not implement `TestRule`, which breaks `RuleChain.around()` calls that expect a `TestRule`. By skipping these cases, the existing `HandleExternalResourceRules` recipe can handle them instead by adding `@ExtendWith(ExternalResourceSupport.class)`.

The precondition operates at the file level—if *any* `TemporaryFolder` is used in a `RuleChain`, the entire file is skipped. This is conservative but simple. A more granular approach could convert standalone `TemporaryFolder` instances while skipping only those in `RuleChain`, but this adds complexity for an edge case (mixed usage in a single file).

1. **Granular per-field detection** - More complex; would require tracking which specific fields are in `RuleChain` and only skipping those

### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/JUnitParamsRunnerToParameterizedTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/JUnitParamsRunnerToParameterizedTest.java
@@ -175,7 +175,6 @@ class JUnitParamsRunnerToParameterizedTest implements RewriteTest {
                   public void personIsNamedAdult(int age, boolean valid) {
                   }
 
-
                   private static Object[] namedPeopleParams() {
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Perfect addition; thanks a lot for adding this precondition! Best to be safe indeed.

PR #899, opened 2026-01-28
Preserve throws clause when checked exceptions exist outside assertThrows.

motlin:
## What's changed?

The `ExpectedExceptionToAssertThrows` recipe now preserves the `throws` clause when statements *before* the `thrown.expect()` call throw checked exceptions.

Previously, the recipe unconditionally removed the `throws` clause from test methods when converting `ExpectedException` to `assertThrows()`. This caused compilation errors when code outside the `assertThrows()` lambda still threw checked exceptions.

**Key changes:**
- Added `findPredecessorStatements()` to track statements before the first `expect*()` call
- Added `statementsBeforeExpectThrowCheckedException()` to detect if those statements throw checked exceptions
- Added `isCheckedException()` to distinguish checked exceptions from `RuntimeException`/`Error`
- Modified `visitMethodDeclaration()` to preserve `throws` when pre-expect code throws checked exceptions

## What's your motivation?

When migrating tests that have setup code before `thrown.expect()`, the recipe was incorrectly removing the `throws` declaration, causing compilation failures.

Example:
```java
@Test
public void testMethod() throws InterruptedException {
    setup();  // throws InterruptedException - NOT wrapped in assertThrows
    this.thrown.expect(IllegalArgumentException.class);
    doSomething();
}
```

The recipe was removing `throws InterruptedException` even though `setup()` is not wrapped in the `assertThrows()` lambda.

The logic for detecting checked exceptions in `statementThrowsCheckedException()` - it recursively visits method invocations and constructor calls to check their thrown exception types. Please verify this covers all cases.

## Any additional context

The fix uses `JavaType.Method.getThrownExceptions()` to inspect method types, which requires proper type attribution. This should work in typical recipe execution contexts.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
@@ -102,7 +102,7 @@ public class ExpectedExceptionToAssertThrows extends Recipe {
                 List<NameTree> thrown = m.getThrows();
                 if (thrown != null && !thrown.isEmpty()) {
                     List<Statement> statementsBeforeExpect = getCursor().pollMessage(STATEMENTS_BEFORE_EXPECT_EXCEPTION);
-                    if (statementsBeforeExpect != null && statementsBeforeExpectThrowCheckedException(statementsBeforeExpect)) {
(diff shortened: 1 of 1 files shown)


timtebeek:
Glad to hear that's working well for you already! I'm admittedly behind on reviews, but that comes with landing Kotlin v2 and Python in the same week, as well as three online trainings. 😅 It's at the top of my list now, but first I'm out climbing. 🧗🏻

PR #901, opened 2026-01-29
Handle method references in JUnit Assert to Assertions migration.

motlin:
## What's changed?

Added support for converting JUnit 4 `Assert` method references to JUnit 5 `Assertions` method references in the `AssertToAssertions` recipe.

Before this change, method invocations like `Assert.assertTrue(true)` were converted correctly, but method references like `Stream.of(true, true).forEach(Assert::assertTrue)` were not being transformed.

The fix adds a `visitMemberReference` method to `AssertToAssertionsVisitor` that handles the conversion of method references.

## What's your motivation?

Method references are commonly used with streams and functional interfaces. Without this fix, code using patterns like `list.forEach(Assert::assertNotNull)` would not be migrated, leaving behind JUnit 4 dependencies after running `JUnit4to5Migration`.

## Any additional context

The `assertThat` method is explicitly excluded from conversion (consistent with the method invocation handling) since it's handled separately for Hamcrest/AssertJ migration.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/AssertToAssertions.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AssertToAssertions.java
@@ -30,7 +30,6 @@ import org.openrewrite.java.tree.*;
 import java.util.ArrayList;
 import java.util.List;
 
-import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
 
 public class AssertToAssertions extends Recipe {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/testing/junit5/AssertToAssertions.java
```diff
@@ -130,19 +131,45 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+            );
+
+            JavaType.Method methodType = mr.getMethodType();
+            if (methodType != null) {
+                methodType = methodType.withDeclaringType(targetType);
             }
-            J.Identifier receiver = (J.Identifier) method.getSelect();
-            if (!(receiver.getType() instanceof JavaType.FullyQualified)) {
+
+            return mr.withContaining(newContaining).withMethodType(methodType);
```
timtebeek: It's going to take an upstream fix, but I want to see if we can get this down to:
```suggestion
            if (isJunitAssertMethod(mr)) {
                doAfterVisit(new ChangeMethodTargetToStatic("org.junit.Assert " + mr.getReference().getSimpleName() + "(..)",
                        "org.junit.jupiter.api.Assertions", null, null, true)
                        .getVisitor());
            }
            return mr;
```

timtebeek: - depends on https://github.com/openrewrite/rewrite/pull/6645

PR #902, opened 2026-01-29
Add JUnit 4 to JUnit 5 exception class migrations.

motlin:
## What's changed?

Added a new recipe `MigrateAssertionFailedError` that migrates JUnit 4 assertion failure exception classes to their JUnit Jupiter equivalents:

- `junit.framework.AssertionFailedError` → `org.opentest4j.AssertionFailedError`
- `org.junit.ComparisonFailure` → `org.opentest4j.AssertionFailedError`

The recipe is included in `JUnit4to5Migration` so it runs automatically as part of the standard JUnit 4 to 5 migration.

## What's your motivation?

When migrating from JUnit 4 to JUnit 5, code that catches or throws JUnit 4's exception classes (`junit.framework.AssertionFailedError`, `org.junit.ComparisonFailure`) should be updated to use the OpenTest4J equivalent (`org.opentest4j.AssertionFailedError`), which is the standard assertion error used by JUnit Jupiter.

The choice to map both `AssertionFailedError` and `ComparisonFailure` to the same target class (`org.opentest4j.AssertionFailedError`). JUnit Jupiter doesn't have a direct equivalent to `ComparisonFailure` - the OpenTest4J `AssertionFailedError` handles comparison failures through its constructor that accepts expected/actual values.

Could have created separate recipes for each source class.

### Checklist


timtebeek (approved):
Perfect, thanks so much! Great to see all your findings trickle in. Much appreciated.

PR #908, opened 2026-02-05
Fix AssertThrowsOnLastStatement to preserve parameterized types when extracting arguments.

motlin:
## What's changed?

Fix `AssertThrowsOnLastStatement` to preserve parameterized types when extracting method arguments to local variables.

**Input code:**
```java
assertThrows(Exception.class, () -> {
    doA();
    testThing(Map.of(), List.of(), Set.of("value"));
});
```

**Replacement before fix:**
```java
doA();
Map of = Map.of();
List of1 = List.of();
Set of2 = Set.of("value");
assertThrows(Exception.class, () ->
    testThing(of, of1, of2));
```

**Replacement after fix:**
```java
doA();
Map<String, String> map = Map.of();
List<String> list = List.of();
Set<String> set = Set.of("value");
assertThrows(Exception.class, () ->
    testThing(map, list, set));
```

Improvements:
1. Preserve types (`Map` to `Map<String, String>`)
2. Improve variable names (`of`, `of1`, `of2` to `map`, `list`, and `set`) derived from method name.

## What's your motivation?

When migrating JUnit 4 tests with `@Test(expected = ...)` that use `Map.of()`, `List.of()`, or `Set.of()` as arguments, the recipe produces code with raw types that causes compilation errors and poor variable names.

- The `buildParameterizedTypeName()` method constructs type strings including generic parameters
- The `isShortFactoryMethodName()` check for `of`, `from`, `copyOf` to derive variable names from return types instead

The variable naming improvement could be split out or dropped.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
@@ -156,6 +156,7 @@ public class AssertThrowsOnLastStatement extends Recipe {
                             variableTypeFqn = e.getType();
                         } else if (e.getType() instanceof JavaType.Parameterized) {
                             JavaType.Parameterized paramType = (JavaType.Parameterized) e.getType();
+                            // TODO look into possibly employing `TypeUtils.toString()` here, possibly with some changes upstream allowing for non-fully-qualified names
                             variableTypeShort = buildParameterizedTypeName(paramType);
                             variableTypeFqn = paramType;
                             maybeAddImport(paramType.getFullyQualifiedName(), false);


src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -186,6 +201,48 @@ private String getVariableName(Expression e, Map<String, Integer> generatedVaria
...
+
+            private String getTypeBasedVariableName(J.MethodInvocation mi) {
+                JavaType type = mi.getType();
+                if (type instanceof JavaType.FullyQualified) {
+                    return ((JavaType.FullyQualified) type).getClassName();
+                }
+                return "x";
+            }
+
+            private String buildParameterizedTypeName(JavaType.Parameterized paramType) {
```
greg-at-moderne, line 218: I am wondering, how does it differ from `org.openrewrite.java.tree.TypeUtils#toString(org.openrewrite.java.tree.JavaType)`?

motlin: I tested it: TypeUtils.toString() returns fully qualified names (e.g., java.util.Map<java.lang.String, java.lang.String>), which doesn't work together with JavaTemplate string interpolation (#{}) which uses the value literally.

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
```diff
@@ -154,6 +154,16 @@ private Statement extractExpressionArguments(Statement lambdaStatement, List<Sta
                         if (e.getType() instanceof JavaType.Primitive) {
                             variableTypeShort = e.getType().toString();
                             variableTypeFqn = e.getType();
+                        } else if (e.getType() instanceof JavaType.Parameterized) {
+                            JavaType.Parameterized paramType = (JavaType.Parameterized) e.getType();
+                            variableTypeShort = buildParameterizedTypeName(paramType);
```
greg-at-moderne, line 160:
```suggestion
                            // TODO look into possibly employing `TypeUtils.toString()` here, possibly with some changes upstream allowing for non-fully-qualified names
                            variableTypeShort = buildParameterizedTypeName(paramType);
```

src/main/java/org/openrewrite/java/testing/junit5/AssertThrowsOnLastStatement.java
greg-at-moderne: OK. Makes sense. Thanks for checking. Added a note to the code.

PR #912, opened 2026-02-21
Add .contextSensitive() to Hamcrest conversion templates and improve type matching in `CollapseConsecutiveAssertThatStatements`

timtebeek:
- Add .contextSensitive() to JavaTemplate builders in HamcrestMatcherToAssertJ, HamcrestNotMatcherToAssertJ, and HamcrestIsMatcherToAssertJ to ensure templates are compiled in proper source context
- Improve CollapseConsecutiveAssertThatStatements with fallback for partially resolved SELF return types (e.g., when assertion methods return AbstractIterableAssert instead of ListAssert)
- Update test expectations: isNotNull() now collapses with other assertions and is removed by SimplifyRedundantAssertJChains
- Add test case for custom types to verify hamcrest-to-AssertJ conversion works

Fixes the issue where the AssertJ best practices recipe required multiple runs on Hamcrest test files before seeing chained assertions.


src/test/java/org/openrewrite/java/testing/assertj/AssertJBestPracticesTest.java
```diff
@@ -143,6 +143,58 @@ void biscuits() {
...
+                  import java.util.List;
+
+                  import static org.assertj.core.api.Assertions.assertThat;
+
+                  class BiscuitTest {
+                      void biscuits() {
+                          List<Biscuit> biscuits = List.of(new Biscuit("Ginger"), new Biscuit("Chocolate"), new Biscuit("Oatmeal"));
+                          assertThat(biscuits).hasSize(3);
+                          assertThat(biscuits).contains(new Biscuit("Chocolate"));
+                          assertThat(biscuits).doesNotContain(new Biscuit("Raisin"));
```
timtebeek: Ideally these are collapsed as well.

PR #966, opened 2026-04-08
Add @MockitoSettings(strictness = Strictness.WARN) for Mockito migration

bmuschko:
## Context

After migrating from Mockito 1.x, tests fail with `UnnecessaryStubbingException` because `@ExtendWith(MockitoExtension.class)` defaults to `Strictness.STRICT_STUBS`, whereas Mockito 1.x/2.x silently ignored unused stubbings (`Strictness.WARN`). Detecting which stubbings are truly unnecessary requires runtime analysis, so the practical fix is to add `@MockitoSettings(strictness = Strictness.WARN)` at the class level to preserve the original behavior.

The existing `MockitoJUnitToMockitoExtension` recipe already handled this when converting a `MockitoRule`, but had two gaps:

1. When `@ExtendWith(MockitoExtension.class)` was **already present** alongside a `MockitoRule`, the recipe removed the rule but did not add `@MockitoSettings`, silently changing strictness from WARN to STRICT_STUBS.
2. Classes that were **already fully migrated** (no `MockitoRule` remaining, just `@ExtendWith`) were not covered at all — no existing recipe would add the missing `@MockitoSettings`.

## Changes

1. **`MockitoJUnitToMockitoExtension`**: Decoupled the `@ExtendWith` and `@MockitoSettings` logic so that `@MockitoSettings` is now added even when `@ExtendWith` is already present (fixes gap 1).

2. **New `AddMockitoSettingsWithWarnStrictness` recipe**: Adds `@MockitoSettings(strictness = Strictness.WARN)` to classes that have `@ExtendWith(MockitoExtension.class)` but no `@MockitoSettings` annotation. This covers classes already migrated without a `MockitoRule` to convert (fixes gap 2). Added to the `Mockito1to3Migration` chain.


src/main/resources/META-INF/rewrite/mockito.yml
```diff
@@ -184,6 +184,7 @@ recipeList:
   - org.openrewrite.java.testing.mockito.ReplacePowerMockito
   - org.openrewrite.java.testing.junit5.MockitoJUnitToMockitoExtension
   - org.openrewrite.java.testing.mockito.AddMockitoExtensionIfAnnotationsUsed
+  - org.openrewrite.java.testing.mockito.AddMockitoSettingsWithWarnStrictness
```
timtebeek: I think we'd want to guard this with a precondition to only apply when the module has a dependency on a pre-version-3 of Mockito right?

What I'm trying to avoid is that folks who never used Mockito 1, 2 or 3, but say started at 4, then run our recipe to upgrade to 5, and suddenly see `MockitoSettings` restore an old default they had never used to begin with, reducing the strictness of their Mockito usage.

Would you agree with the above scenario & assessment?

bmuschko:
I'll make this a draft for now. It's still failing on CI anyway.

src/main/resources/META-INF/rewrite/mockito.yml
bmuschko: Makes sense. I think we have two options here:

1. I made the recipe a standalone recipe so that end users have to run it on-demand. It will only be applicable to projects that are interested in Mockito 1.x/2.x migrations.
2. I am not a fan of this option but here it is. We could add a version parameter to the Java recipe (e.g. a `String maxMockitoVersion` option) and implement a scanning precondition to check the dependency version programmatically.

WDYT?

timtebeek: Would we not be able to guard this using the preexisting https://docs.openrewrite.org/recipes/java/dependencies/search/modulehasdependency as a yaml precondition, which uses a version pattern only including a range of older versions?

bmuschko: done

timtebeek (approved):
Thanks for catching these cases with the preconditioned addition!

PR #977, opened 2026-04-28
Cleanup Mockito AutoCloseable fields

sullis:
## Summary

Cleanup Mockito `AutoCloseable` fields in `RemoveInitMocksIfRunnersSpecified`.

### Changes

- **`RemoveInitMocksIfRunnersSpecified`**: Fix visitor to also handle nested/inner class declarations by checking if the current class is enclosed within another class (`getParentTreeCursor().firstEnclosing(J.ClassDeclaration.class) != null`). Previously, the recipe would skip inner classes that didn't themselves carry the Mockito extension/runner annotations.

- **`MockitoBestPracticesTest`**: Add new test cases covering `AutoCloseable` field cleanup scenarios.

- **`RemoveInitMocksIfRunnersSpecifiedTest`**: Add additional test cases for nested class handling.


timtebeek (approved):
Thanks a lot! Good to see these nested classes covered as well.

PR #993, opened 2026-05-06
Cleanup junit vintage engine

sullis:
## What's changed?

update junit vintage cleanup logic

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
@@ -542,6 +542,128 @@ class JUnit5MigrationTest implements RewriteTest {
     }
 
     @Test
+    void removeJunitVintageEngineFromMavenBuild() {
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/java/testing/junit5/JUnit5MigrationTest.java
```diff
@@ -541,6 +541,128 @@ public void hello() {
...
+          //language=xml
+          pomXml(
+            """
+              <project>
+                  <modelVersion>4.0.0</modelVersion>
+                  <groupId>com.example</groupId>
+                  <artifactId>my-project</artifactId>
+                  <version>1.0.0</version>
+                  <dependencies>
+                      <dependency>
```
timtebeek, line 556: Given the changes above I'd have expected the tests to use managed dependencies. What was the reason you picked non-managed here?

timtebeek (approved):
Great to see, thanks! Good to also have test coverage for the non-managed case added.

PR #995, opened 2026-05-07
Skip wrapper-primitive in `UseAssertSame`; rewrite via `AssertTrueComparisonToAssertEquals`

motlin:
## What's changed?

For an input like:

```java
Map<String, Double> map = new HashMap<>();
double primitive = 0.5;
assertTrue(map.get("k") == primitive);
```

`UseAssertSame` was rewriting it to:

```java
assertSame(map.get("k"), primitive);
```

It should rewrite to:

```java
assertEquals(map.get("k"), primitive);
```

To get there:

- `UseAssertSame` now skips when either operand is a Java primitive (with `Null` and `String` carved out, since OpenRewrite encodes both as `JavaType.Primitive` enum values but their `==` is reference equality).
- `AssertTrueComparisonToAssertEquals` now also accepts mixed primitive-wrapper operands. Wrapper-wrapper is still rejected.

## What's your motivation?

- PR #992 added a primitive-primitive skip to `UseAssertSame`, which catches `assertTrue(123L == 123)`.

The rewritten `assertSame` fails at runtime because Java's `==` unboxes, but `assertSame` autoboxes the primitive into a fresh wrapper and compares by reference:

```
expected: java.lang.Double@4130a648<0.5>
but was:  java.lang.Double@61ff6a49<0.5>
```

Wrapper-wrapper is left to `UseAssertSame` rather than swept into `AssertTrueComparisonToAssertEquals`, since rewriting `Double == Double` to value equality would silently change semantics on intentional reference checks.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/cleanup/AssertTrueComparisonToAssertEquals.java
+++ b/src/main/java/org/openrewrite/java/testing/cleanup/AssertTrueComparisonToAssertEquals.java
@@ -28,6 +28,7 @@ import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
 
 public class AssertTrueComparisonToAssertEquals extends Recipe {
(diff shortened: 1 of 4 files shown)


timtebeek (approved):
Thanks a lot! More comprehensive than the previous attempt indeed.

PR #996, opened 2026-05-09
Fix `JUnit4to5Migration` for classes that extend `org.junit.Assert`

motlin:
The JUnit 4 to 5 migration produces broken code when a class `extends org.junit.Assert` and uses unqualified inherited assertion methods. The `JUnit4to5Migration` aggregator:

- adds Jupiter static imports for `assertEquals` / `assertTrue`,
- reorders the message argument from JUnit 4 position (first) to Jupiter position (last),
- but leaves `import org.junit.Assert` and `extends Assert` unchanged.

Because inherited methods shadow static imports in Java, the reordered call `assertEquals(1, 1, "expected message")` resolves against JUnit 4's `Assert.assertEquals`, which has no `(int, int, String)` overload, so the migrated code will not compile.

I'm not sure what the proper fix should be, so I'm just submitting a test that demonstrates the broken output. The `after` block reflects what the recipes actually produce today, so the test passes and CI stays green. Happy to help with the fix, but I don't have a clear idea of what it ought to be.


timtebeek:
Thanks @motlin .. I'd marked this as draft to signify it should not be merged as is just yet. Instead we'll look to fix the recipe before merging.

PR #1009, opened 2026-05-28
fix: retain doNothing() stubs inside lambda bodies to prevent uncompilable code

### LINKED ISSUE #1008: fix: RemoveDoNothingForDefaultMocks incorrectly removes doNothing() in contexts where removal produces uncompilable code
### What version of OpenRewrite are you using?
- rewrite-testing-frameworks 3.36.0
- rewrite-maven-plugin 6.40.0

### How are you running OpenRewrite?
Maven plugin, single module project.

```
<plugin>
  <groupId>org.openrewrite.maven</groupId>
  <artifactId>rewrite-maven-plugin</artifactId>
  <version>6.40.0</version>
  <configuration>
    <activeRecipes>
      <recipe>org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_5</recipe>
    </activeRecipes>
  </configuration>
</plugin>
```

### What is the smallest, simplest way to reproduce the problem?

```
import org.junit.Test;
import org.junit.runner.RunWith;
import org.mockito.Mock;
import org.mockito.junit.MockitoJUnitRunner;
import java.io.BufferedWriter;
import java.util.List;
import static org.mockito.Mockito.doNothing;

@RunWith(MockitoJUnitRunner.class)
class MyTest {
    @Mock
    private BufferedWriter bufferedWriter;

    @Test
    public void test() {
        List<String> paths = List.of("/path/1", "/path/2");
        paths.forEach(path -> doNothing().when(bufferedWriter).write(path));
    }
}
```

### What did you expect to see?
No change, doNothing() as an expression lambda body or switch expression arm should be retained.
The recipe cannot safely remove it because removing the method invocation leaves a dangling `->`
producing uncompilable code. Examples:

```
paths.forEach(path ->);       // Expression lambda do not compile
case "ACTIVE" ->               // Switch arm do not compile
```

Block lambda and switch block arm cases are not affected, as removal leaves an empty block which is valid Java and can be cleaned up by static analysis.

### What did you see instead?

```
@Test
public void test() {
    List<String> paths = List.of("/path/1", "/path/2");
    paths.forEach(path ->);  // does not compile
}
```

Full stack trace: N/A because no error thrown, recipe silently produces uncompilable code.
Are you interested in contributing a fix? Yes, a fix and tests are ready as a PR.

darrencilia:
## What's changed?
Added a guard in `RemoveDoNothingForDefaultMocks` to skip removal of `doNothing()` stubs when they appear as an expression lambda body or a switch expression arm,  preventing a dangling `->` which does not compile.

Block lambda and switch block arm cases are not guarded as removal leaves an empty block which is valid Java and can be cleaned up downstream by static analysis.

## What's your motivation?
Fixes #1008

When `doNothing().when(mock).method(args)` is the expression body of a lambda or 
a switch expression arm, removing it leaves a dangling `->` which does not compile.

The two helpers `isExpressionLambdaBody()` and  `isSwitchExpressionArm()` confirming the cursor level is correct for each case.

An alternative would be to remove the entire enclosing construct when the body becomes empty. However this is a significantly more complex AST transformation and risks unintended behaviour changes. Retaining is the safer option for compile-breaking cases.

## Any additional context
The bug surfaces through any recipe that chains through `RemoveDoNothingForDefaultMocks`, including `UpgradeSpringBoot_3_5` and all other composite recipes that include `RemoveDoNothingForDefaultMocks` as part of their stack.


darrencilia:
I have updated the oriignal comment and Issue to include my latest commit changes.

For this I added 6 tests:
- `retainsDoNothingInsideLambdaBody` - expression lambda, recipe makes no change
- `removesDoNothingInLambdaBlockLeavingEmptyBlock` - block lambda sole statement, recipe removes and leaves empty block
- `removesDoNothingWhenNotSoleStatementInLambdaBlock` - block lambda multiple statements, recipe removes correctly
- `retainsDoNothingInSwitchExpressionArm` - switch arrow arm, recipe makes no change
- `removesDoNothingInSwitchExpressionArmBlockLeavingEmptyBlock` - switch block arm sole statement, recipe removes and leaves empty block
- `removesDoNothingWhenNotSoleStatementInSwitchExpressionArmBlock` - switch block arm multiple statements, recipe removes correctly

src/main/java/org/openrewrite/java/testing/mockito/RemoveDoNothingForDefaultMocks.java
```diff
@@ -74,6 +74,10 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, Ex
                     public J.@Nullable MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                         J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
                         if (mi != null && isDoNothingOnMockField(mi)) {
+                            // Retain becuase if removed would leave a dangling -> producing uncompilable code
```
timtebeek:
```suggestion
                            // Retain because if removed would leave a dangling -> producing uncompilable code
```

timtebeek (approved):
Thanks again for the help here and the clear test examples. I've inline the small methods just to have fewer indirections when reading the code. We'll merge as soon as CI is green.

PR #1013, opened 2026-06-08
Expand PowerMockWhiteboxToJavaReflection to cover more Whitebox APIs

MBoegers:
The PowerMock → Mockito migration removes the `org.powermock:powermock-reflect` dependency, but `PowerMockWhiteboxToJavaReflection` previously migrated only `setInternalState`/`getInternalState`/`invokeMethod`, leaving the rest of the `Whitebox` API as uncompilable references after the dependency is dropped.

This expands the recipe to also migrate, to plain `java.lang.reflect`:

- `getField` / `getMethod` → `getDeclaredField` / `getDeclaredMethod` + `setAccessible`
- `invokeConstructor` (both overloads) → `Constructor.newInstance`
- static `invokeMethod(Class, …)` → `invoke(null, …)`
- the `Class where` `getInternalState` / `setInternalState` overloads
- non-literal field/method names (fall back to a generated local)

Notes:

- Primitive results are cast through their wrapper (`(Integer)` etc.), since a direct `(int) object` cast does not compile.
- The explicit-`Class[]` / varargs-array overloads of `invokeMethod`/`invokeConstructor` are intentionally left untouched — there is no faithful mechanical translation (a follow-up recipe flags them).
- Field/constructor lookups use `getDeclaredField`/`getDeclaredConstructor` on the *named* class, which differs from PowerMock for members inherited from a superclass (documented in the recipe description).

Adds unit tests covering each new signature plus the primitive-cast and array-overload cases.


src/main/java/org/openrewrite/java/testing/mockito/PowerMockWhiteboxToJavaReflection.java
```diff
@@ -139,18 +152,92 @@ public J.Block visitBlock(J.Block block, ExecutionContext ctx) {
             return b;
         }
 
-        private @Nullable String buildReplacementTemplate(Statement statement, J.MethodInvocation mi,
+        private JavaType.@Nullable Method resolveFor(J.MethodInvocation mi) {
```
steve-aom-elliott, line 175: Might suggest making the name more reflect the fact it's just for the `invoke` methods

src/main/java/org/openrewrite/java/testing/mockito/PowerMockWhiteboxToJavaReflection.java
```diff
@@ -322,10 +616,52 @@ private Object[] buildInvokeMethodArgs(List<Expression> args, JavaType.@Nullable
...
+                if (!param.getTypeParameters().isEmpty()) {
+                    JavaType elem = param.getTypeParameters().get(0);
+                    if (elem instanceof JavaType.FullyQualified) {
+                        return (JavaType.FullyQualified) elem;
+                    }
+                }
+            }
+            return null;
+        }
+
```
steve-aom-elliott, line 683: Can be collapsed somewhat by using `TypeUtils.asParameterized(..)` and `TypeUtils.asFullyQualified(..)`, though this also works.

src/main/java/org/openrewrite/java/testing/mockito/PowerMockWhiteboxToJavaReflection.java
```diff
@@ -297,18 +581,28 @@ private Object[] buildInvokeMethodArgs(List<Expression> args, JavaType.@Nullable
...
+        private JavaType.@Nullable Method findUniqueMethod(JavaType.@Nullable FullyQualified targetType,
+                                                           @Nullable String methodName, int expectedParamCount) {
+            if (targetType == null || methodName == null) {
                 return null;
             }
-            int expectedParamCount = args.size() - 2;
             JavaType.Method match = null;
-            for (JavaType.FullyQualified current = (JavaType.FullyQualified) targetType;
-                 current != null; current = current.getSupertype()) {
+            for (JavaType.FullyQualified current = targetType; current != null; current = current.getSupertype()) {
```
steve-aom-elliott: Can `JavaType.FullyQualified`'s `getVisibleMethods()` be used to collapse some of this?

MBoegers:
I will reshape the PR.

PR #1014, opened 2026-06-08
Migrate removed LocalStack Service enum and getEndpointOverride in Testcontainers 2.x migration

MBoegers:
## Why
`Testcontainers2Migration` renames `org.testcontainers.containers.localstack.LocalStackContainer` → `org.testcontainers.localstack.LocalStackContainer`, but Testcontainers 2.x removed the nested `Service` enum and the `getEndpointOverride(...)` method from the new class. So after the type rename, `LocalStackContainer.Service.SQS` and `getEndpointOverride(...)` become **unresolved references**, breaking compilation in both Java and Kotlin.

## What
Adds a `Testcontainers2LocalStack` step to `Testcontainers2Migration`, run **before** the type rename (so it matches the old type). Fully declarative, composed from core recipes:

- **`ReplaceConstant`** (one per `Service` constant) — replaces `LocalStackContainer.Service.X` with its LocalStack service-name string (e.g. `"sqs"`), including the non-trivial names `API_GATEWAY`→`apigateway`, `DYNAMODB_STREAMS`→`dynamodbstreams`, `CLOUDWATCHLOGS`→`logs`. This makes `withServices(...)` compile against the 2.x `withServices(String...)` API.
- **`ChangeMethodName` + `DeleteMethodArgument`** — `getEndpointOverride(service)` → `getEndpoint()`.

Covered by Java + Kotlin tests, including an end-to-end reproduction through the full `Testcontainers2Migration`.

Part of moderneinc/customer-requests#2437 (the unresolved-references half; the separate ktlint-comment-removal issue is not addressed here).


src/test/java/org/openrewrite/java/testing/testcontainers/Testcontainers2LocalStackTest.java
```diff
@@ -0,0 +1,258 @@
+/*
+ * Copyright 2025 the original author or authors.
```
steve-aom-elliott: 2026

steve-aom-elliott (approved):
Approved, probably just bump the year up in the copyright is all

PR #1015, opened 2026-06-08
Hoist PowerMock Whitebox calls used in expression position

MBoegers:
Ports the expression-position hoisting logic onto the current split Whitebox reflection recipes.

Adds a second pass in the shared `WhiteboxToReflectionVisitor` that handles `Whitebox` result calls nested inside an enclosing statement (`return`, method arguments, `if` conditions, …), which the statement/variable-declaration pass cannot rewrite in place because reflection needs a multi-statement preamble.

For each such call, the reflection preamble and a temp local are hoisted before the enclosing statement, and the call is replaced by a reference to the temp. Calls in short-circuit (`&&`/`||`) right operands or ternary branches are left untouched, since hoisting would change whether the reflective call executes.

Keeps the split per-API recipe structure now on `main`, updates the aggregate recipe metadata to describe the hoisting behavior, and retains 6 unit tests covering: a `return`, a method-call argument, an `if` condition, a `for`-loop body (hoisted into the loop body, not the method block), two nested calls in one statement (preserved order), and a short-circuit-guarded call left unchanged.


timtebeek:
Given the work we've done on this front in the past week, is there anything left to do here?

timtebeek:
@copilot resolve the merge conflicts in this pull request

MBoegers:
I close this PR without merging. The PRs from last week should deliver the main value, the proposed changes here are huge in terms of code changes with questionable value.

PR #1016, opened 2026-06-08
Flag un-migratable PowerMock powermock-reflect usage with a comment

MBoegers:
Flags any `org.powermock.reflect.Whitebox` usage that the PowerMock → Mockito migration cannot rewrite, so removing the `powermock-reflect` dependency never silently leaves uncompilable code without a trace.

Rather than a bespoke recipe, this wires the existing declarative `org.openrewrite.java.AddCommentToMethodInvocations` into `ReplacePowerMockito`, after the `Whitebox` migration and before the PowerMock dependency removal:

```yaml
- org.openrewrite.java.AddCommentToMethodInvocations:
    methodPattern: org.powermock.reflect.Whitebox *(..)
    comment: " PowerMock `Whitebox` call could not be automatically migrated to reflection; migrate manually "
```

Any remaining `Whitebox` call with no faithful reflection equivalent (type-based `getInternalState`/`setInternalState`, `newInstance`, the name-less "best guess" `invokeMethod`, …) gets an idempotent inline comment — `AddCommentToMethodInvocations` checks for an existing identical comment before adding, so repeated migration cycles don't duplicate it.

Adds an end-to-end `ReplacePowerMockitoIntegrationTest` case (with `powermock-reflect-1` on the classpath) exercising migrate-and-flag through the full `ReplacePowerMockito` pipeline.


timtebeek:
Given the work we've done on this front in the past week, is there anything left to flag / do here?

PR #1017, opened 2026-06-09
Don't add MockitoExtension during migration when unsafe (#875)

### LINKED ISSUE #875: MockitoExtension gets added even when not necessary
I am using

- OpenRewrite v8.16.1
- Maven plugin v6.25.0
- rewrite-recipe-bom 3.20.0

## How are you running OpenRewrite?
I am using the Maven plugin, and my project is a single module project.

Consider these classes:
```java
@ExtendWith(MockitoExtension.class)
abstract class BaseTest {
}
```
```java
class ChildTest extends BaseTest {
	@Mock
	Object underTest;

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
```java
class ExplicitInitTest {
	@Mock
	Object underTest;

	@BeforeEach
	void setup() {
		MockitoAnnotations.openMocks(this);
	}

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
(the next one only with Spring Boot < 4.0, I think)
```java
@ExtendWith(SpringExtension.class)
class SpringExtensionTest {
	@Mock
	Object underTest;

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
```java
@Target({ ElementType.TYPE })
@Retention(RetentionPolicy.RUNTIME)
@ExtendWith(MockitoExtension.class)
@interface MyMockitoAnnotation {
}
```
```java
@MyMockitoAnnotation
class CustomAnnotationTest {
	@Mock
	Object underTest;

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
These tests pass without an explicit `@ExtendWith(MockitoExtension.class)` on the test class itself.

OpenRewrite should not alter these classes when running `AddMockitoExtensionIfAnnotationsUsed`, nor Java or Spring Boot upgrades.

## What did you see instead?
The test classes all see `@ExtendWith(MockitoExtension.class)` being added:
```java
@ExtendWith(MockitoExtension.class)
public class ChildTest extends BaseTest {
	@Mock
	Object underTest;

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
```java
@ExtendWith(MockitoExtension.class)
public class ExplicitInitTest {
	@Mock
	Object underTest;

	@BeforeEach
	void setup() {
		MockitoAnnotations.openMocks(this);
	}

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
```java
@ExtendWith(SpringExtension.class)
@ExtendWith(MockitoExtension.class)
public class SpringExtensionTest {
	@Mock
	Object underTest;

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
```java
@ExtendWith(MockitoExtension.class)
@MyMockitoAnnotation
class CustomAnnotationTest {
	@Mock
	Object underTest;

	@Test
	void myTest() {
		assertNotNull(underTest);
	}
}
```
The extension shouldn’t be added when one of the following is true:
* a parent class already has it
* there is an explicit call to `MockitoAnnotations.initMocks()` or `MockitoAnnotations.openMocks()` (it could be in a parent class too!)
* using the `SpringExtension` in SB < 3.4, maybe? (in 3.4 this usage appears to have been _tacitly_ deprecated, and [removed in SB 4](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#mockito-captor-and-mock-annotations), hence #848 )
* the test class (or one of its parents) is meta-annotated with one of these extensions
* maybe other conditions?

## What is the full stack trace of any errors you encountered?
```
[WARNING] Changes have been made to src\test\java\…\ChildTest.java by:
[WARNING]         org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_5
[WARNING]             org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_4
[WARNING]                 org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_3
[WARNING]                     org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2
[WARNING]                         org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_1
[WARNING]                             org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_0
[WARNING]                                 org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_7
[WARNING]                                     org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_6
[WARNING]                                         org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_5
[WARNING]                                             org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_4
[WARNING]                                                 org.openrewrite.java.spring.boot2.SpringBoot2JUnit4to5Migration
[WARNING]                                                     org.openrewrite.java.testing.junit5.JUnit4to5Migration
[WARNING]                                                         org.openrewrite.java.testing.junit5.UseMockitoExtension
[WARNING]                                                             org.openrewrite.java.testing.mockito.Mockito1to4Migration
[WARNING]                                                                 org.openrewrite.java.testing.mockito.Mockito1to3Migration
[WARNING]                                                                     org.openrewrite.java.testing.mockito.CleanupMockitoImports
[WARNING]                                                                     org.openrewrite.java.testing.mockito.AddMockitoExtensionIfAnnotationsUsed
```
(log is similar for the other classes)

I’m not too sure how to implement all these negative cases, especially the ones related to parent classes.

Maybe for the time being, a temporary solution would be to edge on the safe side and only make the change when `SpringExtension` is used, and only during the Spring Boot 4 migration? This appears to have been the main target from #848 (by @MBoegers).

timtebeek:
- Fixes #875

## Problem
`AddMockitoExtensionIfAnnotationsUsed` (run as part of `Mockito1to3Migration`) added `@ExtendWith(MockitoExtension.class)` to any test using `@Mock`/`@Spy`/`@Captor`/`@InjectMocks`, even when Mockito support was already provided another way. This was at best redundant and at worst broke the test lifecycle — e.g. when `SpringExtension` is present (it already initializes `@Mock` fields), when a parent class/interface/meta-annotation provides the extension, or when `MockitoAnnotations.openMocks(this)` is used explicitly.

## Changes
**Removed from the migration (do no harm)**
- `AddMockitoExtensionIfAnnotationsUsed` and its companion `AddMockitoJupiterDependency` are no longer applied as part of `Mockito1to3Migration`. They are kept as **commented-out** entries with an explanatory comment so the intent is captured and they aren't re-added without first teaching the recipes about these cases.

**Hardened the standalone recipe** (`AddMockitoExtensionIfAnnotationsUsed`)
- Skips when any `@ExtendWith`/`@RunWith` is already present (covers `SpringExtension`).
- Skips non-concrete classes (interfaces, enums, annotations, `abstract` bases).
- Only annotates classes that themselves declare a Mockito annotation.

**Preserved the PowerMock migration path**
- `PowerMockRunnerDelegateToRunWith` now replaces `@RunWith(PowerMockRunner.class)` with `@RunWith(MockitoJUnitRunner.class)` when the class uses Mockito annotations (instead of relying on the now-removed generic auto-add), and still removes it entirely otherwise.

**Tests**
- Added cases for the hardened recipe (SpringExtension present, abstract class, interface + helper without mocks).
- Added standalone `AddMockitoJupiterDependencyTest` to keep that recipe covered.
- Updated migration/integration tests to the new, correct behavior (explicit `initMocks` is now preserved/migrated to `openMocks` + `AutoCloseable` instead of being replaced by the extension).

## Notes / not covered
The hardened recipe still does not detect the **inheritance**, **meta-annotation** (e.g. `@MockitoSettings`), or **explicit-`initMocks`** cases — these now only matter to someone invoking the standalone recipe directly, since it no longer runs in the migration.

Full test suite passes locally.

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
+++ b/src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
@@ -1800,8 +1800,11 @@ class JMockitExpectationsToMockitoTest implements RewriteTest {
               """,
             """
               import org.junit.jupiter.api.Test;
+              import org.junit.jupiter.api.extension.ExtendWith;
               import org.mockito.Mock;
+              import org.mockito.junit.jupiter.MockitoExtension;
 
+              @ExtendWith(MockitoExtension.class)
               class MyTest {
                   @Mock
                   private MyInterface myMock;
--- a/src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
+++ b/src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
@@ -1800,11 +1800,8 @@ class JMockitExpectationsToMockitoTest implements RewriteTest {
               """,
             """
               import org.junit.jupiter.api.Test;
-              import org.junit.jupiter.api.extension.ExtendWith;
               import org.mockito.Mock;
-              import org.mockito.junit.jupiter.MockitoExtension;
 
-              @ExtendWith(MockitoExtension.class)
               class MyTest {
                   @Mock
                   private MyInterface myMock;
--- a/src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
@@ -169,7 +169,7 @@ public class AddMockitoExtensionIfAnnotationsUsed extends Recipe {
     private static boolean shouldSkip(J.ClassDeclaration classDecl) {
         // Interfaces, enums, annotations and abstract bases are not concrete Mockito test classes
         if (classDecl.getKind() != J.ClassDeclaration.Kind.Type.Class ||
-                classDecl.getModifiers().stream().anyMatch(m -> m.getType() == J.Modifier.Type.Abstract)) {
+                classDecl.hasModifier(J.Modifier.Type.Abstract)) {
             return true;
         }
         // Only add to classes that declare Mockito annotations. The search spans the class subtree on purpose: a
--- a/src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/AddMockitoExtensionIfAnnotationsUsed.java
@@ -27,12 +27,16 @@ import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.search.FindAnnotations;
 import org.openrewrite.java.search.FindTypes;
 import org.openrewrite.java.search.IsLikelyTest;
+import org.openrewrite.java.service.AnnotationService;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.TypeUtils;
 import org.openrewrite.kotlin.KotlinIsoVisitor;
 import org.openrewrite.kotlin.KotlinParser;
 import org.openrewrite.kotlin.KotlinTemplate;
 import org.openrewrite.kotlin.tree.K;
+
+import java.util.Arrays;
+import java.util.List;
 
 import static java.util.Comparator.comparing;
 import static org.openrewrite.Preconditions.*;
@@ -41,6 +45,13 @@ public class AddMockitoExtensionIfAnnotationsUsed extends Recipe {
 
     private static final String EXTEND_WITH = "org.junit.jupiter.api.extension.ExtendWith";
     private static final String RUN_WITH = "org.junit.runner.RunWith";
+    // Explicit list rather than a wildcard like `org.mockito.*`, which would also match non-injection annotations
+    // such as `@DoNotMock`, `@CheckReturnValue` or `@Incubating` and cause false positives.
+    private static final List<String> MOCKITO_FIELD_ANNOTATIONS = Arrays.asList(
+            "org.mockito.Captor",
+            "org.mockito.Mock",
+            "org.mockito.Spy",
+            "org.mockito.InjectMocks");
 
     @Getter
     final String displayName = "Adds Mockito extensions to Mockito tests";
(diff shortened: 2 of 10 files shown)


src/test/java/org/openrewrite/java/testing/jmockit/JMockitExpectationsToMockitoTest.java
```diff
@@ -1800,11 +1800,8 @@ interface MyInterface {
               """,
             """
               import org.junit.jupiter.api.Test;
-              import org.junit.jupiter.api.extension.ExtendWith;
               import org.mockito.Mock;
-              import org.mockito.junit.jupiter.MockitoExtension;
 
-              @ExtendWith(MockitoExtension.class)
```
DidierLoiseau, line 1807: I think you’ll still need to add the extension when migrating from JMockit to Mockito, otherwise the migration won’t work.

(I’m not familiar with that migration though)

timtebeek: Good catch — you're right for one of the two JMockit styles. I dug into it:

- The `JMockitToMockito` migration already carries the extension over for the **annotation-based** style via `ChangeType`: `@ExtendWith(JMockitExtension.class)` → `@ExtendWith(MockitoExtension.class)` and `@RunWith(JMockit.class)` → `@RunWith(MockitoJUnitRunner.class)`. Almost all of the JMockit tests use this style, and they keep working.
- The gap is the **javaagent style** — `@Mocked` with no `@ExtendWith`/`@RunWith` (which is exactly this test). There's no extension to convert, so the migrated `@Mock` fields would never be initialized.

I've pushed a fix: re-added the (now hardened) `AddMockitoExtensionIfAnnotationsUsed` scoped to `JMockitToMockito` only, plus a version-correct `mockito-junit-jupiter:5.x` dependency. Because the hardened recipe bails when any `@ExtendWith`/`@RunWith` is already present, the annotation-based tests (which already have `MockitoExtension` after `ChangeType`) are left untouched, while the agent-based case now gets the extension. Restored this test's expected output accordingly.

PR #1021, opened 2026-06-16
Refactor `PowerMockWhiteboxToJavaReflection` into per-API recipes

MBoegers:
## What

Refactors the released `PowerMockWhiteboxToJavaReflection` into three focused per-API recipes, aggregated under the existing recipe name as a declarative composite:

- `PowerMockWhiteboxSetInternalStateToJavaReflection`
- `PowerMockWhiteboxGetInternalStateToJavaReflection`
- `PowerMockWhiteboxInvokeMethodToJavaReflection`

The released `org.openrewrite.java.testing.mockito.PowerMockWhiteboxToJavaReflection` name is preserved as a YAML composite in `powermockito.yml`, so by-name usage and the `ReplacePowerMockito` reference are unaffected. **Behavior is unchanged** — the same three APIs (`setInternalState`, `getInternalState`, `invokeMethod`) migrate exactly as before.

`PowerMockWhiteboxSetInternalStateToJavaReflection` also handles the 4-arg `Whitebox.setInternalState(target, field, value, Class)` where-overload, using the supplied `Class` argument as the `getDeclaredField` receiver to correctly resolve fields declared on a superclass.

## How

- Shared statement-rewriting machinery lives in a package-private `WhiteboxToReflectionVisitor`, configured per recipe (constructor-provided `MethodMatcher` + target `java.lang.reflect` import).
- Each recipe gates its visitor with `UsesMethod` on its own matcher.
- `WhiteboxToReflectionVisitor` provides both `fieldLookupPrefix` (3-arg path: `target.getClass().getDeclaredField`) and `fieldLookupPrefixWhere` (4-arg path: `whereClass.getDeclaredField`) helpers.

## Tests

- Each recipe has its own test class with a `@DocumentExample`; the existing test cases are preserved verbatim, split by API.
- `PowerMockWhiteboxToJavaReflectionTest` is retained as the aggregate test (via `recipeFromResources`) plus a mixed-API test exercising the composite.
- `setInternalStateWithWhereClass` and `setInternalStateWithWhereClassVariable` tests cover the 4-arg overload in both the per-recipe and aggregate test classes.
- `recipeCsvValidate` passes.

## Notes

Incorporates the 4-arg `setInternalState` support from main alongside the per-API refactor. Broader Whitebox coverage (static `invokeMethod`, `getField` / `getMethod` / `invokeConstructor`) will follow in a separate PR.

PR #1023, opened 2026-06-16
Support 4-arg Whitebox.setInternalState(target, field, value, Class)

MBoegers:
## What

Extends `PowerMockWhiteboxToJavaReflection` to also handle the 4-arg overload
`Whitebox.setInternalState(target, "field", value, Class)`. The 3-arg form was already
migrated; this overload was left untouched.

## Why

The 4th argument is the **declaring class** to look the field up on, and it matters because
`getDeclaredField` does **not** walk the class hierarchy. For an inherited field,
`child.getClass().getDeclaredField("name")` throws `NoSuchFieldException` — only
`Parent.class.getDeclaredField("name")` resolves it. This overload is exactly what test code
uses to reach a field declared on a superclass.

The only semantic difference from the 3-arg path is the **receiver of `getDeclaredField`**:

| Overload | Generated lookup receiver |
|---|---|
| 3-arg `setInternalState(child, "name", v)` | `child.getClass()` (runtime class only) |
| 4-arg `setInternalState(child, "name", v, Parent.class)` | `Parent.class` (declaring superclass) |

Everything else — `setAccessible(true)`, `field.set(target, value)`, the `throws Exception`,
the `java.lang.reflect.Field` import — is identical to the shipped 3-arg path. No cast, no boxing.

## Before → After

```java
// before
Whitebox.setInternalState(child, "name", "newValue", Parent.class);

// after
Field nameField = Parent.class.getDeclaredField("name");
nameField.setAccessible(true);
nameField.set(child, "newValue");
```

## Notes

- New `MethodMatcher` for `setInternalState(Object, String, Object, Class)`; the `String` 2nd
  param disambiguates it from the `setInternalState(Object, Class, Object, Class)` sibling overload.
- The 4th `Class` argument flows straight into a `#{any(java.lang.Class)}` placeholder, so a
  non-literal `Class<?>` variable as the 4th arg works too (generates `where.getDeclaredField(...)`).
- No catalog/YAML change — the recipe is already wired in and `displayName`/`description` are unchanged.

## Tests

Two new tests in `PowerMockWhiteboxToJavaReflectionTest`:
- `setInternalStateWithWhereClass` — inherited field via a literal `Parent.class`.
- `setInternalStateWithWhereClassVariable` — a `Class<?>` variable as the 4th arg.

`./gradlew check` passes (license, recipe CSV completeness/content validation, full test suite).

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/mockito/PowerMockWhiteboxToJavaReflection.java
+++ b/src/main/java/org/openrewrite/java/testing/mockito/PowerMockWhiteboxToJavaReflection.java
@@ -40,6 +40,8 @@ public class PowerMockWhiteboxToJavaReflection extends Recipe {
     private static final String WHITEBOX_FQN = "org.powermock.reflect.Whitebox";
     private static final MethodMatcher SET_INTERNAL_STATE =
             new MethodMatcher("org.powermock.reflect.Whitebox setInternalState(java.lang.Object, java.lang.String, java.lang.Object)");
+    private static final MethodMatcher SET_INTERNAL_STATE_WHERE =
+            new MethodMatcher("org.powermock.reflect.Whitebox setInternalState(java.lang.Object, java.lang.String, java.lang.Object, java.lang.Class)");
     private static final MethodMatcher GET_INTERNAL_STATE =
             new MethodMatcher("org.powermock.reflect.Whitebox getInternalState(java.lang.Object, java.lang.String)");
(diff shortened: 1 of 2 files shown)


MBoegers:
https://github.com/openrewrite/rewrite-testing-frameworks/pull/1021 is an ongoing refactoring with @steve-aom-elliott to break the recipe for expansion. This is the most pressing missing API.

timtebeek (approved):

PR #1026, opened 2026-06-17
Add `PowerMockWhiteboxGetMethodToJavaReflection` recipe

MBoegers:
## What
Adds a per-API recipe migrating `Whitebox.getMethod(Class, String, Class...)` to `Class.getDeclaredMethod(String, Class...)` + `setAccessible(true)`, registered in the `PowerMockWhiteboxToJavaReflection` composite. The result *is* the reflective `Method`, so the declared result variable is reused as the local:

```java
Method m = Whitebox.getMethod(MyService.class, "greet", String.class);
// →
Method m = MyService.class.getDeclaredMethod("greet", String.class);
m.setAccessible(true);
```

## How
A `hasArrayArg` helper on `WhiteboxToReflectionVisitor` skips calls that pass an explicit `Class[]` varargs array (left for manual migration); the parameter `Class` literals otherwise pass straight through.

The `resultLocalName` / `fieldVarName` / `methodVarName` helper group was shared with #1025 (`getField`). Now that #1025 has merged, this PR drops the duplicated helper block and retains only the `hasArrayArg` helper unique to this recipe.

## Notes
- `getDeclaredMethod` does not walk the class hierarchy — documented in the description.
- Second of three new-API recipes (`getField` → **`getMethod`** → `invokeConstructor`).

## Tests
4 cases: param type (`@DocumentExample`), no params, primitive param (`int.class`), and an explicit-`Class[]` no-op guard. `recipeCsvValidate` passes.


MBoegers:
@copilot resolve the merge conflicts

timtebeek (approved):

PR #1044, opened 2026-06-20
Verify junit5.StaticImports respects existing shadowing methods

### LINKED ISSUE #459: Edge case: junit5.StaticImports should respect existing `assertEquals`
```java
    public static void assertEquals(List<Text> expected, List<Text> actual) {
        // Need to compare string values since Texts with the same string are not considered equal
        Assertions.assertEquals(expected.toString(), actual.toString());
        assertEquals(expected.toString(), actual.toString());
```

`Assertions.assertEquals` is rewritten to `assertEquals` (with a static import), which does not compile.

    java: incompatible types: java.lang.String cannot be converted to java.util.List<javafx.scene.text.Text>

OK, this is an edge case. I will rename the method. Just want to bring it up if you are interested.

timtebeek:
- Closes #459

When a class declares a method whose name collides with a JUnit Jupiter assertion (e.g. a helper `assertEquals(List, List)`), a fully-qualified `Assertions.assertEquals(..)` call inside that class must keep the explicit `Assertions.` qualifier. Converting it to a static import would let the local method shadow the imported one, producing a compile error like:

```
java: incompatible types: java.lang.String cannot be converted to java.util.List<...>
```

The underlying `org.openrewrite.java.UseStaticImport` recipe (which `junit5.StaticImports` delegates to) already guards against this case. This PR adds a regression test that pins the behavior so it doesn't regress.

I confirmed the test is meaningful: renaming the local method so it no longer shadows the assertion makes the recipe convert the call as expected.

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/StaticImportsTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/StaticImportsTest.java
@@ -18,6 +18,7 @@ package org.openrewrite.java.testing.junit5;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.Issue;
 import org.openrewrite.config.Environment;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.test.RecipeSpec;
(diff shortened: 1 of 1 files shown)


timtebeek:
Closing alongside #459. The shadowing edge case is already handled upstream by `org.openrewrite.java.UseStaticImport`; the regression test on this branch confirmed the behavior but isn't being merged. Reopen if we'd like to land the test as a guard.

PR #1046, opened 2026-06-20
Convert `assertThrows` with an `Executable` variable

### LINKED ISSUE #511: `JUnitAssertThrowsToAssertExceptionType` does not convert cases where `executable` is a variable
- https://github.com/openrewrite/rewrite-testing-frameworks/releases/tag/v2.6.0

```java
@Test
void mapKorrekturgrundReasonThrows() {
   // when
   Executable executable = () -> IccTypeMapper.INSTANCE.mapKorrekturgrundReason(null);

   // then
   assertThrows(IllegalArgumentException.class, executable);
}
```

```java
@Test
void mapKorrekturgrundReasonThrows() {
   // when
   Executable executable = () -> IccTypeMapper.INSTANCE.mapKorrekturgrundReason(null);

   // then
   assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(executable);
}
```

## What did you see instead?
No change, as the recipe fails to match the case where the argument is a variable.
https://github.com/openrewrite/rewrite-testing-frameworks/blob/7ba17bbbfb2c9f5437a377d4e022c101a31579e2/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java#L68-L74

## Additional context
As reported on
- https://stackoverflow.com/q/78366075/53444

timtebeek:
- Fixes #511

### Problem
`JUnitAssertThrowsToAssertExceptionType` was already rewriting `assertThrows(..., executable)` into `assertThatExceptionOfType(...).isThrownBy(executable)` even when `executable` is a variable, but the result did not compile: the variable stayed typed as JUnit's `org.junit.jupiter.api.function.Executable`, while AssertJ's `isThrownBy` requires `ThrowableAssert.ThrowingCallable`. These are distinct functional interfaces — a lambda/method reference coerces to either, a typed variable does not.

### Change
When the executable argument is a variable typed exactly as `Executable`, the recipe now also retypes its declaration (and references) to `ThrowingCallable`, adding/removing imports accordingly.

This is guarded: the conversion only proceeds when *every* usage of that variable is an `assertThrows` executable argument (or the variable's own declaration/assignment). If the same variable is also passed somewhere that requires an `Executable` (e.g. `assertAll`), the types can't both be satisfied, so the call is left untouched rather than emitting non-compiling code.

### Tests
- `variableExecutable` — local variable (the issue's core case)
- `variableExecutableWithMessage` — 3-arg form with `.as(...)`
- `fieldExecutable` — field declaration
- `doNotRetypeExecutableSharedWithOtherJUnitAssertion` — guard: variable shared with `assertAll` is left unchanged

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
+++ b/src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
@@ -22,7 +22,6 @@ import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.Tree;
 import org.openrewrite.TreeVisitor;
-import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
@@ -167,10 +166,9 @@ public class JUnitAssertThrowsToAssertExceptionType extends Recipe {
                 if (cu == null) {
                     return false;
                 }
-                AtomicBoolean safe = new AtomicBoolean(true);
-                new JavaIsoVisitor<Integer>() {
+                return new JavaIsoVisitor<AtomicBoolean>() {
                     @Override
-                    public J.Identifier visitIdentifier(J.Identifier identifier, Integer integer) {
+                    public J.Identifier visitIdentifier(J.Identifier identifier, AtomicBoolean safe) {
                         if (safe.get() && Objects.equals(identifier.getFieldType(), variable)) {
                             Object parent = getCursor().getParentTreeCursor().getValue();
                             boolean ok = false;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/assertj/JUnitAssertThrowsToAssertExceptionType.java
```diff
@@ -120,6 +154,85 @@ private Optional<Boolean> hasReturnType() {
...
+                                        enclosing.getArguments().get(1) == identifier;
+                            }
+                            if (!ok) {
+                                safe.set(false);
+                            }
+                        }
+                        return super.visitIdentifier(identifier, integer);
+                    }
+                }.visit(cu, 0);
+                return safe.get();
```
timtebeek: Use `.reduce` here to inline the `safe` variable.

PR #1052, opened 2026-07-14
KotlinTestMethodsShouldReturnUnit: exclude @TestFactory methods

neil-mushell:
## What's changed?                                                                                                                                     
`KotlinTestMethodsShouldReturnUnit`'s `TEST_ANNOTATION_PATTERN` (`"org..* *Test*"`) uses a substring wildcard (`*Test*`) that unintentionally also matches `@TestFactory`, even though the recipe's own description only targets `@Test`, `@ParameterizedTest`, `@RepeatedTest`, and `@TestTemplate`. This adds an explicit exclusion for `@TestFactory` plus a regression test.                                                                                  
                                                                                                                                                         
## What's your motivation?                                                                                                                             
Unlike those four annotations, `@TestFactory` methods are required by JUnit Jupiter to return a value (e.g. `Collection<DynamicTest>`), not `Unit`. Applying this recipe to a `@TestFactory` method currently rewrites its declared return type to `Unit` while the body still evaluates to something like `List<DynamicTest>`, producing a Kotlin type-mismatch compile error. We hit this adopting the recipe fleet-wide and found real `@TestFactory` usages it would have broken.                                                                                                                                     
                                                                                                                                                         
Whether an explicit exclusion check is preferable to tightening `TEST_ANNOTATION_PATTERN` itself — the pattern DSL doesn't appear to support alternation, so a separate exclusion seemed simpler.                                                           

@timtebeek
@greg-dennis

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/cleanup/KotlinTestMethodsShouldReturnUnit.java
+++ b/src/main/java/org/openrewrite/java/testing/cleanup/KotlinTestMethodsShouldReturnUnit.java
@@ -35,6 +35,11 @@ import static java.util.Collections.emptyList;
 public class KotlinTestMethodsShouldReturnUnit extends Recipe {
 
     private static final String TEST_ANNOTATION_PATTERN = "org..* *Test*";
+    // `TEST_ANNOTATION_PATTERN`'s `*Test*` segment matches any simple name containing "Test", which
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Perfect, thanks for spotting and reporting!

PR #1054, opened 2026-07-17
Fix Hamcrest not() unwrap infinite loop and support CoreMatchers

Jammy-Louie:
## What's changed

Two commits:

1. **Fix an infinite loop** in `HamcrestMatcherToJUnit5` / `HamcrestInstanceOfToJUnit5` on a `not(...)` matcher that isn't `org.hamcrest.Matchers.not`.
2. **Migrate `CoreMatchers` variants** of `not`/`instanceOf`/`isA` instead of skipping them.

## The hang

The unwrap loop continues on a name check but only makes progress on a type check, so they disagree:

```java
while ("not".equals(matcherInvocation.getSimpleName())) {          // any not
    matcherInvocation = new RemoveNotMatcherVisitor().visit(...);  // only strips org.hamcrest.Matchers.not
}
```

A `CoreMatchers.not` (or a `not` with unresolved type) never gets stripped, the name stays `"not"`, and the loop spins forever. Repro:

```java
import static org.hamcrest.CoreMatchers.not;
import static org.hamcrest.CoreMatchers.equalTo;
assertThat(a, not(equalTo(b))); // hangs
```

**Fix:** loop on the matcher that decides strippability, so continuation and progress can't disagree:

```java
while (RemoveNotMatcherVisitor.NOT_MATCHER.matches(matcherInvocation)) {
```

No change to existing migrations.

## CoreMatchers support

Broadens the three matchers from `org.hamcrest.Matchers` to `org.hamcrest.*Matchers` (matches exactly `Matchers` and `CoreMatchers`, identical methods), so `CoreMatchers.not`/`instanceOf`/`isA` migrate too.

## Tests

`notFromCoreMatchers` (both recipes, asserting the real migration, `@Timeout`-guarded so a regressed hang fails fast) and `instanceOfAndIsAFromCoreMatchers`. Full suite passes.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestInstanceOfToJUnit5.java
+++ b/src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestInstanceOfToJUnit5.java
@@ -70,7 +70,7 @@ public class HamcrestInstanceOfToJUnit5 extends Recipe {
                     }
 
                     J.MethodInvocation matcherInvocation = (J.MethodInvocation) hamcrestMatcher;
-                    while ("not".equals(matcherInvocation.getSimpleName())) {
+                    while (RemoveNotMatcherVisitor.NOT_MATCHER.matches(matcherInvocation)) {
                         maybeRemoveImport("org.hamcrest.Matchers.not");
                         maybeRemoveImport("org.hamcrest.CoreMatchers.not");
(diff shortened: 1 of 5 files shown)


timtebeek (approved):
Perfect, thanks a lot! This explains why I was seeing timeouts with AssertJ indeed.

PR #1058, opened 2026-07-23
HamcrestMatcherToJUnit5: support Kotlin sources

timtebeek:
`HamcrestMatcherToJUnit5` threw "No JavaSourceFile parent found" on Kotlin test sources because it always used `JavaTemplate`; it now branches on `K.CompilationUnit` and applies a `KotlinTemplate` (via `KotlinParser`) for Kotlin sources. It uses a Kotlin-specific template variant where the Java form is invalid Kotlin (`Collection.size()` → `.size`) and fixes a latent unbalanced-paren typo in the `containsString` template that only the Kotlin parser rejected. Type attribution of the generated assertions relies on the upstream fix openrewrite/rewrite#8324 (KotlinTemplate attributes members selected on substituted placeholders), so it needs a rewrite snapshot ≥ 8.88.0 build 59. The only remaining manual attribution is the enclosing `assertX(condition, reason)` call, which Kotlin leaves untyped because JUnit's `(…, String)` and `(…, Supplier)` overloads are ambiguous when the reason is a substituted placeholder. Adds Kotlin tests covering equalTo, not, greaterThan, containsString, nullValue, collections, and reason arguments.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToJUnit5.java
+++ b/src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToJUnit5.java
@@ -16,6 +16,7 @@
 package org.openrewrite.java.testing.hamcrest;
 
 import lombok.Getter;
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestMatcherToJUnit5.java
```diff
@@ -174,10 +198,97 @@ public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, Execu
...
+            }
+        }
+        JavaType.Method getter = findMethod(type, name, 0);
+        return getter == null ? null : getter.getReturnType();
+    }
+
+    private static int arity(J.MethodInvocation mi) {
+        List<Expression> arguments = mi.getArguments();
+        return arguments.size() == 1 && arguments.get(0) instanceof J.Empty ? 0 : arguments.size();
+    }
```
timtebeek: We're going to first see if we can improve the type attribution upstream for Kotlin snippets, as we wouldn't want to have to repeat this logic anywhere we do variable substitution in templates.

timtebeek: - https://github.com/openrewrite/rewrite/pull/8324

PR #1062, opened 2026-07-23
Fix Hamcrest not() unwrap crash when wrapping a bare value

Jammy-Louie:
## What's changed

`HamcrestMatcherToJUnit5` and `HamcrestInstanceOfToJUnit5` invoke `RemoveNotMatcherVisitor` via `visit(matcherInvocation, ctx)`, which roots the sub-visitor at `Cursor.ROOT`. The `matcherInvocation` passed in is a detached subtree, so its cursor path has no enclosing `JavaSourceFile`.

When a `not(...)` matcher wraps a bare value rather than another Hamcrest matcher, `RemoveNotMatcherVisitor` wraps the value in `equalTo(...)` and calls `maybeAddImport`. That triggers an `ImportService` lookup which walks the cursor for a `JavaSourceFile` ancestor and throws `IllegalArgumentException: No JavaSourceFile parent found`.

The fix passes the current cursor as the parent (`visit(matcherInvocation, ctx, getCursor())`) so the sub-visitor inherits a path that includes the `JavaSourceFile` ancestor. Applied to both call sites.

## Example

Before this fix, the recipe crashed on:

```java
assertThat(str1, not(str2)); // str2 is a bare value, not a Hamcrest matcher
```

Now it correctly migrates to:

```java
assertNotEquals(str1, str2);
```

## Testing

Added `notWithBareValue` to `HamcrestMatcherToJUnit5Test`, which reproduces the crash without the fix and passes with it. All existing Hamcrest recipe tests continue to pass.

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestInstanceOfToJUnit5.java
+++ b/src/main/java/org/openrewrite/java/testing/hamcrest/HamcrestInstanceOfToJUnit5.java
@@ -73,7 +73,7 @@ public class HamcrestInstanceOfToJUnit5 extends Recipe {
                     while (RemoveNotMatcherVisitor.NOT_MATCHER.matches(matcherInvocation)) {
                         maybeRemoveImport("org.hamcrest.Matchers.not");
                         maybeRemoveImport("org.hamcrest.CoreMatchers.not");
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Yes perfect, nice find!

PR #1063, opened 2026-07-23
Migrate TestNG SoftAssert to AssertJ SoftAssertions

juherr:
## What's changed

Adds a `TestNgSoftAssertToAssertJ` recipe, wired into the existing `TestNgToAssertj` declarative recipe, to migrate TestNG soft assertions (`org.testng.asserts.SoftAssert`) to AssertJ soft assertions (`org.assertj.core.api.SoftAssertions`).

```java
// before
SoftAssert softAssert = new SoftAssert();
softAssert.assertEquals(actual, expected);   // TestNG: actual first, expected second
softAssert.assertTrue(cond, "msg");
softAssert.assertAll();

// after
SoftAssertions softAssert = new SoftAssertions();
softAssert.assertThat(actual).isEqualTo(expected);
softAssert.assertThat(cond).as("msg").isTrue();
softAssert.assertAll();
```

## How it works

The instance assertion calls are rewritten first, keeping the receiver, then a `ChangeType` step handles the variable type, the constructor, imports, and the identically named `assertAll()`/`fail(...)` methods.

The assertion methods (`assertEquals`, `assertTrue`, ...) are declared on `org.testng.asserts.Assertion` — `SoftAssert` only adds `assertAll` — so the matcher targets `Assertion` and a receiver-type guard keeps direct `Assertion` (hard) usage untouched.

## Covered methods

- `assertEquals` / `assertNotEquals` — preserving TestNG's `(actual, expected)` order, plus the message and floating-point delta overloads (`isCloseTo(..., within(...))` / `isNotCloseTo(...)`)
- `assertTrue` / `assertFalse`
- `assertNull` / `assertNotNull`
- `assertSame` / `assertNotSame`
- `assertEqualsNoOrder` → `containsExactlyInAnyOrder`
- `assertAll()` and `fail(...)` keep the same name and are handled by `ChangeType`

## Notes

- Broadens the `assertj-core` `onlyIfUsing` dependency trigger from `org.testng.*` to `org.testng..*`, so classes using only `org.testng.asserts.SoftAssert` still get the dependency added.
- The lambda form `SoftAssertions.assertSoftly(s -> { ... })` is intentionally not emitted; the explicit instance + `assertAll()` shape is a 1:1 mapping.

Tests added to `TestNgToAssertJTest` cover the full flow, every overload branch, non-variable receivers (inline `new SoftAssert()` and fields), the `assertEqualsNoOrder` mapping, `fail(...)` passthrough, boxed-`Double` deltas, and a negative test asserting direct `Assertion` usage is left unchanged.


timtebeek (approved):
Perfect, thanks a lot! Good to see this soft assert gap closed as well. We'll likely do a .patch release later today, or the regular release mid next week.

PR #1064, opened 2026-07-24
Migrate TestNG Assertion (instance) to AssertJ

juherr:
## What's changed

Adds `TestNgAssertionToAssertJ`, wired into the `TestNgToAssertj` aggregate, to migrate TestNG's hard, instance-based assertion class `org.testng.asserts.Assertion` to static AssertJ `assertThat(...)`. Because `Assertion` fails immediately (unlike `SoftAssert`), it maps to the static `assertThat` entry point — the same target as `org.testng.Assert`.

```java
// before
Assertion assertion = new Assertion();
assertion.assertEquals("actual", "expected");   // TestNG order: actual, then expected

// after
assertThat("actual").isEqualTo("expected");      // the now-dead local instance is removed
```

### The recipe
- Rewrites instance assertion calls to static `assertThat(...)`, scoped by **exact** receiver type (`TypeUtils.isOfClassType(..., "org.testng.asserts.Assertion")`), so custom `Assertion` subclasses are left untouched and it stays disjoint from `SoftAssert extends Assertion`.
- A `visitBlock` pass removes the now-unused **local** `Assertion` instance once its calls are rewritten (single-cycle; only when the variable has no remaining references). Fields are left in place.
- Covered: `assertEquals`/`assertNotEquals` (TestNG `(actual, expected)` order, plus message → `.as(...)` and floating-point delta → `isCloseTo(..., within(...))`/`isNotCloseTo(...)`), `assertTrue`/`assertFalse`, `assertNull`/`assertNotNull`, `assertSame`/`assertNotSame`, `assertEqualsNoOrder` → `containsExactlyInAnyOrder`, and `fail(...)` → static `Assertions.fail(...)` (no-arg `fail()` → `fail("")`, matching `JUnitFailToAssertJFail`).

### Shared helper
Extracts the duplicated `isFloatingPointType` into a package-private `TestNgAsserts` and points every caller at it: `TestNgAssertEqualsToAssertThat`, `TestNgAssertNotEqualsToAssertThat`, `TestNgSoftAssertToAssertJ`, and the new recipe.

### Test change worth noting
`TestNgToAssertJTest#hardAssertionIsNotChanged` is removed: it asserted that the aggregate leaves a hard `org.testng.asserts.Assertion` instance unchanged, which is no longer true now that this recipe migrates it. Its replacement, `assertionInstanceMigratedViaAggregate`, pins the new behaviour, and `customAssertionSubclassIsNotChanged` keeps the subclass-is-left-alone guarantee.

Tests cover every overload branch, the dead-local removal (plus retention-when-referenced and field-not-removed), and the `fail` variants.


timtebeek (approved):
Thanks again! With a slight bit of polish I'll merge as soon as CI passes.

PR #1069, opened 2026-07-30
Use assertInstanceOf for expectCause(isA/instanceOf/is(X.class))

neil-mushell:
## What's changed?                                                                                                                                                                 
                                                                                                                                                                                     
  `ExpectedExceptionToAssertThrows` now special-cases `expectCause(isA(X.class))`, `expectCause(instanceOf(X.class))`, and `expectCause(is(X.class))`, emitting the type-safe        
  `assertInstanceOf(X.class, exception.getCause())` instead of the type-unsafe `assertThat(exception.getCause(), isA(X.class))`.                                                     
                                                                                                                                                                                     
  This mirrors the existing `HamcrestInstanceOfToJUnit5` recipe's treatment of `instanceOf`/`isA` matchers, applied to this recipe's `expectCause(...)` handling specifically.       
                                                                                                                                                                                     
  **Key changes:**                                                                                                                                                                   
  - Added `IS_A_MATCHER`, `INSTANCE_OF_MATCHER`, `IS_MATCHER_CORE_MATCHERS`, `IS_MATCHER_CORE_IS` method matchers (using the `org.hamcrest.*Matchers` wildcard, since                
  `isA`/`instanceOf`/`is` are reachable from both `CoreMatchers` and the consolidated `Matchers`)                                                                                    
  - Added `maybeAssertInstanceOfForCause()` to build the `assertInstanceOf(...)` replacement when the matcher argument is a class literal                                            
  - Added `extractClassLiteralFromInstanceOfMatcher()` — `is(X.class)` is only treated as an instanceof check when its argument is a class literal, to avoid false positives on      
  `is("string")` or `is(someOtherMatcher)`                                                                                                                                           
                                                                                                                                                                                     
  ## What's your motivation?                                                                                                                                                         
                                                                                                                                                                                     
  `assertThat(exception.getCause(), isA(X.class))` doesn't give compile-time type safety and reads less directly than `assertInstanceOf(X.class, exception.getCause())`, which is    
  also what JUnit 5 recommends for this exact pattern.                                                                                                                               
                                                                                                                                                                                     
  The `is(X.class)` special-casing — `is()` is overloaded between an equality matcher and a `Matcher<T>` decorator, so I only treat it as an instanceof check when the argument is   
  literally a class literal (`J.FieldAccess` ending in `.class`), not a variable or nested matcher.                                                                                  
                                                                                                                                                                                     
  ### Checklist                                                                                                                                                                      
  - [x] I've read and applied the recipe conventions and best practices                                                                                                              

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
@@ -74,6 +74,11 @@ public class ExpectedExceptionToAssertThrows extends Recipe {
         private static final MethodMatcher EXPECTED_MESSAGE_MATCHER = new MethodMatcher("org.junit.rules.ExpectedException expectMessage(org.hamcrest.Matcher)");
         private static final MethodMatcher EXPECTED_EXCEPTION_MATCHER = new MethodMatcher("org.junit.rules.ExpectedException expect(org.hamcrest.Matcher)");
         private static final MethodMatcher EXPECTED_EXCEPTION_CAUSE_MATCHER = new MethodMatcher("org.junit.rules.ExpectedException expectCause(org.hamcrest.Matcher)");
+        // *Matchers wildcard: isA/instanceOf/is are all reachable from both CoreMatchers and the consolidated Matchers.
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks a lot! Quality work here.. Good to see the coverage here still expanding. Hope it helps you migrate!

PR #1074, opened 2026-08-05
ExpectedExceptionToAssertThrows: inline single-use reassigned locals 

neil-mushell:
## What's changed?                                                                                                                                                                                                                    
  <!-- A brief description of the changes in this pull request -->                                                                                                                                                                      
                                                                                                                                                                                                                                        
  A local reassigned via `=` before ExpectedException.expect*() and then read by the statement(s) that get spliced into the assertThrows(...) lambda fails to compile: Java lambdas require captured locals to be effectively final, and
  a reassignment (even one before the capture point) disqualifies it.                                                                                                                                                                   
                                                                                                                                                                                                                                        
  For the narrow, common shape -- reassigned exactly once, RHS doesn't reference itself, not read again before expect*(), and read exactly once in the moved statements -- delete the reassignment and inline its RHS at that single use
  site instead. Chains and multi-use reads are left untouched for now.                                                                                                                                                                  
                                                                                                                                                                                                                                        
  `ExpectedExceptionToAssertThrows` now inlines a reassigned local at its splice site instead of letting the recipe emit a lambda that captures a non-effectively-final variable. Scope is deliberately narrow: only fires when the     
  local is reassigned exactly once before `expect*()`, the RHS of that reassignment doesn't reference the local itself, the local isn't read again before `expect*()`, and it's read exactly once among the statements moving into the  
  `assertThrows(...)` lambda. Chained reassignments and multi-use reads after `expect*()` are left exactly as they behave today -- not fixed here, but not made worse either.                                                           
                                                                                                                                                                                                                                        
  ## What's your motivation?                                                                                                                                                                                                            
  <!-- This can link to close a separate issue, or be described on the pull request itself -->                                                                                                                                          
                                                                                                                                                                                                                                        
  Ran into this while migrating a large JUnit 4 codebase off `ExpectedException`: the recipe silently produced non-compiling output whenever a captured local had been reassigned earlier in the method. This fixes the common          
  single-reassignment/single-use shape without touching the harder chained/multi-use shapes, which need a different fix (capturing into a fresh `final` local rather than inlining) and are left for a follow-up.                       
                                                                                                                                                                                                                                        
  <!-- You can also start a discussion on particular aspects of your implementation on the files tab yourself. -->                                                                                                                      
                                                                                                                                                                                                                                        
  The `updateCursor(simplified)` call before delegating to `super.visitBlock` -- without it, the existing splice logic reads a stale statement list via the cursor-message mechanism (`findSuccessorStatements`), since                 
  `super.visitBlock` doesn't automatically rebind the visitor's cursor to a block passed in as a plain parameter.                                                                                                                       
                                                                                                                                                                                                                                        
  <!-- @mention them here -->                                                                                                                                                                                                           
                                                                                                                                                                                                                                        
  <!-- Any other ways to solve the problem, or ways to work around the problem. -->                                                                                                                                                     
                                                                                                                                                                                                                                        
  Considered generalizing this to always capture the value into a fresh `final` local (handles chains and multi-use too), but that's a larger, separate change -- left for a follow-up PR to keep this one focused and easy to review.  
                                                                                                                                                                                                                                        
  ## Any additional context                                                                                                                                                                                                             
  <!-- Any thoughts you would like to share in addition to the above. -->                                                                                                                                                               
                                                                              
  ### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/ExpectedExceptionToAssertThrows.java
@@ -88,25 +88,26 @@ public class ExpectedExceptionToAssertThrows extends Recipe {
         @Override
         public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
             J.ClassDeclaration cd = super.visitClassDeclaration(classDecl, ctx);
-            doAfterVisit(new LambdaBlockToExpression().getVisitor());
 
             // A method that could not be migrated still needs the rule, so leave the field and its imports in place.
-            if (getCursor().getMessage(CANNOT_MIGRATE) != null) {
-                return cd;
+            if (getCursor().getMessage(CANNOT_MIGRATE) == null) {
+                cd = cd.withBody(cd.getBody().withStatements(ListUtils.map(cd.getBody().getStatements(), statement -> {
+                    if (statement instanceof J.VariableDeclarations) {
+                        //noinspection ConstantConditions
+                        if (TypeUtils.isOfClassType(((J.VariableDeclarations) statement).getTypeExpression().getType(),
+                                "org.junit.rules.ExpectedException")) {
+                            maybeRemoveImport("org.junit.Rule");
+                            maybeRemoveImport("org.junit.rules.ExpectedException");
+                            return null;
+                        }
+                    }
+                    return statement;
+                })));
             }
 
-            cd = cd.withBody(cd.getBody().withStatements(ListUtils.map(cd.getBody().getStatements(), statement -> {
-                if (statement instanceof J.VariableDeclarations) {
-                    //noinspection ConstantConditions
-                    if (TypeUtils.isOfClassType(((J.VariableDeclarations) statement).getTypeExpression().getType(),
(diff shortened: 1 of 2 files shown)


neil-mushell:
@timtebeek would be grateful when you have a chance to take a look.

timtebeek:
Thanks for the ping; having a look now.

timtebeek (approved):
Thanks a lot for the help here @neil-mushell ! I've pushed some changes for further edge cases found, and made the recipe not change a method or remove the class field if not safe to do so. That would have otherwise still lead to compilation errors of the type you're looking to reduce here, and is in line with our do-no-harm policy. It does mean you'll see no partial migrations, and might end up with fields retained without a clear explanation why, so I figured let you know before we merge. Let me know if you're ok to see this go out in this form!

neil-mushell:
> Thanks a lot for the help here @neil-mushell ! I've pushed some changes for further edge cases found, and made the recipe not change a method or remove the class field if not safe to do so. That would have otherwise still lead to compilation errors of the type you're looking to reduce here, and is in line with our do-no-harm policy. It does mean you'll see no partial migrations, and might end up with fields retained without a clear explanation why, so I figured let you know before we merge. Let me know if you're ok to see this go out in this form!

Thanks for enhancing this. LGTM.

PR #1111, opened 2026-08-27
Add pmd recipe

sullis:
## What's changed?

PMD 7 removed a batch of Java rules, and PMD fails on a `<rule ref>` it cannot resolve. So a PMD 6 ruleset that references any of them stops the build outright after upgrading to PMD 7, one unresolvable `ref` at a time.

This adds `PmdXml6to7Migration`, an XML recipe that updates a PMD `ruleset` file for PMD 7:

- repoints the 46 removed rules that have a single unambiguous PMD 7 successor at that successor — note the successor often lives in a different category than the rule it replaces, e.g. `errorprone.xml/DataflowAnomalyAnalysis` → `bestpractices.xml/UnusedAssignment`,
- drops the four rules removed without any successor (`AvoidFinalLocalVariable`, `AvoidUsingShortType`, `CloneThrowsCloneNotSupportedException`, `SimplifyStartsWith`), since leaving them behind is what breaks PMD 7,
- does not add a successor twice when several removed rules collapse onto one (all four `UseAssert*InsteadOf*` rules become `SimplifiableTestAssertion`) or when the ruleset already selects it,
- keeps nested `<properties>` on a rule it repoints.

```xml
<rule ref="category/java/errorprone.xml/DontImportSun" />
<rule ref="category/java/errorprone.xml/EmptyCatchBlock" />
```
becomes
```xml
<rule ref="category/java/errorprone.xml/EmptyCatchBlock" />
<rule ref="category/java/errorprone.xml/UnsupportedJdkApiUsage" />
```

A replacement is inserted in `ref` order relative to the rules already there, taking the indentation of its neighbor. Rules that aren't being replaced are never reordered, so a ruleset that happens to be unsorted stays as its author wrote it.

Two things are deliberately left alone:

- rules PMD 7 splits across several successors (`{Boolean,Byte,Integer,Long,Short}Instantiation`, `MIsLeadingVariableName`, `VariableNamingConventions`) — picking one successor would silently change what the ruleset covers, so this leaves the choice to a human,
- any XML document whose root isn't `ruleset`, rather than matching a `ruleset` tag that happens to appear elsewhere.

Also adds `rewrite-xml` to the build and registers the recipe in `recipes.csv`.

## PMD project

https://docs.pmd-code.org/latest/index.html

https://github.com/pmd/pmd

Removed rules are listed in the [PMD 7.0.0 release notes](https://docs.pmd-code.org/pmd-doc-7.0.0/pmd_release_notes_pmd7.html).


sullis:
Alternate implementation:  https://github.com/openrewrite/rewrite-static-analysis/pull/1034

PR #1112, opened 2026-08-27
fix: comments must not break rewritten assertions

cortlepp:
## What's changed?

The `AssertToAssertions` recipe currently does not handle comments correctly when reordering arguments, which can lead to compilation errors after applying the recipe.

@timtebeek 
### Checklist
- [x] I've run `./gradlew build` locally, and committed any resulting changes to `recipes.csv`
- [x] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch

### DIFF
--- a/src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/AssertToAssertionsTest.java
@@ -164,6 +164,40 @@ class AssertToAssertionsTest implements RewriteTest {
     }
 
     @Test
+    void commentWithMessage() {
+        //language=java
+        rewriteRun(
+          java(
+            """
+              import org.junit.Test;
+              import static org.junit.Assert.assertTrue;
+
+              public class MyTest {
+                  @Test
+                  public void test() {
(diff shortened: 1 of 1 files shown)


cortlepp:
@timtebeek before trying to come up with a fix I wanted to check with you what a good solution would look like (the one I gave in the test is not necessarily it). I think technically the issue is that the whitespace rewriting seems not to take linebreaks into account, which can break single line comments (like in the test example). We could of course try to fix that, or we could may convert the single line comments to multi-line ones `/* */` which would allow us to keep the overall "shape" of the expression intact (which the current implementation does, but without  proper comment handling). IMO option two might be easier to get right and it would preserve the original intent of why the expression was written in a that specific way. But I don't have a strong opinion on this, I think the main issue is that comments can currently cause compilation errors and that we should fix that in some way.

timtebeek:
hi! I'm about to take some time off, so I'll respond briefly, but recommend colleagues for further follow up.

I'd lean towards keeping the comments in place at the end, and turning `//` into `/* */` could run into trouble with `// */ still a comment`.

PR #1117, opened 2026-09-04
Add junit-jupiter when only junit.framework.TestCase is used

### LINKED ISSUE #1113: JUnit4to5Migration removes junit:junit without adding Jupiter when tests extend junit.framework.TestCase
Hi! Thanks for all the work on these recipes — migrating a 20-year-old codebase with them was genuinely impressive, and this was the one place I got stuck.

`org.openrewrite.java.testing.junit5.JUnit4to5Migration` removes `junit:junit` but never adds JUnit Jupiter when a project's tests extend `junit.framework.TestCase`, leaving a project that no longer compiles. The recipe reports `BUILD SUCCESS`.

This is squarely within the recipe's own scope: `JUnit4to5Migration` includes
`org.openrewrite.java.testing.junit5.MigrateJUnitTestCase`, whose own display name is
*"Migrate JUnit 4 `TestCase` to JUnit Jupiter"* and which operates on
`junit.framework.TestCase` / `junit.framework.Assert`. Those classes ship inside the
`junit:junit` 4.x artifact, so this is an ordinary JUnit 4 codebase — no JUnit 3
support is being requested here.

There's also a nasty second-order effect: `rewrite:run` binds after
`process-test-classes`, so once the project stops compiling, OpenRewrite can no longer
be run to fix it.

## Reproduction — a JUnit 4 project

`pom.xml`:

```xml
<dependencies>
  <dependency>
    <groupId>junit</groupId>
    <artifactId>junit</artifactId>
    <version>4.13.2</version>
    <scope>test</scope>
  </dependency>
</dependencies>
```

`src/test/java/com/example/LegacyTest.java`:

```java
package com.example;

import junit.framework.TestCase;

public class LegacyTest extends TestCase {
    public void testAddition() {
        assertEquals(4, 2 + 2);
    }
}
```

Run:

```bash
mvn org.openrewrite.maven:rewrite-maven-plugin:6.46.1:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:3.44.0 \
  -Drewrite.activeRecipes=org.openrewrite.java.testing.junit5.JUnit4to5Migration
```

### Actual

```
[WARNING] Changes have been made to pom.xml by:
[WARNING]     org.openrewrite.java.dependencies.RemoveDependency: {groupId=junit, artifactId=junit}
[WARNING] Changes have been made to src/test/java/com/example/LegacyTest.java by:
[INFO] BUILD SUCCESS
```

`pom.xml` afterwards has **zero `<dependency>` elements**, while the test source now
imports `org.junit.jupiter.api`. `mvn test-compile`:

```
[ERROR] LegacyTest.java:[3,29] package org.junit.jupiter.api does not exist
[ERROR] LegacyTest.java:[5,1]  static import only from classes and interfaces
[ERROR] LegacyTest.java:[8,6]  cannot find symbol
[ERROR] LegacyTest.java:[10,9] cannot find symbol
```

A project declaring `junit:junit:3.8.2` behaves identically — the dependency version is
not the trigger.

### Expected

The composite should leave the project compilable: `junit-jupiter` added in test scope
when `junit:junit` is removed and no JUnit 5 dependency is present.

## Root cause

`AddJupiterDependencies` builds its internal `AddDependency` with
`onlyIfUsing = org.junit..*`:

```
$ javap -p -c org.openrewrite.java.testing.junit5.AddJupiterDependencies | grep 'ldc.*String'
...
11: ldc  #37   // String org.junit..*
 4: ldc  #31   // String org.junit.jupiter
 6: ldc  #33   // String junit-jupiter
 8: ldc  #35   // String 5.x
```

A test that only extends `junit.framework.TestCase` uses no `org.junit.*` type, so the
guard never fires. The sibling `RemoveDependency: junit:junit` in the same recipe list
has no such guard and always runs.

Because `AddDependency` is a `ScanningRecipe`, its scanner evaluates the pre-edit LST,
so the fact that `MigrateJUnitTestCase` / `UpdateTestAnnotation` later rewrite the
sources to `org.junit.jupiter.api.Test` doesn't help. Tests annotated with
`org.junit.Test` match the guard and are unaffected — which is why this only shows up
on the `junit.framework.TestCase` style.

## Suggested fix

Broaden the guard in `AddJupiterDependencies` so it also matches `junit.framework..*`
alongside `org.junit..*`. Happy to open a PR with that plus a test case next to the
existing ones if you think that's the right shape — didn't want to presume, since
making the `RemoveDependency` conditional instead is also defensible.

## Not a duplicate of

I looked for existing reports first:

- **#506** — a JUnit 4 project where Jupiter wasn't added; the log there shows
  `AddDependency` firing correctly with `onlyIfUsing=org.junit..*`, so the cause was
  different and specific to that project.
- **#457** — Mockito / `mockito-all` ordering, partially addressed in 216983b. Related
  in being another `onlyIfUsing` sensitivity, different path.
- **#1103**, **#1097**, **#1110** — same family of "JUnit4to5 leaves X behind",
  different X.
- **#1108** — the retained `TestCase(String)` constructor. Directly adjacent: on this
  repro, adding the dependency by hand then exposes #1108, so a
  `junit.framework.TestCase` codebase currently needs two manual interventions to get
  through what looks like a single-command migration.

## Environment

- `rewrite-maven-plugin` 6.46.1, `rewrite-testing-frameworks` 3.44.0 (both from Maven Central)
- JDK 21.0.12.1, Maven 3.9.16, surefire 3.5.4

Found while reviving a 2006 Swing/JDBC desktop application (a German Kindergarten
card-index app recovered from SourceForge CVS) and migrating it from Java 8 to Java 21.
Thanks again for the recipes — the rest of that migration was almost entirely automated.

manduinca:
## What's changed?
`AddJupiterDependencies` now also scans for uses of `junit.framework..*`, next to `org.junit..*`, when deciding whether to add `junit-jupiter`. The two scanners share the accumulator, so either one marks the project as using JUnit and the scope logic is unchanged.

## What's your motivation?
Fixes #1113. In a project whose tests only extend `junit.framework.TestCase`, `JUnit4to5Migration` removed `junit:junit` but never added Jupiter, because the scanner runs on the sources before `MigrateJUnitTestCase` rewrites them and no `org.junit` type is in use at that point. The migrated project didn't compile.

Whether broadening the guard here is the shape you want, versus making the `RemoveDependency` in the composite conditional. This one keeps `AddJupiterDependencies` correct on its own, which also covers running it standalone on a JUnit 3 codebase.

A single glob covering both packages doesn't exist (`org.junit` and `junit.framework` share no prefix), and `AddDependency` takes one `onlyIfUsing` pattern, hence the two scanners.

## Any additional context
Three tests: one on `AddJupiterDependencies` with a `TestCase` subclass in the test source set, a negative one with no JUnit at all (nothing is added), and one on the `JUnit4to5Migration` composite reproducing the issue (pom with `junit:junit` only), asserting `junit-jupiter` is added and `junit` removed. The two positive ones fail on `main` and pass with the change.

On building locally: since #1116 the build plugins resolve from the Code Genome repository, which asks for credentials I don't have, and `latest.release` (2.23.4) depends on rewrite 8.91.4, which isn't on Maven Central. With the plugins pinned to 2.23.1 locally (not part of this PR) `./gradlew build` passes: 1859 tests, 0 failures, `recipes.csv` validated.

### Checklist
- [x] I've run `./gradlew build` locally, and committed any resulting changes to `recipes.csv` (with the plugin pin described above; `recipes.csv` unchanged)
- [x] I've [formatted the lines I changed](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#code-style-and-formatting), without reformatting code I didn't touch

### DIFF
--- a/src/main/java/org/openrewrite/java/testing/junit5/AddJupiterDependencies.java
+++ b/src/main/java/org/openrewrite/java/testing/junit5/AddJupiterDependencies.java
@@ -47,7 +47,18 @@ public class AddJupiterDependencies extends ScanningRecipe<AddDependency.Accumul
 
     @Override
     public TreeVisitor<?, ExecutionContext> getScanner(AddDependency.Accumulator acc) {
-        return addJupiterDependency().getScanner(acc);
+        // The scanner sees the sources before `MigrateJUnitTestCase` rewrites them, so a
+        // codebase of `junit.framework.TestCase` subclasses uses no `org.junit` type yet.
+        TreeVisitor<?, ExecutionContext> usesJUnit4 = addJupiterDependency("org.junit..*").getScanner(acc);
+        TreeVisitor<?, ExecutionContext> usesJUnit3 = addJupiterDependency("junit.framework..*").getScanner(acc);
+        return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                usesJUnit4.visit(tree, ctx);
+                usesJUnit3.visit(tree, ctx);
+                return tree;
+            }
+        };
     }
 
     @Override
@@ -73,8 +84,12 @@ public class AddJupiterDependencies extends ScanningRecipe<AddDependency.Accumul
     }
 
     private static AddDependency addJupiterDependency() {
+        return addJupiterDependency("org.junit..*");
+    }
+
+    private static AddDependency addJupiterDependency(String onlyIfUsing) {
         return new AddDependency("org.junit.jupiter", "junit-jupiter", "5.x", null,
-                "org.junit..*", null, null, null, null, null,
+                onlyIfUsing, null, null, null, null, null,
                 null, null, null, null);
     }
 
--- a/src/test/java/org/openrewrite/java/testing/junit5/AddJupiterDependenciesTest.java
+++ b/src/test/java/org/openrewrite/java/testing/junit5/AddJupiterDependenciesTest.java
@@ -99,4 +99,75 @@ class AddJupiterDependenciesTest implements RewriteTest {
           )
         );
     }
+
+    @Test
+    void doNotAddWithoutJUnit() {
+        rewriteRun(
+          mavenProject("project",
+            srcTestJava(
+              //language=java
+              java(
+                """
+                  public class NotATest {
+                      public void addition() {
+                          assert 4 == 2 + 2;
+                      }
+                  }
+                  """
+              )
+            ),
+            pomXml(
+              //language=xml
+              """
+                <project>
+                    <modelVersion>4.0.0</modelVersion>
+                    <groupId>org.example</groupId>
+                    <artifactId>project</artifactId>
+                    <version>0.0.1</version>
+                </project>
+                """
+            )
+          )
+        );
+    }
+
+    @Issue("https://github.com/openrewrite/rewrite-testing-frameworks/issues/1113")
+    @Test
+    void addWhenOnlyJUnit3TestCaseIsUsed() {
+        rewriteRun(
+          spec -> spec.parser(JavaParser.fromJavaVersion().classpathFromResources(new InMemoryExecutionContext(), "junit-4")),
+          mavenProject("project",
+            srcTestJava(
+              //language=java
+              java(
+                """
+                  import junit.framework.TestCase;
+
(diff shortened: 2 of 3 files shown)


timtebeek (approved):
Thanks a lot! Great to see coverage expanded back to v3

PR #1122, opened 2026-09-13
Add WireMock 3 to 4 migration recipe

sullis:
Adds `org.openrewrite.java.testing.wiremock.Wiremock3to4Migration`, a composite recipe for upgrading WireMock 3.x to 4.x, along with the sub-recipes it needs.

WireMock 4 requires Java 17, ships Jetty 12.1 by default, splits the previously monolithic `org.wiremock:wiremock` artifact into modules, repackages the Jetty/servlet integration classes, and makes the primary domain classes immutable.

### Declarative recipes (`wiremock4.yml`)

| Recipe | What it does |
| --- | --- |
| `Wiremock3to4Migration` | Entry point: bumps `wiremock` and `wiremock-standalone`, folds the now-redundant `wiremock-jetty12` back into `wiremock` (4.x dropped the separate Jetty 12 distribution), then runs everything below |
| `AddWiremockJUnitDependencies` | JUnit 4/5 support moved to `wiremock-junit4` / `wiremock-junit5`, which are no longer transitive; adds whichever the project actually uses |
| `AddWiremockJettyDependencies` | `jetty-client`, `jetty-proxy`, `jetty-xml` and `jetty-ee10-webapp` are no longer transitive, so they're declared where used; `jetty-servlet`/`jetty-servlets` move to their Jetty 12 EE10 coordinates |
| `MigrateWiremockRepackagedTypes` | 35 `ChangeType` entries for the repackaged Jetty, servlet and Apache HTTP client classes; `jetty11` and `jetty12` both became `jetty` |

Plus `ChangeMethodName` for `StubMapping.getUuid()` → `getId()` and `RequestMethod.value()` → `getName()` (`value()` widened its return type to `Object` so multi-method patterns can serialize as something other than a string).

### Imperative recipes

- **`ReplaceRemovedConstructors`** — the removed convenience constructors of `StubMapping`, `ResponseDefinition`, `RequestPattern` and `WireMock` become the equivalent builder call, and `RequestPattern.everything()` becomes `RequestPattern.ANYTHING`.
- **`ReplaceSettersWithTransform`** — in-place mutation such as `stubMapping.setRequest(pattern);` becomes `stubMapping = stubMapping.transform(builder -> builder.setRequest(pattern));`, collapsing consecutive setters on the same receiver into one `transform` call.
- **`RemoveDuplicateContentTypeHeader`** / **`RemoveDuplicateContentTypeStubHeader`** — WireMock 3 ran on Jetty 11, which stripped every `Content-Type` response header but the last, so a stub configured with several served only one. 4.x returns all of them and some clients reject that, so only the last value (the one 3.x actually sent) is kept. Covers both Java stub definitions and JSON stub files; request header matchers are left alone since those hold matcher objects rather than plain strings.
- **`MigrateStubMappingUuidToId`** — 4.x dropped the redundant `uuid` field from serialized stub mappings. A stub carrying only `uuid` still parses, silently getting a random identifier instead, which breaks `removeStub`, `editStub` and `PUT /__admin/mappings/{id}`. Renames `uuid` to `id`, or drops it where an `id` is already present. Only objects that look like stub mappings are touched (a `request` or `response` member plus a `uuid` holding a UUID).
- **`MigrateContentTypeHeaderCharset`** — `ContentTypeHeader.charset()` used to fall back to `UTF_8` when the header was missing or carried no charset; 4.x returns an `Optional` that is empty in those cases. Appends `orElse(StandardCharsets.UTF_8)` so the value stays what it was, rather than the easier `get()`/`orElse(null)` that would turn a missing charset into an exception or a null.
- **`MigrateRequestMethodIsOneOf`** — `isOneOf` went from an instance method returning a boolean to a static method building a matcher, so `method.isOneOf(GET, POST)` quietly stops being a boolean. Rewritten as `RequestMethod.isOneOf(GET, POST).match(method).isExactMatch()`.

### Notes

- Targets `4.0.0-beta.38`.
- `RequestMethod.GET_OR_HEAD` is a multi-method matcher in 4.x rather than a method named `GET_OR_HEAD`, so code reading a name off that particular constant still needs a human look. This is called out in the composite recipe's description rather than rewritten.
- Adds `org.wiremock:wiremock:3.13.2` to the parser classpath and `rewrite-json` as an implementation dependency (needed for the stub-file recipes).
- Each recipe has a test class; `Wiremock3to4MigrationTest` covers the composite end to end.


sullis:
ready for review @greg-at-moderne  @timtebeek

steve-aom-elliott:
@sullis I think you might want to rebase off of remote main since it looks like the failing tests are ones that have been since fixed there.

sullis:
> @sullis I think you might want to rebase off of remote main since it looks like the failing tests are ones that have been since fixed there.

I have rebased off of remote main.

greg-at-moderne:
Can you try this test:
```java
@Test
void finalFieldReceiverIsLeftAlone() {
    rewriteRun(
      //language=java
      java(
        """
          import com.github.tomakehurst.wiremock.stubbing.StubMapping;

          class Stubs {
              private final StubMapping mapping = null;

              void describe() {
                  this.mapping.setName("get user");
              }
          }
          """
      )
    );
}
```

src/main/java/org/openrewrite/java/testing/wiremock/MigrateStubMappingUuidToId.java
```diff
@@ -0,0 +1,145 @@
...
+    @Getter
+    final String description = "WireMock 3 serialized a stub mapping's identifier as both `id` and `uuid`, but 4.x " +
+            "dropped the redundant `uuid` field. A stub that carries only `uuid` still parses under 4.x, silently " +
+            "getting a randomly generated identifier instead, which breaks anything addressing the stub by id such " +
+            "as `removeStub`, `editStub` or `PUT /__admin/mappings/{id}`. Rename `uuid` to `id`, or drop it where an " +
+            "`id` is already present. Only objects that look like stub mappings are considered, meaning they have a " +
+            "`request` or `response` member and a `uuid` holding a UUID.";
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
```
greg-at-moderne, line 56: Probably worth adding a precondition on file path as it might catch too much - i.e. files not related to WireMock at all (?)

src/main/java/org/openrewrite/java/testing/wiremock/ReplaceRemovedConstructors.java
```diff
@@ -0,0 +1,323 @@
...
+
+    @Value
+    private static class Replacement {
+        String owner;
+        String pattern;
+        String code;
+
+        MethodMatcher getMatcher() {
+            return new MethodMatcher(pattern);
+        }
```
greg-at-moderne, line 314: Any reason not to re-use the matchers?

greg-at-moderne (commented):
Great job. Left a few comments. It'd be nice if you could have a look.

src/main/resources/META-INF/rewrite/wiremock4.yml
```diff
@@ -0,0 +1,296 @@
...
+      oldArtifactId: jetty-servlet
+      newGroupId: org.eclipse.jetty.ee10
+      newArtifactId: jetty-ee10-servlet
+      newVersion: 12.1.x
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.eclipse.jetty
+      oldArtifactId: jetty-servlets
+      newGroupId: org.eclipse.jetty.ee10
+      newArtifactId: jetty-ee10-servlets
+      newVersion: 12.1.x
```
timtebeek: I worry a bit about changes beyond WireMock when we change the Jetty dependencies folks might have for reasons other than their use with Wiremock, especially as those would have been transitive before, and hence unlikely to be added explicitly in support of Wiremock alone. What are your thoughts here? Do we need to keep these changes at all?

src/main/resources/META-INF/rewrite/wiremock4.yml
```diff
@@ -0,0 +1,265 @@
...
```
timtebeek, line 73: I worry a bit about changes beyond WireMock when we change the Jetty dependencies folks might have for reasons other than their use with Wiremock, especially as those would have been transitive before, and hence unlikely to be added explicitly in support of Wiremock alone. What are your thoughts here? Do we need to keep these changes at all?

src/main/resources/META-INF/rewrite/wiremock4.yml
```diff
@@ -0,0 +1,265 @@
...
+      artifactId: jetty-xml
+      version: 12.1.x
+      onlyIfUsing: org.eclipse.jetty.xml.*
+      scope: test
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: org.eclipse.jetty.ee10
+      artifactId: jetty-ee10-webapp
+      version: 12.1.x
+      onlyIfUsing: org.eclipse.jetty.ee10.webapp.*
+      scope: test
```
timtebeek, line 97: Similarly here I wonder if we might inadvertently add explicit Jetty dependencies in the test scope only, which could break applications that rely on Jetty being present at runtime. I suppose it's hard to know if Jetty was only used for Wiremock, but I'd lean towards perhaps removing these steps out of caution as well. Do you agree?

src/main/resources/META-INF/rewrite/wiremock4.yml
timtebeek: Removed in https://github.com/openrewrite/rewrite-testing-frameworks/pull/1122/commits/ea5a8b0430653950be1726acd06d474464399353 ; it'll be unlikely these were added purely for wiremock if they also came in transitively previously.

src/main/resources/META-INF/rewrite/wiremock4.yml
timtebeek: Removed the scope, such that we do not limit that accidentally; the desired scope is derived through `onlyIfUsing` usage.

